PR #20, opened 2016-07-26
Do not check ssl on server validate

alkuzad:
should be parameterized somewhere but this is fix similar to the fix for disable ssl in electron itself.

### DIFF
--- a/app/renderer/js/main.js
+++ b/app/renderer/js/main.js
@@ -15,7 +15,7 @@
 		        	let newurl = 'https://' + url.replace(/^https?:\/\//,'')
 		        	let checkURL = newurl + '/static/audio/zulip.ogg';
 
-		        	request(checkURL, function (error, response, body) {
+		        	request({uri: checkURL, strictSSL: false}, function (error, response, body) {
 		        		if (!error && response.statusCode !== 404) {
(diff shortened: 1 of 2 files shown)


akashnimare:
@alkuzad Can you test it against this url - `http://akashnimare.in` . If we use `strictSSL` It will throw a 200 response even though the file https://akashnimare.in/static/audio/zulip.ogg does not exists. In this case this logic fails -

``` javascript
            if (!error && response.statusCode !== 404) {

                document.getElementById('main').innerHTML = 'Add' 
                document.getElementById('urladded').innerHTML = newDomain + '  Added';
                db.push('/domain', domain);
                ipcRenderer.send('new-domain', domain);
            }
            else {
                document.getElementById('main').innerHTML = 'Add' 
                document.getElementById('urladded').innerHTML = "Not a vaild Zulip server";
            }

```

alkuzad:
@akashnimare right now I'm behind proxy so I can not test your case easily but I've changed `zulip.ogg` to `zulip.oggnotexisting` and it throws 404 as expected

akashnimare:
@alkuzad I'm holding this PR as of now. Let me come up with a better solution to validate server.

PR #51, opened 2016-08-14
Remove node-json-db requirement, fixes #50

jedahan:
Seems that electron doesn't play nice with node-json-db.
Might be a filesystem access thing.

I have no idea what i'm doing, but now I have a nice shiny Zulip.app in ~/Applications that works, so hooray!

PR #53, opened 2016-08-15
Fix npm-test (xo code style)

jedahan:
As best as I could, now npm test only gives us a couple of warnings.


akashnimare:
Thanks @jedahan 😸 . Let's wait for https://github.com/zulip/zulip-electron/pull/51.

akashnimare:
@jedahan Great work. I see only 3 warnings now 👏 .

PR #83, opened 2016-11-09
Remove duplicate of script

GervaisdeM:
I'm not sure how both the renamed version of the script and the originally named version got merged into the repo. This PR removes one of them.


GervaisdeM:
@akashnimare There are two copies of the script in the repo. They are identical, so I suggest we remove one.

akashnimare:
@GervaisdeM yeah, I just noticed this. Thanks 😄 .

Issue #120, opened 2017-03-04
Error launching app using dev mode


geeeeeeeeek:
**Observation:**
- `Operating System`: macOS 10.12.1 (16B2555)
- `Clear steps to reproduce the issue`: `npm run dev`
- `npm version`: 4.1.2
- `node version`: v7.7.1
- `electron version`: v1.6.1
- `Relevant error messages and/or screenshots`:
![image](https://cloud.githubusercontent.com/assets/7262715/23579990/fd720fe0-0133-11e7-8101-ca74139b4495.png)

**Solution:**
According to an existed solution from [stackoverflow](http://stackoverflow.com/a/41847842), add back the missing `main` key in `package.json`.

akashnimare:
@geeeeeeeeek thanks for reporting this bug.

PR #121, opened 2017-03-04
#120 Fix error launching caused by missing  keyword.

geeeeeeeeek:


akashnimare:
Thanks @geeeeeeeeek 🚀

Issue #124, opened 2017-03-06
Server Errors(HTTP 5XX) not handled in UI


Lplenka:
Since today(06/03/2017) we witnessed the server(chat.zulip.org) was down 
Currently in electron app there is no support to display 5XX HTTP status code i.e. server errors , so if server is down  currently it just shows a clueless blank screen.After around a minute this is shown 
![screenshot from 2017-03-06 23-23-57](https://cloud.githubusercontent.com/assets/13533873/23624134/aba9250c-02ca-11e7-8fdc-9ac452516f72.png)

This error page was shown very late,there should be support to show such error as soon as user clicks the connect button.
-`Operating System` : Linux Mint 18.1
-`Zulip-Version` : 1.4.15

timabbott:
I think that's actually just what the website did as well at the time; this is what happens when a server is completely off the Internet unexpectedly.

Lplenka:
@timabbott  yes you are right , but in electron app this screen appears literally after a minute ,while in browser you can see the loading icon spinning , they atleast know something is wrong server side , but in electron app user is completely clueless for that span of time , I think we should add support in UI to handle this because our ultimate aim is a rich user experience by proper error handling too.

I hope I am clear @timabbott  @akashnimare

timabbott:
Ahh, I see what you mean; yeah we should have some sort of loading indicator ...

Lplenka:
@timabbott I think we can handle this error by keeping a check on server response status code (especially 5XX errors)and adding a loading indicator.

I would like to work on this

@zulipbot claim

zulipbot:
Congratulations, @Lplenka, you've successfully claimed your first issue! Since you're a new contributor, you now have pull (read-only) access to this repository.

Please review our [code contribution guidelines](http://zulip.readthedocs.io/en/latest/index.html#code-docs) and sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/) before opening any pull requests for any issues that you're working on.

Also, if you haven't done so already, please join us on the Zulip development server at [chat.zulip.org](https://chat.zulip.org/)! Our community of developers use **chat.zulip.org** as our primary communication forum for developing Zulip; you can read more about the **chat.zulip.org** community in our [documentation](http://zulip.readthedocs.io/en/latest/chat-zulip-org.html).

[...]

PR #130, opened 2017-03-18
Fixes Issue #124

Lplenka:
There was no support for handling server 5XX error, This PR fixes issue #124 .
Also sometimes setImmediate Reference Error occurs
setimmediate_error

I have added the removed Node global symbols back to the global scope when node integration is turned off.
This fixes the setImmediate error.
Separate commits for addressing separate part of the issue.

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -3,13 +3,9 @@ const path = require('path');
 const fs = require('fs');
 const os = require('os');
 const electron = require('electron');
-const {
-	app
-} = require('electron');
+const {app} = require('electron');
 const ipc = require('electron').ipcMain;
-const {
-	dialog
-} = require('electron');
+const {dialog} = require('electron');
 const https = require('https');
 const http = require('http');
 const electronLocalshortcut = require('electron-localshortcut');
@@ -18,13 +14,8 @@ const JsonDB = require('node-json-db');
 const isDev = require('electron-is-dev');
 const tray = require('./tray');
 const appMenu = require('./menu');
-const {
-	linkIsInternal,
-	skipImages
-} = require('./link-helper');
-const {
-	appUpdater
-} = require('./autoupdater');
+const {linkIsInternal,skipImages} = require('./link-helper');
+const {appUpdater} = require('./autoupdater');
 
 const db = new JsonDB(app.getPath('userData') + '/domain.json', true, true);
 const data = db.getData('/');
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -62,7 +62,7 @@ const targetURL = function () {
 function server_error(targetURL) {
 	if (targetURL.indexOf('localhost:') < 0 && data.domain) {
 		var req = https.request(targetURL + '/static/audio/zulip.ogg', (res) => {
-			console.log('Server statusCode:', res.statusCode);
+			console.log('Server StatusCode:', res.statusCode);
 			console.log('You are connected to:', res.req._headers.host);
 			if (res.statusCode >= 500 && res.statusCode <= 599) {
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -123,5 +126,8 @@
       "browser",
       "mocha"
     ]
+  },
+  "dependencies": {
```
akashnimare: App's dependencies should be in app/package.json.

app/main/index.js
```diff
@@ -3,17 +3,28 @@ const path = require('path');
 const fs = require('fs');
 const os = require('os');
 const electron = require('electron');
-const {app} = require('electron');
+const {
```
akashnimare: Change it to `const {app} = require('electron');`

app/main/index.js
```diff
@@ -3,17 +3,28 @@ const path = require('path');
...
+	app
+} = require('electron');
 const ipc = require('electron').ipcMain;
-const {dialog} = require('electron');
+const {
```
akashnimare: Same here `const {dialog} = require('electron');`

app/main/index.js
```diff
@@ -3,17 +3,28 @@ const path = require('path');
...
+const http = require('http');
 const electronLocalshortcut = require('electron-localshortcut');
 const Configstore = require('configstore');
 const JsonDB = require('node-json-db');
 const isDev = require('electron-is-dev');
 const tray = require('./tray');
 const appMenu = require('./menu');
-const {linkIsInternal, skipImages} = require('./link-helper');
-const {appUpdater} = require('./autoupdater');
+const {
```
akashnimare: Same. Keep in mind that all the exported variables should be in one line.
`const {linkIsInternal, skipImages} = require('./link-helper');`

app/main/index.js
```diff
@@ -3,17 +3,28 @@ const path = require('path');
...
+	linkIsInternal,
+	skipImages
+} = require('./link-helper');
+const {
```
akashnimare: Same.

Lplenka:
My Bad :(, That thing happens because of my Atom text editor's code beautify, It automatically adds space to imported variable

akashnimare:
No problem. It happens 😅

Lplenka:
@akashnimare  I have done the required changes, have a look

package.json
```diff
@@ -128,6 +128,5 @@
     ]
   },
   "dependencies": {
```
akashnimare: We don't need this here so please remove it. Only `devDependencies` should be here.

Lplenka: Okay sure

app/main/index.js
```diff
@@ -56,6 +58,31 @@ const targetURL = function () {
...
+			if (res.statusCode >= 500 && res.statusCode <= 599) {
+				return dialog.showErrorBox('SERVER IS DOWN!', 'We are getting a ' + res.statusCode + ' error status from the server ' + res.req._headers.host + '. Please try again after some time or you may switch server.');
+			}
+		});
+		req.on('error', e => {
+			console.error(e);
+		});
+		req.end();
+	} else if (data.domain) {
+		const req = http.request(targetURL + '/static/audio/zulip.ogg', res => {
```
akashnimare: When you want to request an https resource, you need to use https.get, not http.get.

Lplenka: That's only for localhost like "localhost:9991" server

Lplenka: I tried https, It doesn't work for localhost:9991,
so that else if condition is only to avoid error during switching to localhost or loading localhost directly.
I hope I am clear :)

akashnimare: Hmm, make sense.

akashnimare:
Merged. Thanks @Lplenka 💯

Lplenka:
Thank you :smile:

Issue #110, opened 2017-02-18
Unread counts on tray icon on Windows


mgrabow1:
Zulip-electron does not show unread counts in tray (on tray icon) on Windows (7+).

I think it should :) - just like the old windows native client. (as a complement to Windows notifications).

Lplenka:
@akashnimare  @timabbott  I think the unread counts doesn't appear in tray icon of our electron app in none of the  OS not only windows

Lplenka:
@zulipbot claim

zulipbot:
Hello @Lplenka, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for a week. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on this issue within 3 days, or else I'll automatically remove your assignment from this issue.

If you've decided to stop working on this issue, simply comment `@zulipbot abandon` to abandon this issue so someone else can claim it and continue from where you left off.

Thank you for your understanding and your valuable contributions to Zulip!

Lplenka:
I am working on it

synicalsyntax:
Hello @Lplenka, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for a week. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 3 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

PR #131, opened 2017-03-21
Fixes issue #110

Lplenka:
`zulip-electron:0.5.8`
`Operating System:Linux Mint 18`

Bug:
There was no support for showing unread count in app's tray icon in any OS
Fix:
This PR fixes issue #110 
Now unread count is shown if you hover over tray icon :)
Here is how it looks,
traycount


timabbott:
Shouldn't we show unread counts unconditionally without your needing to hover?  I feel like that's best practice with this sort of thing.

Lplenka:
@timabbott If you mean it should be like our  title of our web UI
chat

yes, it would be better but currently, I have only tried to fix a patch(this issue only).
As you must be knowing our web UI has many icons specifically designed for a specific number of unread counts like there is an icon for unread 6, unread 7, unread 10 etc present in `static/images/favicon` folder.
Current Electron app doesn't have these.

So I just tried to fix the issue without involving the design part

akashnimare:
> Shouldn't we show unread counts unconditionally without your needing to hover? I feel like that's best practice with this sort of thing.

@timabbott totally agreed. It's not so useful and not desirable. We''re looking for something like Telegram's badge feature - 

badge

@Lplenka as of now you're using `tray.setToolTip()` which sets a text on hover in tray icon. We need to use `setOverIcon()` which will set the badges. Maybe try this  -
``` js
setBadge = function (text) {
  if (process.platform === "darwin") {
    app.dock.setBadge("" + text);
  } else if (process.platform === "win32") {
    var win = remote.getCurrentWindow();

    if (text === "") {
      win.setOverlayIcon(null, "");
      return;
    }

    // Create badge
    var canvas = document.createElement("canvas");
    canvas.height = 140;
    canvas.width = 140;
    var ctx = canvas.getContext("2d");
    ctx.fillStyle = "red";
    ctx.beginPath();
    ctx.ellipse(70, 70, 70, 70, 0, 0, 2 * Math.PI);
    ctx.fill();
    ctx.textAlign = "center";
    ctx.fillStyle = "white";

    if (text.length > 2) {
      ctx.font = "75px sans-serif";
      ctx.fillText("" + text, 70, 98);
    } else if (text.length > 1) {
      ctx.font = "100px sans-serif";
      ctx.fillText("" + text, 70, 105);
    } else {
      ctx.font = "125px sans-serif";
      ctx.fillText("" + text, 70, 112);
    }

    var badgeDataURL = canvas.toDataURL();
    var img = NativeImage.createFromDataUrl(badgeDataURL);

    win.setOverlayIcon(img, text);
  }
};
```
This does not work in main process but it'll give you the head start. 

> unread counts like there is an icon for unread 6, unread 7, unread 10 etc. present in static/images/favicon folder. Current Electron app doesn't have these.

There is a thing called [native-image](https://github.com/electron/electron/blob/master/docs/api/native-image.md) for this.

akashnimare:
Also check out these issues -
https://github.com/electron/electron/issues/7325
https://github.com/electron/electron/issues/7440
https://discuss.atom.io/t/native-image-examples/26648/2

PR #134, opened 2017-04-02
Fixes #110

### LINKED ISSUE #110: Unread counts on tray icon on Windows

Lplenka:
There was no support for showing unread count in the Tray icon, I have added the functionality.
trayicon

There is no support for overlay icon in Linux as of now please refer issue [#9020](https://github.com/electron/electron/issues/9020).

This fix works fine.
Any suggestions are welcome :smile:


sinwar:
@Lplenka It'll be great if you use different branch than master to create pull request.

akashnimare:
> It'll be great if you use different branch than master to create pull request. 
🙏

app/renderer/js/tray.js
```diff
@@ -0,0 +1,169 @@
...
+
+ipcRenderer.on('toggletray', event => {
+	if (event) {
+		if (window.tray) {
+			window.tray.destroy();
+			if (window.tray.isDestroyed()) {
+				window.tray = null;
+			}
+		} else {
+			window.createTray();
```
akashnimare: should be `createTray()`

akashnimare:
on macOs - 
image

app/renderer/js/tray.js
Lplenka: Yes, will change it.

Lplenka:
Size issue on different OS

akashnimare:
The one big problem with this PR is that we don't want to completely replace the tray icon with that canvas image. The ideal solution is to set that canvas on top of Zulip tray icon. Something like this -

icon

So that it look like this -

screen shot 2017-04-04 at 12 34 55 am

You gotta use `setOverlayIcon` method for this. As of now this method only supports Windows as mentioned [here](https://electron.atom.io/docs/api/browser-window/#winsetoverlayiconoverlay-description-windows).
More info - 
https://github.com/electron/electron/issues/4011
https://github.com/electron/electron/issues/3148

Lplenka:
Okay, Since there is no support to overlay tray in Linux & MacOS I thought this will work for the time being. Because we cannot implement some thing which electron doesn't have itself.

Lplenka:
The original zulip Icon will always be there if no unread messages are present.
If new messages come canvas will appear with the specific unread count.
This is the best fix I could find until the electron actually supports overlay.
I hope I am making sense.

akashnimare:
Wait on. Let me try to setup `setOverlayIcon` on windows and see if we still can get something like this (atleast on windows) -

akashnimare:
> The original zulip Icon will always be there if no unread messages are present.
If new messages come canvas will appear with the specific unread count.
This is the best fix I could find until the electron actually supports overlay.
I hope I am making sense.
Definitely, it makes sense. It's probably a good fix until electron supports this api on macOS + Linux 😃

Lplenka:
Yes, we can handle this differently in different OS. If Windows has overlay we can use that, but then we will have a different implementation in different OS. Would that be fine?
I have seen other electron apps too, none of them has this feature of Overlay, so I could only come up with this idea.

akashnimare:
> If Windows has overlay we can use that, but then we will have a different implementation in different OS. Would that be fine?
Yes.

timabbott:
For the webapp, we just have a collection of 100 generated images that we use for favicons (if the number exceeds 100, we just show an infinity symbol).  So one doesn't need overlay support -- just the ability to change the image -- to produce arbitrary numbers.

Lplenka:
@timabbott yes I feel like that's the only more possible way to deal with this Issue.

akashnimare:
I think we could get these favicons using - 
```js
	page.on('page-favicon-updated', (event, favicons) => {
		console.log(favicons.slice(-1)[0]);
	});
```
output - 
```sh
➜  zulip-electron git:(master) ✗ npm start

> zulip@0.5.8 start /Users/akka/dev/zulipwork/zulip-electron
> electron ./app/main

Wait for app ready
Server StatusCode: 200
You are connected to: chat.zulip.org
https://chat.zulip.org/static/images/favicon/favicon-38.png
https://chat.zulip.org/static/images/favicon/favicon-93.png
https://chat.zulip.org/static/images/favicon/favicon-93.png
https://chat.zulip.org/static/images/favicon/favicon-93.png
```
I was just wondering if somehow we could set these favicon to our tray icon 🤔

Lplenka:
> I was just wondering if somehow we could set these favicon to our tray icon :thinking:.

We can receive the image using http request to specific URL like
> request
  .get('http://mysite.com/doodle.png')
  .on('error', function(err) {
    console.log(err)
  })
  .pipe(fs.createWriteStream('doodle.png'))'

But I think size issue will still remain for different OS :confused:

akashnimare:
Will try to run this PR on windows tonight plus I'll see if we could use `setOverlayIcon`.

Lplenka:
> Will try to run this PR on windows tonight plus I'll see if we could use setOverlayIcon.

Okay :+1:

akashnimare:
Will review it again on Monday.

akashnimare:
Does it work on Windows 10?

akashnimare:
Please move this file -
* `app/renderer/js/tray.js` to `app/main/tray.js`

Lplenka:
> Does it work on Windows 10? Can you paste a screenshot?

Yes works perfectly fine, here are some screenshots.

screenshot 36
screenshot 37
screenshot 38
windwostray

Lplenka:
@akashnimare  Is this necessary?

> Please move this file -
app/renderer/js/tray.js to app/main/tray.js (tray.js should be in the browser process)

Initially, I tried running this file in `app/main/tray.js` but things didn't work if tray.js was imported into `app/main/index.js`, the tray icon image was never rendered. So I shifted to Renderer process and everything worked as expected.

 I have added every functionality of Tray icon using  `IpcMain` and `IpcRenderer` for proper communication between Main and Renderer processes. The icon works perfectly well while switching servers, reloading page, logging in/out etc.

Also if Tray icon is in renderer process it will be helpful to add required functionalities for our "Multiple Server Support" feature.

akashnimare:
> Initially, I tried running this file in app/main/tray.js but things didn't work if tray.js was imported into app/main/index.js, the tray icon image was never rendered. So I shifted to Renderer process and everything worked as expected.

You can't import tray.js in Main process since it belongs to Renderer. 

> Also if Tray icon is in renderer process it will be helpful to add required functionalities for our "Multiple Server Support" feature.

Make sense.

akashnimare:
@Lplenka why does it create multiple tray icon in some cases (ex. navigating to a new URL)? Although it gets disappeared after few seconds. Can we prevent this event?

Lplenka:
> You can't import tray.js in Main process since it belongs to Renderer.

I meant currently our app does that, we import `app/main/tray.js` to `app/main/index.js` and call function `tray.create();`.

While In this PR since I shifted tray.js file to renderer, I am just calling this file in `app/main/preload.js` so the file runs every time app's window is created.

Lplenka:
> @Lplenka why does it create multiple tray icon in some cases (ex. navigating to a new URL)? Although it gets disappeared after few seconds. Can we prevent this event?

Yes I specifically dealt with this problem in `index.js`, Added this,

>          //  To destroy tray icon when navigate to a new URL
 	win.webContents.on('will-navigate', e => {
 		if (e) {
 			win.webContents.send('destroytray');
 		}
 	});

Lplenka:
> Cross tested this PR on all the platforms. Works great. This is probably the best solution to show unreadcounts in the tray icon. We could improve it later once electron officially support it with setOverlayicon etc. @Lplenka keep in mind that eventually we should add something like this
#134 (comment)
Note - Sorry for the delay. Thanks for all the work :rocket:  :heart:

Thanks! :smile: and  yes I will definitely work on that as soon as required feature is implemented in Electron.

Issue #123, opened 2017-03-06
Add a webpage/popup for desktop app settings


akashnimare:
A dedicated webpage/popup would be great for all the settings. Something like - 

<img src="https://thomas101.github.io/wmail/images/settings_1.png"/>
cc @timabbott @geeeeeeeeek

geeeeeeeeek:
Got it.

gautamchitnis:
Is anyone working on this issue currently?

Lplenka:
@gautamchitnis probably no one is working because it's not claimed yet and no comments since 11 days, you can claim it. Have a look at (this)[https://github.com/zulip/zulipbot/blob/master/README.md].

gautamchitnis:
@zulipbot claim

zulipbot:
Congratulations, @gautamchitnis, you've successfully claimed your first issue! Since you're a new contributor, you now have pull (read-only) access to this repository.

Please review our [code contribution guidelines](http://zulip.readthedocs.io/en/latest/index.html#code-docs) and sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/) before opening any pull requests for any issues that you're working on.

Also, if you haven't done so already, please join us on the Zulip development server at [chat.zulip.org](https://chat.zulip.org/)! Our community of developers use **chat.zulip.org** as our primary communication forum for developing Zulip; you can read more about the **chat.zulip.org** community in our [documentation](http://zulip.readthedocs.io/en/latest/chat-zulip-org.html).

**Note:** Depending on your [notification settings](https://github.com/settings/notifications), granting you the permission in GitHub's model that allows you to be assigned issues by **zulipbot** may result in making you a "watcher" on this repository, which means that you'll get notifications for every issue and pull request in this repository; for a large repository like [zulip/zulip](https://github.com/zulip/zulip), that means receiving over 100 emails per day. If you don't want to receive notifications unless you're actually involved with the issue (which is usually the right choice for large projects), you should [unwatch this repository](https://help.github.com/articles/unwatching-repositories/).

gautamchitnis:
I have setup a dev env and have claimed this issue. If I cannot come up with a solution in a day or so, I'll abandon the issue straight away.

gautamchitnis:
@zulipbot abandon

apollonian11:
@zulipbot claim

zulipbot:
[...]

PR #140, opened 2017-04-20
[WIP] #1 Add nice support for multiple Zulip servers

geeeeeeeeek:
#1 Working in progress. @akashnimare @timabbott Comments are welcomed. 😸


geeeeeeeeek:
**Changes:**
 - UI template for multi-server manager done.
 - Remove title bar. Using draggable sidebar as a replacement.
screen shot 2017-04-20 at 11 13 17 pm

**Next step:**
 - Add interactions and basic logic for the manager by Javascript.

app/renderer/css/servermanager.css
```diff
@@ -0,0 +1,86 @@
...
+.server-button.active .server-ring {
+    background: #43bba6;
+    border-radius: 24px;
+    width: 48px;
+    height: 48px;
+    position: absolute;
+    left: -2px;
+    top: 3px;
+    z-index: 10;
+}
```
timabbott: A newline is missing here.  We should consider making our linter check for this.

app/main/index.js
```diff
@@ -170,11 +174,12 @@ function createMainWindow() {
 		icon: iconPath(),
 		minWidth: 600,
 		minHeight: 400,
+		titleBarStyle: 'hidden-inset',
 		webPreferences: {
 			preload: path.join(__dirname, '../renderer/js/preload.js'),
 			plugins: true,
 			allowDisplayingInsecureContent: true,
-			nodeIntegration: false
+			nodeIntegration: true
```
timabbott: @akashnimare I seem to recall that enabling `nodeIntegration` caused weird problems with the webapp's use of jQuery from node; do you remember the details?  Regardless, we should add a comment documenting why it's disabled if we need it to still be disabled, so folks know there's a reason.

borisyankov:
Since I think @geeeeeeeeek is following the Slack design, here it is for reference:
https://ga1.imgix.net/apps/91563/595ecc46734158fd08883a8818b2bbab

Their use of circle/bar at the edge is better since it doesn't look like a list item bullet.
On the other hand we don't need both the outline and the bullet.

In the mobile app we do show a check mark to the selected item, but that is because we show a full line vs just icons here. If we are going with the outline, remember we can not do that for an arbitrary shape, and we have to agree upfront if the realm icons will be circled or not.

borisyankov:
On an unrelated note, how is the title bar going to work on Linux/Windows?
I am usually quite opposed on destandardizing vital UI elements when no good reason.

geeeeeeeeek:
@timabbott 
 - The dot on the left and highlighted ring suggests the server is being selected and displayed on the right hand side.
 - Make sense. I thought of using the 'favicon' but to my knowledge most organizations won't change the default icon. Then it might be difficult to distinguish different servers with the same icon. What do you think?
 - Thought about it too. Unread count is very reasonable and meaningful. But I think it's better to get this prototype running and add this feature a bit later.
- For the user list in the right hand side, I don't see a feasible way to show it unless we increase the width or users do so.

Actually, per discussion in #1, I am following the design that @akashnimare gave:


Thanks @borisyankov for your comments,
 - I think Slack's design looks better. What do you think if I use round corner square and a left bar instead, like Slack? I will give you a prototype some hours later.
 - In Linux and Windows, the title bar will be displayed on top, not like hidden-inset in macOS. It needs a platform-dependent CSS which I haven't done yet. You can get the idea from my [electronic wechat](https://github.com/geeeeeeeeek/electronic-wechat).

I will cc the discussions to the design stream as we'll :)

app/main/index.js
zhongyi-tong: Actually, the only option on the table is to enable nodeIntegration, or we cannot use the WebView element, then the whole idea won't work…

I'm not sure what impact it will have on jQuery… It sounds a bit weird.…

borisyankov:
We briefly discussed with @timabbott the possibility of realm's icon being aliased to a known url.
Like yourrealm.com/icon.png 
This way to show 5 realms you just do 5 img tags instead of doing 5 get requests and then update the html etc. The more I think about it, the more I like this idea.

timabbott:
Yeah, we can definitely alias the icon to a known URL -- would be trivial to do.  I'll open an issue.

I also prefer the rounded corner square to the circles, and I think a bar on the left is probably better than the dot for indicating which is selected (though in that Slack screenshot in particular, it looks like a bug because of the white box nearby).

Totally agreed re: unread count being a reasonable second-round feature; just wanted to make sure we agree about the vision and are thinking about where to put things :)

timabbott:
Opened https://github.com/zulip/zulip/issues/4564 for the icon URL project.

timabbott:
When @akashnimare is online, he'll be able to remind us about the issue with jquery.  I think it might be the case that we fixed it in both the server project and disabled this for now in the desktop app, so it might be OK to enable the node integration again.  Not sure though -- we should check!

geeeeeeeeek:
Updated design:

screen shot 2017-04-21 at 2 39 07 pm

geeeeeeeeek:
Changes:
 - Add DomainUtil to manage domains in one place.
 - Setup the layout for ServerManagerView.

Please note that the following schema will be used when adding a new server:
```json
{
            url: 'https://chat.zulip.org',
            alias: 'Zulip 2',
            avatar: 'https://chat.zulip.org/static/images/logo/zulip-icon-128x128.271d0f6a0ca2.png'
}
```
 - `url`: Domain of the server.
 - `alias`: Name of the server. Useful for window title, or elsewhere. Easier and more friendly to distinguish each server.
 - `avatar` (optional): An image url. Used to display on the left sidebar. I noticed @timabbott thought about adding an endpoint for icon on Zulip Server. But I think it will not be done in a while and users may not be in time to update their server. We will detect that icon, if not existed, turn to user's customized avatar, and then the default Zulip icon.

Objects of different servers are stored in JsonDB endpoint `/domains[]` as an array.

Comments are welcomed! It's important that we're on the same page and then move things forwards!

timabbott:
We should probably rename `avatar` to `icon`, just since that's what it's called in the server and API.

Maybe also s/alias/name/ would be good as well.

akashnimare:
## Note - Before proceeding further on this PR we should first solve the node integration problem.

#### Problem - Due to the Node.js integration of Electron, there are some extra symbols inserted into the DOM like module, exports, require. This causes problems for some libraries like jQuery since they want to insert the symbols with the same names. With nodeIntegration enable we will get -

image

This is caused because library will probably contain something similar to this:
```js
if ( typeof module === "object" && typeof module.exports === "object" ) {
  // set jQuery in `module`
} else {
  // set jQuery in `window`
}
```
This will create problems since the library will use module exports to create definition on it, instead of using the global object. So even though the library is correctly loaded, it will find module (since Electron is using Node.js), and it will put definitions there.

## Solutions - 
* Make some changes on web app level so that it works with electron app.
* Use `nodeIntegration:false` and put all the necessary scripts in `preload.js`. I did the same with [domain.js](https://github.com/zulip/zulip-electron/blob/master/app/renderer/js/domain.js). We gotta load the functions within `window`.

Earlier I have tried to fix this problem on electron level but had no luck. I think we should try to fix it in the web app. There are few solutions mentioned here which might work -

https://github.com/electron/electron/issues/254
https://github.com/electron/electron/issues/254#issuecomment-183504714
https://github.com/electron/electron/issues/254#issuecomment-183483641
https://github.com/electron/electron/issues/254#issuecomment-183498783

geeeeeeeeek:
@akashnimare Thanks for the clarification. I will look into that and see what to do.

geeeeeeeeek:
**Changes:**
 - Change `avatar` to `icon` in domains config schema as @timabbott suggested.
 - Finish the basic functionality of `ServerManagerView`. Now it has the ability to switch servers smoothly with a loading indicator and CSS transitions.

**Screenshot:**
b

**Next Steps:**
 - Refactor the previous `Add Server` and `Switch Server` pages, with new domain config schema and consistent styles.
 - Integrate with `index.js`. Many code might have been broken we move contents from `BrowserWindow` to `WebView`.

geeeeeeeeek:
Per discussions in #123, we need a dedicated Preference page. As far as I'm concerned, I think add/edit/remove servers should be a part of it. Once the user click `Settings` button, the Settings tab will show up on the left and Settings page loaded in the webview.

Here is my implementation (WIP):
screen shot 2017-04-24 at 12 00 30 am

I'll finish the Servers settings part and leave the rest to #123 (@apollonian11). Can we agree on the design? @timabbott @akashnimare

timabbott:
I agree that managing servers should probably be part of the electron app's settings UI.

I think that's a reasonable structure to start with.  It's possible we want to do something the recalls the styling of settings in the webapp, but we can always make it fancier later.

akashnimare:
Just pulled it and ran into the following issue -

image

Is it working or still in progress?

akashnimare:
> Can we agree on the design?

This looks good. For now, just add `edit/remove servers setting`. We'll add few more prefs later.

> Many code might have been broken we move contents from BrowserWindow to WebView.

Let me know if you need any help in this.

geeeeeeeeek:
@akashnimare Users will be redirected to Settings page to add their first server. The original add server page is no longer needed. I'll fix the broken index.js after that.

geeeeeeeeek:
**Big step forward!**
 - Change of technical design. Use one webview for each tab, to enable:
   - Servers switching on the flying. Load at the first time. When you switch back to it, it takes no time to display.
   - Users are be online at all these servers at the same time.
 - Set up Servers Management part in PreferenceView (Still in progress).

c

**Next Steps:**
 - Finish Servers Management settings.
 - Integrate with other parts of the app (`index.js` especially).

This PR is estimated to be 100% done in this week.

geeeeeeeeek:
**Changes:**
All done. Some actions were applied to `BrowserWindow` but now `WebView`. All actions (including `menu` and `tray`) are properly handled. A large proportion of code has been rewritten.

**Next Steps:**
 - Review my code. Test it on other platforms.
 - Fix CI errors (possibly related to the linter).
 - Write a follow-up plan.

You can expect this PR to be ready to merge on the weekends. But code review and testing can start now! @akashnimare @timabbott

timabbott:
Just tried testing it out on Linux.  The initial state hung looking like this:
* image (with
image in the JS console).  I assume this basically is because the new version expects a different cache format than the old version?  We should handle this case.

Lplenka:
> I assume this basically is because the new version expects a different cache format than the old version? We should handle this case.

Yes @timabbott  this is happening because the app is looking for the `domains` array.
Which is not present in your system's  `domain.json`

He must have edited the `domain.json` manually while working on this PR.
I haven't pulled this PR, but this fix should work, I used this in my prototype for "Multiple server support"

@geeeeeeeeek add this to `app/renderer/js/main.js`
This basically adds an empty `domains` array to the `domain.json` if not already there.

```javascript
const data = db.getData('/');
if (!data.domains) {
	db.push("/domains", []);
}
```

geeeeeeeeek:
My fault. Should have initialized `/domains` if not exists.

geeeeeeeeek:
- Domains config initialization bug fixed. 
 - Rebased from `master`. 
 - Linter errors fixed. 

Continue verifying potential bugs and testing on multiple platforms.

timabbott:
I'm still seeing that same error running the latest version of this branch.

I think maybe we need to write code to convert an old-format domains file to the new format?  I tried deleting the `domains.json` file in the Git repo, and that didn't resolve the issue, so it's possible the bug is something different (certainly that exception looks more like an error trying to find a file that doesn't exist).

timabbott:
This works on my machine now!  

A few other issues:
* "Reload to apply changes" is mispelled
* I do notice that every time I open the app or click something, the JS console opens

geeeeeeeeek:
@timabbott Fixed those two! Testing in progress.

image

timabbott:
Awesome; the core functionality works great for me!  I think @akashnimare is going to be offline for a few days, so there may be a bit of delay before you get a full code review.  In the meantime:
* The UI for adding a new server could probably be cleaned up a bit (the "Validate and Save" button at the top is sorta confusing, and the confirmation dialog + reload flow could use work).
* I seem to have lost my second server when restarting the app without quitting cleanly in the same session where I added it? I'm not totally sure I can reproduce this, but probably worth checking the save logic.
* We can start trying to figure our how best to do unread counts.  The way the old desktop app did unread counts involved the fact that it put a special `window.bridge` element in the javascript context, and we called that if it exists from the places one might want to do something special on desktop; see e.g.: https://github.com/zulip/zulip/blob/master/static/js/notifications.js#L178.  We could potentially just hook into that API, or create a new one guided by where that is used (though the advantage of using the existing API is it's available for every Zulip server, even old ones).
* I think it's probably worth getting a few beta testers for this PR by asking for testers on chat.zulip.org, 
* I imagine there's work that could be done on cleaning up the Git history to be easier to review :).

Huge thanks for all your work on this so far @geeeeeeeeek!

timabbott:
I just found one more somewhat subtle bug; in Zulip, when you focus the window, it marks any messages in view as read (if the bottom_whitespace is visible, at least).  That seems to not work with the app right now; I suspect because the switcher area is getting the focus.

geeeeeeeeek:
@timabbott Thanks for the review.

 - I will update the label of Save Server button.
 - I could not reproduce this issue. Did you see a `success` alert after saving?
 - Might be a good idea to read `window.bridge` instead of from title.
 - As for the last bug you mentioned, I don't know the cause yet. From previous experience, even if the webview is not focused (like the second webview which is hidden), the badge number should be properly updated. 

Plus, I will also take some time off till the end of the month. Preparing for my undergraduate thesis and some other stuff. If you need any help on this PR, please do let me know. I'll get back to you as soon as I can.

Meanwhile, I'm thinking of moving the code handling multi-tab webviews to a separate repo. The repo will provide the whole UI and APIs like `addTab`, `getTab`, and event listeners. You can track my progress on [electron-tabview](https://github.com/geeeeeeeeek/electron-tabview). These work is estimated to be done (or not) before May 30. My `electronic-wechat` app will use the framework to support multiple account login. And I think `zulip-electron` can reuse it.

timabbott:
By end of the month, do you mean April or May?  I imagine once @akashnimare is back in a couple days he'll review and hopefully we can integrate this soon, and then work on further improvements from there.  

Aside from whatever he finds reading the code, I think basically the focus bug is the only blocker issue I'm still aware of (it doesn't just affect marking things as unread, but also whether you need to click in before clicking on messages starts a reply, etc.).  I'm seeing it on Linux; could be platform-specific though I kinda doubt it.

geeeeeeeeek:
@timabbott By May 30. I mean I will not be able to work on other features/bugs during the time. But should be able to follow up this PR. As for the bug, I will have more tests.

timabbott:
@geeeeeeeeek makes sense.  BTW I closed https://github.com/zulip/zulip/issues/4564, since I think the `GET /api/v1/server_settings` endpoint is probably a better way to do it.  I think I CCed you one the tricky notes around using that related to backwards-compatbility.

akashnimare:
I'm done with academic stuff 😫  
will start to review this PR now.

zhongyi-tong:
Great. My graduation thesis will soon be done as well!

tests/index.js
```diff
@@ -20,7 +20,7 @@ describe('application launch', function () {
 
   it('shows an initial window', function () {
     return this.app.client.getWindowCount().then(function (count) {
-      assert.equal(count, 1)
+      assert.equal(count, 2)
```
SimplyAhmazing: Just curious about why the window count increased to 2? There doesn't seem an extra window displayed at start.

SimplyAhmazing: Got it. ```client.getWindowCount(): Gets the number of open windows. <webview> tags are also counted as separate windows.```

app/renderer/js/main.js
```diff
@@ -0,0 +1,251 @@
...
+			if (title.indexOf('Zulip') === -1) {
+				return;
+			}
+
+			let messageCount = (/\(([0-9]+)\)/).exec(title);
+			messageCount = messageCount ? Number(messageCount[1]) : 0;
+
+			this.badgeNumberList[index] = messageCount;
+
+			const sum = this.badgeNumberList.reduce((a, b) => {
```
SimplyAhmazing: This PR shows unread message badge counts as a total for all organizations. The unread messages also sum up as one visits the organizations... 

I.e.: When I open the app, the badge count shows the unread messages from organization 1. When I navigate via the side panel to organization 2, the badge count sums up the unread messages from both organization 1 & 2.. and so on..

I think it would be more desireable to show the unread badge count as either,

1. The sum of all of the users organizations
2. Or the badge count should switch when a user switches organizations.

By 'badge count' I am referring to both the tray and dock numbers on OSX. Have not tried on other OSs.

/cc @akashnimare

akashnimare: Sounds good to me! We gotta update the badge counts instead of showing the sum of all the unread messages from different servers.That will make more sense.

SimplyAhmazing:
Made some fixes to this branch: https://github.com/zulip/zulip-electron/tree/140-fixes

* shrunk side bar
* made badge count reflect currently selected organization badge count
* added window frame to make dragging easier
* fixed focus issue with `webview.focus()` so keyboard arrow keys allow for navigation

screen shot 2017-05-22 at 9 36 07 pm
](url)

SimplyAhmazing:
@geeeeeeeeek Huge thanks for your work on this PR. We're currently at the PyCon sprints and wanted to clear the backlog of open PRs and issues.  There's still a good few things we'll improve with this feature and we'll keep at it.

zhongyi-tong:
@SimplyAhmazing @akashnimare Thanks for your review as well! I will continue to refine this story in GSoC. 😈

Issue #158, opened 2017-05-26
Hidden pop-up is focused  in background on context switch


jamak:
on OSX,  on ⌘-tab switch back into the electron app, a hidden window (I assume the settings pop-up?) is focused. In the browser, when you return to the tab, the home stream is focused. This is as of f6358a0 .

To reproduce:
* open zulip-electron
* switch to another application
* switch back to zulip-electron

akashnimare:
I'm not able to reproduce this. Can you please make a gif/video of the same?

akashnimare:
@SimplyAhmazing pushed a fix in https://github.com/zulip/zulip-electron/commit/deafa315dffb10c364556650e6f352c8f5621224. Can you verify if it works for you?

jamak:
Yep, works for me. Nice job, @SimplyAhmazing!

PR #159, opened 2017-05-28
Bugfix #158 focus webview on tab app switch

SimplyAhmazing:
This PR solves the issue reported by @jamak here: https://github.com/zulip/zulip-electron/issues/158

The issue is when you open Zulip and type text into a chat area, then switch applications via Command+Tab to another app, and return to the Zulip app, you will no longer have focus on the chat area. If you type on your keyboard the key events won't be sent to the text area you had activated before you switched out of the app. This doesn't happen with applications like Microsoft Word, you can tab switch to another app and return to continue typing where you had left your cursor.

This PR solves this issue by listening for a `focus` event on the browserWindow in the main process and forwarding the focus event to the renderer process. The listener on the renderer process will then focus the currently activated webview.

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -125,6 +125,10 @@ function createMainWindow() {
 		show: false
 	});
 
+	win.on('focus', () => {
+		win.webContents.send('focus')
+	})
+
 	win.once('ready-to-show', () => {
 		win.show();
 	});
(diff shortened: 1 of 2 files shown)


akashnimare:
It works as expected now. Thanks, @SimplyAhmazing.

PR #162, opened 2017-05-30
Fix WebView resize problem

zhongyi-tong:
**Steps to reproduce:**
1. Add more than one domains.
2. Load both webviews.
3. Switching back to the other webview.

screen shot 2017-05-30 at 7 40 34 pm

**Cause:**
This bug was introduced from the upstream. It is documented in [electron#5110](https://github.com/electron/electron/issues/5110). The bug has not yet been fixed but we have a workaround.

It's weird that I run into this bug for the first time tonight. Have you ever seen this before?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-// Place your settings in this file to overwrite default and user settings.
-{
-}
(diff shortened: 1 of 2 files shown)


akashnimare:
@geeeeeeeeek I have got this bug once. Merging. Thanks!

Issue #163, opened 2017-05-30
Use local icons for servers


zhongyi-tong:
- Load default icon from local.
 - Download the icon of entered zulip server and load it from there instead of loading from a URL.

zhongyi-tong:
@zulipbot claim

zulipbot:
Congratulations, @geeeeeeeeek, you've successfully claimed your first issue! Contributing to Zulip will (hopefully) be a fun and rewarding experience, and I'm here to help get you off to a good start.

First things first:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* Sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/), so that Zulip can use your code.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](http://zulip.readthedocs.io/en/latest/index.html#code-docs), as well as the rest of the developer documentation on that site.


See you on the other side (that is, the pull request side)!

akashnimare:
- [x] Load default icon from local.
 - [x] Download the icon of entered zulip server and load it from there instead of loading from a URL.

akashnimare:
Fixed via https://github.com/zulip/zulip-electron/commit/9ab75b980039e501a235117fc0b538b43aa88da4.

PR #166, opened 2017-06-01
Load realm settings from server_settings endpoint

zhongyi-tong:
According to #163, we should load server icon from the `/api/v1/server_settings` endpoint, or load default icon from local.

Endpoint:
image

Settings:
image


timabbott:
This only problem with this is that that endpoint is new in Zulip 1.6 (to be released this week), so we need to do something reasonable for backwards-compatibility...

zhongyi-tong:
@timabbott Fixed.

zhongyi-tong:
I'd like to implement this in the future, when we have a formal settings page. We need a more friendly add-server-page and management page.

PR #175, opened 2017-06-13
normalize whitespace

keturn:
Normally I wouldn't PR for a single-character whitespace change, but there's something run by zulip-electron-launcher that keeps adding this space back in every time, so it always thinks I have uncommitted changes in my working tree that are different from what's in origin/master...


akashnimare:
Thanks, @keturn 😄

PR #180, opened 2017-06-17
Refinements on the main process

zhongyi-tong:
screen shot 2017-06-18 at 2 16 45 am

 - Further refactoring of code on the main process.
 - Functional tabs are now closable.
 - Move About page into a WebView.


app/renderer/js/utils/domain-util.js
```diff
@@ -4,9 +4,17 @@ const {app} = require('electron').remote;
 const JsonDB = require('node-json-db');
 const request = require('request');
 
+let instance = null;
+
 const defaultIconUrl = 'https://chat.zulip.org/static/images/logo/zulip-icon-128x128.271d0f6a0ca2.png';
```
akashnimare: Let's download this icon and serve it locally (from `renderer/img`).

akashnimare (changes requested):

app/renderer/js/utils/link-util.js
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	isInternal(currentUrl, newUrl) {
+		const currentDomain = wurl('hostname', currentUrl);
+		const newDomain = wurl('hostname', newUrl);
+
+		const skipImages = '.jpg|.gif|.png|.jpeg|.JPG|.PNG';
+
+		// We'll be needing this to open images in default browser
+		return (currentDomain === newDomain) || newUrl.match(skipImages);
```
akashnimare: This should be -
```javascript
return (currentDomain === newDomain) && !newUrl.match(skipImages);
```
We don't want to open the image link within the app; it should be opened in the default browser.

akashnimare (changes requested):

akashnimare (approved):

akashnimare:
Merged. Thanks, @geeeeeeeeek.

Issue #188, opened 2017-06-23
close settings bug


akashnimare:
A user reported this - 

If you close the settings tab while on a different tab, any tab that isn't the first zulip server, then you get a split view of the first zulip server and the tab that you were on.

![close_settings_bug](https://user-images.githubusercontent.com/2263909/27501178-55c1597e-5889-11e7-8e01-bc3d76b4edfd.gif)

zhongyi-tong:
It happens on my machine.. I will send a fix soon.

akashnimare:
Fixed in https://github.com/zulip/zulip-electron/commit/cda8aa3b09ce15da60e1bde176eba641363fdefd.

PR #189, opened 2017-06-27
Fix close settings causing split webviews (#188).

zhongyi-tong:


akashnimare:
There is one other problem -

image

we should not allow creating multiple setting pages.

zhongyi-tong:
@akashnimare Fixed. This occurs only when there is no domains. I forgot to distinguish `0` and `undefined`.

akashnimare:
@geeeeeeeeek looks good. Thanks 😄

PR #190, opened 2017-06-28
Feature/settings page

zhongyi-tong:
c

Setting items we discussed before:
 - Show tray icon
 - Show/hide desktop notifications
 - Mute desktop notifications
 - Disable/enable spellchecker
 - Auto login

The framework of settings page is done. But not sure we can finish them all in 1.0.0.

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -215,6 +215,10 @@ app.on('ready', () => {
 		}
 		page.send('tray', messageCount);
 	});
+
+	ipc.on('forward', (event, listener) => {
+		page.send(listener);
+	});
 });
 
 app.on('will-quit', () => {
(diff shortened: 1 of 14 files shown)


timabbott:
Nice!  I think it's totally reasonable to merge an initial settings page with like 1 setting on the "General" screen and then plan to add the rest as their own issues, to avoid the framework getting blocked on figuring out all the settings we want.

For things like "show desktop notifications", do you think we want those to sync with the similar webapp/server setting, or not?  I think that's the main interesting question here :)

zhongyi-tong:
screen shot 2017-07-03 at 2 09 46 am
screen shot 2017-07-03 at 2 09 53 am
screen shot 2017-07-03 at 2 10 46 am

zhongyi-tong:
@timabbott For things like "show desktop notifications", I think it's different per server. We could have a overall toggle that turns on/off all notifications. If turned on, use server-specific preference. Otherwise, no notifications at all. 😀

akashnimare:
This looks great for now. Merging. Thanks, @geeeeeeeeek 🚀

PR #193, opened 2017-07-05
Make ConfigUtil runnable in both processes.

zhongyi-tong:


akashnimare:
That did the job. Thanks 🤝

PR #202, opened 2017-07-10
Save server icon && Provide keyboard shortcut to switch between servers.

zhongyi-tong:
screen shot 2017-07-11 at 12 44 32 am

 - Provide keyboard shortcut to switch between servers.
 - Save server icon to a local folder instead of fetching it everytime it loads in the front end.
 - Has some bugs for certain format of icons.

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -138,6 +138,23 @@ function createMainWindow() {
 	return win;
 }
 
+function registerLocalShortcuts(page) {
+	// TODO - use global shortcut instead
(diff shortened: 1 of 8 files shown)


app/main/index.js
```diff
@@ -203,6 +207,8 @@ app.on('ready', () => {
 	ipc.on('reload-main', () => {
 		page.reload();
 		page.send('destroytray');
+		electronLocalshortcut.unregisterAll(mainWindow);
+		registerLocalShortcuts(page);
```
akashnimare, line 211: Great. This fixes that shortcut un-registering bug.

PR #206, opened 2017-07-12
settings: Tweak text of initial screen to be a bit clearer.

gnprice:
The label "New Server" sounds to me like it's suggesting I create
a new actual server -- which I don't want to do, I want to connect
to an existing server where a community I know is talking.
Change the text to track a bit closer to what's really going on.

I think there's probably more that would be useful to do to
make this flow smoother for a new user, but this is a start.

### DIFF
--- a/app/renderer/js/pages/preference/servers-section.js
+++ b/app/renderer/js/pages/preference/servers-section.js
@@ -20,7 +20,7 @@ class ServersSection extends BaseComponent {
 				<div class="actions-container">
 					<div class="action green" id="new-server-action">
 						<i class="material-icons">add_box</i>
-						<span>New Server</span>
+						<span>Add Server</span>
 					</div>
 				</div>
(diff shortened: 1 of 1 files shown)


penCsharpener:
and what about "Connect a Server" (apart from being longer)

akashnimare:
That makes sense. Thanks, @gnprice.

gnprice:
Thanks @akashnimare for the fast response!

> and what about "Connect a Server" (apart from being longer)

Yeah, that'd feel kind of long to me for the text of a button. Certainly could tweak the text further.

I think there are several other things we could do, though, to simplify and clarify this flow significantly, and then there might not be a button quite like this in the first place. I think [this chat discussion](https://chat.zulip.org/#narrow/stream/electron/subject/adding.20servers.2Ed/near/248820) would be a good place to discuss, followed by issues/PRs for concrete things.

Issue #192, opened 2017-07-02
Add more options in settings page


akashnimare:
@geeeeeeeeek already added settings framework in - https://github.com/zulip/zulip-electron/commit/47cfe86a06d30bf09754d0f7739e741e9041de52. Few more settings we need to add -
- [x] Show/hide desktop notification
- [ ] <strike>Disable/enable spellchecking</strike>
- [x] Show app unread badge
- [ ] <strike> Show unread count in tray </strike>
- [x] Auto launch
- [x] Play notification sound
- [x] Clear cache/Reset Application Data
- [x] Enable beta updates

fragm3:
Working on: Show app unread badge

zulipbot:
Congratulations, @fragm3, you've successfully claimed your first issue! Contributing to Zulip will (hopefully) be a fun and rewarding experience, and I'm here to help get you off to a good start.

First things first:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* Sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/), so that Zulip can use your code.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](http://zulip.readthedocs.io/en/latest/index.html#code-docs), as well as the rest of the developer documentation on that site.


See you on the other side (that is, the pull request side)!

fragm3:
Hi, I have implemented Show unread count in tray, should I make a pull request now?

zhongyi-tong:
Great! PR is welcome!

fragm3:
Done with: Show unread count in tray, working on Disable/enable spellchecking and Show app unread badge now

fragm3:
Done with: Show app unread badge, would be moving forward with Disable/enable spellchecking and auto launch option

zulipbot:
Hello @fragm3, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for a week. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 3 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

zulipbot:
[...]

PR #226, opened 2017-07-20
Tray unread message toggle

fragm3:
Type: Enhancement
OS: MacOS
Issue: #192 

The toggle tray unread message is working fine and I have tested it.

### DIFF
--- a/app/renderer/js/pages/preference/general-section.js
+++ b/app/renderer/js/pages/preference/general-section.js
@@ -74,7 +74,7 @@ class GeneralSection extends BaseComponent {
 		this.settingsOptionTemplate(silentOption);
 	}
 
-	trayUnreadOptionTemplate(trayUnreadUpdateOption){
+	trayUnreadOptionTemplate(trayUnreadUpdateOption) {
 		this.settingsOptionTemplate(trayUnreadUpdateOption);
 	}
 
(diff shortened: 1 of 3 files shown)


zhongyi-tong:
Hi @fragm3, there is some linting errors you need to deal with before sending a PR. Please run `npm run test` on your machine and make sure no error appears. Thanks :)

fragm3:
Okayyy will do!

fragm3 (commented):
Did some linting, test passing!

app/renderer/js/pages/preference/general-section.js
```diff
@@ -120,6 +132,22 @@ class GeneralSection extends BaseComponent {
 		});
 	}
 
+	initTrayUnreadOption() {
+		this.$trayUnreadSettings = document.querySelector('#tray-unread-option-settings .setting-control');
+		this.$trayUnreadSettings.innerHTML = '';
+
+		const trayUnreadOption = ConfigUtil.getConfigItem('toggle-update-tray', 'true');
```
akashnimare, line 139: It should be just `true`. otherwise it would be invalid -
```json
 "toggle-update-tray": "true"
```

akashnimare:
@fragm3 does it work? We want to save those settings so when a user restarts/reload app it gets persisted. You're just toggling it. Same problem with #227.

akashnimare:
We don't need to add this setting because of https://github.com/zulip/zulip-electron/commit/3cd4890e60b892725281a0aca43063347eaba29d.

akashnimare:
This setting is not required anymore since we have removed this feature. Thanks for working on this.

PR #230, opened 2017-07-21
Hide/Toggle left sidebar

### LINKED ISSUE #222: Hide/Toggle left sidebar
(recreated in the correct project, per:  https://github.com/zulip/zulip/issues/5821)

Adding the left-side configuration bar results in the electron 1.2.0 client being exceedingly wide -- and the width can't be adjusted without hiding the user list.

Is it possible to make the width of the user list adjustable (streams also would be nice!), so it's possible to reduce the overall width of the client?

Somewhat related, it would also be a plus to be able to move or hide the left-side configuration bar -- it's very distracting, particularly when you've only got one server configured, and while it's nice to be able to configure the three options, it's confusing to have two (different) configuration sets.

zhongyi-tong:
c

Fix #222.


akashnimare:
Looks good. Can we set a better position of arrow? because we don't want to overlap app's content with arrow-

image
image

zhongyi-tong:
But where... I couldn't think of a better place. What about move it to the top? But it covers the content anyway...

akashnimare:
@brockwhittaker, @timabbott can you suggest a better position?

timabbott:
Well, when the left sidebar is visible, we could easily put a widget in the upper-left corner, inside the box containing the Zulip logo.  In smaller windows, it's a lot harder; that's probably a rare case for the desktop app but I could imagine having it cover over the top half of the hamburger menu?

I think this is a good topic to post on #design for ideas.  

One thing to keep in mind is that we could integrate this more tightly with the web code and offer support for more directly controlling what happens.

akashnimare:
Yeah, we can easily put it in the upper-left corner when the **width is > 776 px.** The problem comes when the window size is smaller than this.

zhongyi-tong:
New Design (with a config item and a menu item):
image

app/renderer/js/pages/preference/general-section.js
```diff
@@ -120,6 +132,22 @@ class GeneralSection extends BaseComponent {
 		});
 	}
 
+	initSidebarToggle() {
+		this.$sidebarOptionSettings = document.querySelector('#ui-option-settings #sidebar-option .setting-control');
+		this.$sidebarOptionSettings.innerHTML = '';
+
+		const sidebarOption = ConfigUtil.getConfigItem('show-sidebar', false);
```
akashnimare: It would be better if we show sidebar initially. If we hide it on the first load, users might not able to find it. What do you think?

akashnimare:
Merged. Thanks, @geeeeeeeeek.

app/renderer/js/pages/preference/general-section.js
akashnimare: @geeeeeeeeek something is wrong. No matter what I set here it's getting `null` initially.

PR #235, opened 2017-07-27
Add 'Hard Reload' mode for debugging.

zhongyi-tong:
screen shot 2017-07-27 at 1 13 28 pm


akashnimare:
We need to remove local shortcut i.e. [this](https://github.com/zulip/zulip-electron/blob/master/app/main/index.js#L148) since both are doing the same thing.

PR #236, opened 2017-07-27
Issue/fetch server details from zulip api

### LINKED ISSUE #214: Fetch server details using zulip api
image

* Make the `Name` field optional and fetch it from the server.

zhongyi-tong:
- Automatically fetch server details from `server_settings` API (if supported).
 - Make ServerName optional.
 - Rename server `Name` to `Label`.
Fix #214.

### DIFF
--- a/app/renderer/css/main.css
+++ b/app/renderer/css/main.css
@@ -70,8 +70,8 @@ html, body {
     align-items: center;
     padding: 10px;
 }
+
 .action-button:hover {
-    background: #eee;
     cursor: pointer;
 }
 
--- a/app/renderer/css/preference.css
+++ b/app/renderer/css/preference.css
@@ -106,11 +106,9 @@ body {
 }
 
 img.server-info-icon {
-    background: #a4d3c4;
-    border-radius: 4px;
-    width: 28px;
-    height: 28px;
-    padding: 8px;
+    width: 36px;
+    height: 36px;
+    padding: 4px;
 }
 
 .server-info-left {
@@ -124,19 +122,14 @@ img.server-info-icon {
 
 .server-info-row {
     display: flex;
-    line-height: 27px;
-    margin: 8px 0;
-    height: 27px;
+    margin: 8px 0 0 0;
 }
 
-.server-info-key {
-    width: 40px;
-    margin-right: 20px;
+.server-info-alias {
     font-weight: bold;
-    text-align: right;
 }
 
-.server-info-value {
+.server-info-url {
     flex-grow: 1;
     font-size: 14px;
     height: 24px;
@@ -165,6 +158,7 @@ img.server-info-icon {
     align-items: center;
     padding: 0 10px;
     border-radius: 2px;
+    margin-right: 10px;
 }
 
 .action i {
--- a/app/renderer/css/preference.css
+++ b/app/renderer/css/preference.css
@@ -194,11 +194,8 @@ img.server-info-icon {
 }
 
 .hidden {
-    height: 0 !important;
-    width: 0 !important;
+    display: none;
     margin: 5px !important;
-    opacity: 0 !important;
-    transition: opacity 0.3s;
 }
 
 .red {
--- a/app/renderer/js/pages/preference/new-server-form.js
+++ b/app/renderer/js/pages/preference/new-server-form.js
@@ -17,19 +17,9 @@ class NewServerForm extends BaseComponent {
 				</div>
 				<div class="server-info-right">
 					<div class="server-info-row">
-						<span class="server-info-key">Url</span>
-						<input class="server-info-value" placeholder="(Required)"/>
+						<input class="server-info-url" placeholder="Enter the url of your Zulip server..."/>
 					</div>
 					<div class="server-info-row">
-						<span class="server-info-key">Label</span>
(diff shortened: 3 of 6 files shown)


timabbott:
Does this result in users just needing to enter the URL in a single-field for a server when first configuring it?  I think we want that model, and if you want to add a label or something, that can be a follow-up editing phase.

zhongyi-tong:
Yes. URL is the only field required.

timabbott:
Can you post a quick screenshot?  I think showing the other fields when adding a new server, even as optional, is confusing, so I want to make sure we're doing that correctly :)

zhongyi-tong:
image

@timabbott The `server_settings` endpoint is only supported when Zulip version > 1.6. So if we only show the URL field, we are not able to get the name and icon.

What if I collapse these fields with a ∇  button?

timabbott:
I think Zulip <1.6 is going to be pretty rare these days.  So I think what we should do is this:
* Just ask for a URL.   if the server supports `server_settings`, that's all you need.
* If the server is old, maybe have a second stage where we print that they're running an old server version and they need to get the server upgraded, or provide a name and icon?

zhongyi-tong:
Sounds good :)

akashnimare:
> I think Zulip <1.6 is going to be pretty rare these days. 

@timabbott if this is the case then we should only ask for the URL and then fetch all the details (icon, realm_name etc) from the server API. Also, we should validate server URL against the `"result":"success"` from the server API instead of checking the `zulip.ogg` audio file.

zhongyi-tong:
@akashnimare Then obsolete versions can not be validated, which might affect some users. If we just ask for a URL, we can improve the user experience without affecting those users. What do you think?

akashnimare:
I think If we have very small numbers of users who are using Zulip < 1.6 then we should drop the `zulip.ogg` model and encourage those users to use Zulip v1.6 since it's much better than the older versions. @timabbott do we have any counts of those users who are using Zulip < 1.6 ? The other option is just to wait for Zulip 1.7 and then drop the support for `zulip.ogg` model.

timabbott:
We don't have concrete data, but one important detail is that Zulip <1.5 links to the old app anyway.  Still, I think it's worth having it not be impossible to use the app with older Zulip.  

But it can be imperfect; e.g. , we could just use the server's domain as the "name" and a randomly generated gravatar for the default icon.  

I think probably we don't need to support the user customizing the name/icon fields at all; those should just be pulled from the server when connecting (and cached locally so you can display them before connecting to the realm)

Actually, that's probably a better plan than what I'd proposed before, since then we don't have to have a settings UI for setting those fields at all; we just can show a list of the servers (displayed nicely with their name and avatar) and just have options to add/remove servers from the list.  Cleaner and less code.

akashnimare:
@geeeeeeeeek consider a case -
* User enter https://chat.zulip.org
* Set an icon - https://chat.fhir.org/user_avatars/2/realm/icon.png
* relam_icon is being set as server icon instead of above icon.

Fallback for server icon should be like this -

`Entered Icon > realm_icon from api > default icon`

zhongyi-tong:
Yep. The priority of icon is now exactly like what you said :)

akashnimare:
> Yep. The priority of icon is now exactly like what you said :)

@geeeeeeeeek no it isn't. Please check again. As of now, priority look like this -

`realm_icon from api > entered icon > default icon`

zhongyi-tong:
According to @timabbott's suggestion, simplified the UI. There are no more customized fields for Icon and Name, which are either fetched from API or filled with the default icon and domain name.

screen shot 2017-07-31 at 3 47 15 pm

akashnimare:
@geeeeeeeeek looks like server alias is clickable and triggers addDomain event.


nimbus-record-video-2017-07-31-16-08-44

zhongyi-tong:
I see. It overlaps with the hidden New Server Form. Let me see how to fix it.

zhongyi-tong:
@akashnimare Fixed. Have a try :)

akashnimare:
Yup, it's working properly. PR looks good, I'll cross test it once again tonight and then merge it :)

akashnimare:
Merged. Thanks, @geeeeeeeeek.

akashnimare:
> I think If we have very small numbers of users who are using Zulip < 1.6 then we should drop the zulip.ogg model and encourage those users to use Zulip v1.6 since it's much better than the older versions. @timabbott do we have any counts of those users who are using Zulip < 1.6 ? The other option is just to wait for Zulip 1.7 and then drop the support for zulip.ogg model.

@timabbott is this a right time to drop the support for `zulip.ogg` model? Our current logic generates a false alarm on `*zulipchat.com` servers since the `zulip.ogg` file is present on invalid/non-existent servers also. So for an example - app validates the server like `https://plan.zulipchat.com` (since that ogg file is there).

timabbott:

```
```

And that's very different from the 404 that might mean we should check for `zulip.ogg`.  

It's potentially reasonable to start requiring `/server_settings` at this point; we've had it since 1.6.0 last June, and I'd be surprised if there are many Zulip servers older than that.  But we'd need to be pretty careful about providing a good error message, so that users' experience isn't just "the desktop app auto-updated to broken for everyone", and that's probably more work than fixing this check correctly.

PR #238, opened 2017-07-31
Send destroy tray after hard reload.

zhongyi-tong:
Per the comment in #235, seems that the hard-reload shortcut is not working from the menu item. We need to keep the `electronLocalshortcut`.

This PR adds the missing `destroytray` after `hard-reload`.

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -144,10 +144,13 @@ function registerLocalShortcuts(page) {
 	electronLocalshortcut.register(mainWindow, 'CommandOrControl+R', () => {
 		page.send('reload-viewer');
 	});
+
 	// Reload full app not just webview, useful in debugging
 	electronLocalshortcut.register(mainWindow, 'CommandOrControl+Shift+R', () => {
 		mainWindow.reload();
+		page.send('destroytray');		
 	});
+
 	// Also adding these shortcuts because some users might want to use it instead of CMD/Left-Right
 	electronLocalshortcut.register(mainWindow, 'CommandOrControl+[', () => {
 		page.send('back');


akashnimare:
It does work if we send the `hard-reload` event to `main process` via `webview`. As of now, we are directly sending it to main process. This is how it could work -

in `app/renderer/main.js`
```javascript
		ipcRenderer.on('hard-reload', () => {
			ipcRenderer.send('full-app-reload')
		});
````
in `app/main/index.js`
```javascript
	ipc.on('full-app-reload', () => {
		mainWindow.reload();
                page.send('destroytray');
	});

akashnimare:
@geeeeeeeeek I have made it better in https://github.com/zulip/zulip-electron/commit/2c8cf2b959dc7cef1bd53f2e2dbc31a7df470188. Hard reload works great now. While I was fixing this I have noticed that `reload` shortcut (<kbd>CMD+R </kbd>) does not reload running server. It always redirects to the first server. Clicking on `Menu > Reload` and the `Reload icon` in left side bar works as expected but not <kbd>CMD+R</kbd>. You can reproduce this -
* Add two servers
* Open second server 
* <kbd>CMD+R </kbd>. It redirects to the first server instead of reloading the second server.

zhongyi-tong:
The reload destroys the whole viewer and reinitializes it. And the first
server shows up by default. Do you think we need to remember the last
opened tab and restore it?

akashnimare:
@geeeeeeeeek Yeah, that would be great. We need to work on reload event properly. It needs few improvements, such as -
* Reload the setting page when the user deletes a server. As of now, it redirects to the first server.  
* Same thing when user adds a new server.

PR #241, opened 2017-08-01
Add an os check for webview focus events.

### LINKED ISSUE #216: On activating Window 2 click needed to write into compose box [Linux/Windows]
when the Zulip desktop app is inactive and I click into the compose box with the mouse then it only activates the Windows. I have to click a second time to get the cursor into the compose box and start typing. 

```quote
Tim Abbott: Hmm, this sounds like a window focusing issue with the multi-organization support...
```

zhongyi-tong:
Fix #216.

**macOS:**

**Windows:**

### DIFF
--- a/app/renderer/js/components/webview.js
+++ b/app/renderer/js/components/webview.js
@@ -104,7 +104,7 @@ class WebView extends BaseComponent {
 		// The bug is introduced from Electron and this is a tempory fix.
 		// See https://github.com/zulip/zulip-electron/issues/216
 		const osName = SystemUtil.getOS();
-		if (osName === 'Mac' || osName === 'Linux') {
+		if (osName === 'Mac') {
 			this.$el.focus();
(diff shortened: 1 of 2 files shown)


app/renderer/js/components/webview.js
```diff
@@ -100,7 +100,13 @@ class WebView extends BaseComponent {
 	}
 
 	focus() {
-		this.$el.focus();
+		// Explicit focusing the webview causes a bug the input cannot regain focus on Winodws.
+		// The bug is introduced from Electron and this is a tempory fix.
+		// See https://github.com/zulip/zulip-electron/issues/216
+		const osName = SystemUtil.getOS();
+		if (osName === 'Mac' || osName === 'Linux') {
```
akashnimare: Well, I was wrong. Focus event only required on macOS.

PR #252, opened 2017-08-01
Clarify and expand the developer docs a bit

gnprice:
From discussion [in #electron](https://chat.zulip.org/#narrow/stream/electron/subject/setup/near/260866).

### DIFF
--- a/development.md
+++ b/development.md
@@ -43,30 +43,39 @@ $ cd zulip-electron
 ```
 
 Install project dependencies:
-
 ```sh
 $ npm install
 ```
-Start the app:
 
+Start the app:
 ```sh
 $ npm start
 ```
 
-Start and watch changes
-
+Start and watch changes:
 ```sh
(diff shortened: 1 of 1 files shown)


development.md
```diff
@@ -1,56 +1,84 @@
...
+This command will produce distributable packages or installers for the
+operating system you're running on:
+* on Windows, a Windows installer file
+* on macOS, a `.dmg` file
+* on Linux, a plain `.zip` file as well as a `.deb` file and an
+  `AppImage` file.
+To generate all three types, you will need all three operating
+systems.
 
-# Troubleshooting
```
akashnimare, line 55: We should keep the `Troubleshooting` section. @gnprice what do you think?

akashnimare: I see you have moved it up 😅

akashnimare:
Looks good. Merged. Thanks, @gnprice 🚀

PR #260, opened 2017-08-02
add instruction for Windows setup

penCsharpener:
it might be that you only need the second powershell command but it worked fine after I run both. The `git init` might seem too obvious for some but it stalled the process and turned a 20 min process into almost 1 hr trying other things and googling.


penCsharpener:
the other thing I didn't include here is that I always run `npm run pack` before `npm run dist`. Never tried just `npm run dist` on it's own but might try another time.

PR #265, opened 2017-08-06
Feature/switch servers in menu

### LINKED ISSUE #245: No easy way to select a server after hidding the sidebar
If you hide the server switch sidebar, on the settings page, there's no obvious way to get back to seeing your single Zulip server's UI.  It turns out you can do it with `ctrl-1`, but that's really not obvious.

Some ideas for fixing:
* The "Servers" section of the settings page should let you jump to that server, maybe be clicking on the icon or something.
* We should list the servers you have setup, with hotkeys in a "servers" menu next to the file menu?
* We could list the servers in the taskbar menu.

Doing something here is a priority, since this is basically the first  thing that anyone is going to try to do after toggling the setting to hide the sidebar.

zhongyi-tong:
image

Fix #245.

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -153,7 +153,9 @@ app.on('activate', () => {
 });
 
 app.on('ready', () => {
-	appMenu.setMenu();
+	appMenu.setMenu({
+		tabs: []
+	});
 	mainWindow = createMainWindow();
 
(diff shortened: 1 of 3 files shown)


akashnimare:
@geeeeeeeeek Thanks for working on this 🚀

PR #272, opened 2017-08-13
Use ES6 destructuring for grabbing elements

### LINKED ISSUE #271: Use of ES6 Destructuring in index.js
Can we possibly use ES6 (ES2015) destructuring in [index.js](https://github.com/zulip/zulip-electron/blob/master/app/main/index.js) for : 

>const {app} = require('electron');
const ipc = require('electron').ipcMain;

which will prevent requiring electron multiple times as we have already done once at the start, thoughts on this suggestion ?

anshumanv:
OS : W10 X64
Zulip Version : 1.3.0-beta
Fixes #271

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -1,12 +1,13 @@
 'use strict';
 const path = require('path');
 const electron = require('electron');
-const {app, ipcMain} = electron;
 const electronLocalshortcut = require('electron-localshortcut');
 const isDev = require('electron-is-dev');
 const windowStateKeeper = require('electron-window-state');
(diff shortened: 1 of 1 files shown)


anshumanv:
Please help me here, I hope there are no discrepancies in changing `ipc` to `ipcMain` in using destructuring.

anshumanv:
Thanks for the fix @akashnimare

akashnimare:
I don't think there would be any problem. Merging. Thanks, @Anshuman-Verma.

PR #273, opened 2017-08-15
Issue/new server icon cache

zhongyi-tong:
1. Use the hash mechanism to generate local icon filenames. As a result, when you try to save an existed icon, your user-data folder won't be filled with duplicated files.
2. Update `icon` and `name` when app reloads. It updates lazily so we are still able to utilize the cache capability to display icons, but if a user's `server_settings` has updated, the new icon will show on the next reload.

### DIFF
--- a/app/renderer/js/main.js
+++ b/app/renderer/js/main.js
@@ -49,6 +49,7 @@ class ServerManagerView {
 		if (servers.length > 0) {
 			for (let i = 0; i < servers.length; i++) {
 				this.initServer(servers[i], i);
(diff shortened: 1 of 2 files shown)


akashnimare:
@geeeeeeeeek `duplicateDomain` function doesn't seem to work. Can you check what's wrong?

zhongyi-tong:
@akashnimare sure, right now.

zhongyi-tong:
@akashnimare Fixed in the latest commit.

akashnimare:
Works great. Thanks, @geeeeeeeeek.

PR #278, opened 2017-08-21
Feature/connect through proxy

### LINKED ISSUE #198: Add a configuration option for connecting through an HTTP proxy
Some users want to connect from a machine in one network to a zulip server on another network where direct connections are not possible and all traffic must go via an HTTP proxy.
e.g. client on a corporate network, server on the public internet.

Transparent / intercepting proxies require no configuration on the client, but lots of organizations still use traditional HTTP proxies. The CONNECT method allows establishing a TCP tunnel for HTTPS connections.

Electron appears to support this via the [Session.setProxy](https://electron.atom.io/docs/api/session/#sessetproxyconfig-callback) API.

The best behaviour from a user's POV is to fetch proxy settings from the operating system.
* On Windows, this will be somewhere in the registry (configurable through control panel / internet explorer; or may be pushed centrally for enterprise versions)

The second best option is a configuration option in the app's settings.
This will be equivalent to configuring Firefox, and the app won't work "out of the box" but the configuration option will be easy to find by looking in the settings.

The "minimum viable product" would be a command line switch or environment variable; which makes connection possible for expert users.

The electron API offers both `pacScript` and `proxyRules`.
A PAC (proxy auto configuration) script is a javascript document usually provided as a URL; it tells the browser on a per-URL basis which proxy to use (or 'DIRECT' for none).
The PAC script is the most likely configuration in a corporate environment
`proxyRules` gives the same capabilities as a table.
Most browsers have a single proxy server, and a list of exceptions when using static configuration.

For testing, 'squid' is a relatively easy to configure HTTP proxy available on many Linux distros.

zhongyi-tong:
screen shot 2017-08-20 at 11 05 22 pm

 - Sort and merge sections in the General Section of Settings.
 - Add a Network Section, which enables users to input their proxy settings.
 - Allow users to connect to servers through a proxy. Fix #198.
 - Refactor the code structure of Setting pages.
 - Initialize the user guide `help.md`.

@akashnimare Would you please cross-check the feature?


akashnimare:
This looks great. I'll cross check it tonight.

app/renderer/js/main.js
```diff
@@ -282,6 +305,12 @@ class ServerManagerView {
 			this.activateTab(index);
 		});
 
+		ipcRenderer.on('reload-proxy', () => {
+			this.loadProxy().then(() => {
+				alert('Proxy settings saved!');
```
akashnimare: Do we need to show this alert when the user turns `off` the proxy option? How about this -

``` javascript
ipcRenderer.send('forward-message', 'toggle-sidebar', newValue);
```

```javascript
ipcRenderer.on('reload-proxy', (event, newValue) => {
	this.loadProxy().then(() => {
		if (newValue !== false) {
			alert('Proxy settings saved!');
		}
	});
});

**Comment by shane-kearns** [comment]
I've tested this on Linux and the proxy settings are working.
The examples in the user guide and UI placeholder text don't include `https=proxy.foocorp` which would be needed for connecting to a zulip server when a http proxy must be used for all access (This works, and uses the proxy CONNECT method as expected).
Proxy PAC url worked as expected.

File: app/renderer/js/pages/preference/network-section.js [hunk 2]
```diff
@@ -0,0 +1,102 @@
...
+		this.generateSettingOption({
+			$element: document.querySelector('#use-proxy-option .setting-control'),
+			value: ConfigUtil.getConfigItem('useProxy', false),
+			clickHandler: () => {
+				const newValue = !ConfigUtil.getConfigItem('useProxy');
+				ConfigUtil.setConfigItem('useProxy', newValue);
+				this.toggleProxySettings(newValue);
+				if (newValue === false) {
+					// Reload proxy if the proxy is turned off
+					ipcRenderer.send('forward-message', 'reload-proxy');
```
zhongyi-tong: @akashnimare Here, when the user turns off the proxy, it sends `reload-proxy` and will trigger an alert.

zhongyi-tong:
@shane-kearns Actually that section in the user guide is copied from Electron. I don't know if others rule may apply...

app/renderer/js/pages/preference/network-section.js
akashnimare: Yeah, but I think we don't need to trigger that alert when user turns off the proxy.

p

zhongyi-tong: Oh, I see.

app/renderer/js/main.js
akashnimare: @geeeeeeeeek ^^^ this is what we need to use 😄

akashnimare:
@shane-kearns thanks for testing. Is there anything else we need to add/remove in this?

help.md
```diff
@@ -0,0 +1,94 @@
...
+
+It's possible to connect to your server through a proxy. 
+You can enter the proxy settings in the `Network` section of App Settings. 
+There are three fields provided:
+* `PAC script` - The URL associated with the PAC file.
+* `Proxy rules` - Rules indicating which proxies to use.
+* `Proxy bypass rules` - Rules indicating which URLs should
+    bypass the proxy settings.
+
+Sets the proxy settings.
```
shane-kearns: I would add a simple example here, which will work for typical setups:

For a typical setup where internet access is required to use an HTTP proxy,
but URLs on the local network should be accessed directly, configure as follows:

`Proxy rules` = `proxy.example.com`
 - Your HTTP proxy server

`Proxy bypass rules` = `*.example.com;10.0.0.0/8`
 - Directly connect to your own domain and private IP subnet

For more complex setups, read below to configure complex proxy rules and proxy bypass rules.

akashnimare:
Thanks, @geeeeeeeeek, @shane-kearns 👍

PR #313, opened 2017-10-07
Tooltip for add server button

YJDave:
Add tooltip "**Add organization**" for symbol " **+** " in left-side tab, which is used to add/modify server(organization) url.

tooltip


akashnimare:
Looks good. Merged. Thanks, @YJDave 🚀

PR #322, opened 2017-10-18
Spellchecker Improvements 

akashnimare:
This PR adds couple of functionalities in Spellchecker -
* On macOS, the language of the Spellchecker will be equal to the language of the server. That means if you have set  `Deutsch` as the server language then the spellchecker will check spelling against the same. 
* On Linux/Windows, Spellchecker can automatically switches to the language user is typing in, so we don't need to set it explicitly. Spellchecker work as it is on these platforms.
* There is an option in general settings to disable/enable the spellchecker.

image

This PR also fixes a bug which was caused by not setting the default settings properly.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -31,7 +31,7 @@
     "electron-is-dev": "0.3.0",
     "electron-localshortcut": "2.0.2",
     "electron-log": "2.2.7",
-    "electron-spellchecker": "1.2.0",
+    "electron-spellchecker": "1.1.2",
     "electron-updater": "2.8.9",
     "node-json-db": "0.7.3",
     "request": "2.81.0",
--- a/app/renderer/js/pages/preference/general-section.js
+++ b/app/renderer/js/pages/preference/general-section.js
@@ -57,6 +57,10 @@ class GeneralSection extends BaseSection {
 						<div class="setting-description">Start app at login</div>
 						<div class="setting-control"></div>
 					</div>
+					<div class="setting-row" id="enable-spellchecker-option">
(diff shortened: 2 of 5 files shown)


akashnimare:
@cedricium can you cross check if everything works fine? Specially these things - 
* Enable/disable setting option for spellchecker (make sure you restart app).
* On Linux/Windows, Spellchecker should automatically switches to the language user is typing in.
* On macOS, it should spellcheck against the default server language. 
* Other settings options (make sure they work as expected).

cedricium:
@akashnimare will check it out later today and let you know how everything goes.

cedricium:
@akashnimare looks great! On Windows, works just like you said - despite the server/user language, spellcheck will pick up on the typed language and make suggestions that way (don't have Linux so not able to test that specifically).

The same goes for Mac too. Turning on and off spellcheck was easy enough. Just one note: it appears that spellcheck depends on the **user's language preference**, not the server's language. If the server had a language set to Spanish and the user had English, spellcheck would not recognize Spanish words. Switching the user's language would make spellcheck work as intended.

akashnimare:
@cedricium thanks for testing it out. From server language I meant the **user's language preference**.

akashnimare:
> it appears that spellcheck depends on the user's language preference, not the server's language. If the server had a language set to Spanish and the user had English, spellcheck would not recognize Spanish words. Switching the user's language would make spellcheck work as intended.

I think spellchecker should depends on user's language preference only. That way they can control spellchecker. Normal users don't have any control over org language settings thus they can't choose spellchecker language.

PR #326, opened 2017-10-29
Remove python version file

SimplyAhmazing:
This PR fixes the app build breaking on Travis Linux on the master branch. The reason the build is failing on master is because the repo contains a `.python-version` file which is used by [pyenv](https://github.com/pyenv/pyenv) to automatically set the python version. I used this file locally to manage my python 2/3 environments for various projects.

Recently Travis started rolling out `pyenv` as a part of their build images and repos that contained this file would cause `pyenv` on travis to use the python version specified by the `.python-version` file. Removing this file from the repo allows Travis `pyenv` to not try to use it and thus the build ends up defaulting to the default python which allows `node-gyp` to run without errors.


akashnimare:
@SimplyAhmazing 🙏  I was wondering why the tests are failing on Travis Linux. Thanks 🚀

Issue #331, opened 2017-11-08
Write dev tools log to a file 


akashnimare:

priyank-p:
@zulipbot claim.

zulipbot:
Welcome to Zulip, @cPhost! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-electron/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* Sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/), so that Zulip can use your code.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

PR #345, opened 2017-12-02
logger: add console helper to log for both file and console

priyank-p:
Features 
Custom TImestamp, no timestamp or no timestamp
File to log out 

Fixes: https://github.com/zulip/zulip-electron/issues/331

### DIFF
--- a/app/renderer/js/console.js
+++ b/app/renderer/js/console.js
@@ -1,0 +1,69 @@
+const NodeConsole = require('console').Console;
+const fs = require('fs');
+
+const { app } = require('electron').remote;
+const isDev = require('electron-is-dev');
+
+const browserConsole = console;
+const logDir = `${app.getPath('userData')}/Logs`;
+if (!fs.existsSync(logDir)) {
+  fs.mkdirSync(logDir);
+}
+
+function customConsole(opts, type, ...args) {
+  const { nodeConsole, timestamp } = opts;
+  if (timestamp) {
+    args.unshift(timestamp());
+  }
+
+  if (!isDev) {
+    const nodeConsoleLog = nodeConsole[type] || nodeConsole.log;
+    nodeConsoleLog.apply(null, args);
+  }
+  browserConsole[type].apply(null, args);
+}
+
+function getTimestamp() {
+  const date = new Date();
+  const timestamp =
+    `${date.getMonth()}/${date.getDate()} ` +
+    `${date.getMinutes()}:${date.getSeconds()}`;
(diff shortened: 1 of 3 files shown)


priyank-p:
I have no idea how to pass the linter tests it says use tab instead of spaces but whenever is press tab my vscode editor add two spaces any reason

priyank-p:
One more point `package-lock.json` will probably reduce the chances most common errors where we have to delete node_modules and reinstall them and faster `npm install`

app/renderer/js/console.js
```diff
@@ -0,0 +1,69 @@
+const NodeConsole = require('console').Console;
```
akashnimare, line 1: Replace `NodeConsole` with `nodeConsole` or something else since we usually follow **camelCase** for identifier names (variables and functions).

akashnimare (changes requested):

app/renderer/js/console.js
priyank-p: That would be a constructor that we later use to create `nodeConsole` by passing writable stream

akashnimare:
@cPhost so this doesn't log anything by default, right? Can you give us an example of how we could use this in main or renderer process?

priyank-p:
@akashnimare i am posting example of it in zulip chat.

app/renderer/js/console.js
cPhost: `NodeConsole` is a constructor that we later use by passing `WriteableStream` of file. Are you sure you want this capitalized because later we use it as `new NodeConsole()`.

priyank-p: The linter also requires `NodeConsole` to be like as it is now.

akashnimare: Hmm, make sense.

akashnimare (approved):

akashnimare:
Tried it locally, works great. BTW do we want to add the `.log` extension to the filename?

priyank-p:
By default it would `console.log` we do not need to add `.log` extension it a convention that all app follows specifically npm `npm_debug.log` but we do not need to since it for use to debug and a text file. so should i remove the `.log` from default.

jajodiaraghav:
@akashnimare Can this task be approved in the GCI app. Asking since you have approved the PR but haven't merged yet.

akashnimare:
@cPhost no, it's fine.

akashnimare:
Merged, thanks @cPhost. This needs a bit of code-refactoring though.

PR #349, opened 2017-12-06
linux: Improve debian installer

akashnimare:
This adds a functionality to remove all the app's settings, config etc when user uninstalls the application.


andersk (commented):
Debian `postrm` scripts also run on upgrades, so this was causing settings to disappear during upgrades (#1079)! I removed this code in v5.6.0 (109795ca3e6b247070712f60bbc62aeaa45deb84).

PR #350, opened 2017-12-06
[WIP] Setup karma unit tests

SimplyAhmazing:
This PR,

* Sets up [karma](https://github.com/karma-runner/karma) for executing renderer process tests in electron
* Adds one test file for testing spellchecker
* Uses [jasmine](https://jasmine.github.io/) as the testing framework for renderer process unit tests
* Refactors tests dir into `tests/e2e` & `tests/unit`


akashnimare:
So unit test which basically test spellchecker works as expected on macOS but on Linux and Windows it's getting failed. The spellchecker test is pretty important and we should make it work before merging this PR!

SimplyAhmazing:
I'm not sure why the tests is failing on CI but I have the following  ideas,

1. What if there is something wrong with [electron-spellchecker](https://github.com/electron-userland/electron-spellchecker). The signal Desktop app uses [node-spellchecker](https://github.com/atom/node-spellchecker). The `electron-spellchecker` module doesn't have any CI build scripts. `node-spellchecker` does have build scripts. Could it be that one is more reliable / consistent than the other?

2. Since the `spellchecker` tests passes locally on our machines, could it be that `electron-spellchecker` doesn't have the appropriate dictionaries available when running on the CI machines? Looking at the [Signal Desktop](https://github.com/WhisperSystems/Signal-Desktop/blob/development/package.json#L112) setup, they seem to handle the shipping of dictionaries when packaging the app.

The documentation for `electron-spellchecker` says that,

> Spell checks in all of the languages that Google Chrome supports by reusing its dictionaries.

I'm not sure if this means the package uses the dictionaries that are shipped with the `Chrome` browser bundled by `Electron` or if they use the dictionary of the users installed `Chrome` browser. If `electron-spellchecker` relies on the users installed `Chrome` browser dictionaries then maybe because the `CI`s don't have Chrome installed the package tests fail on CI but they pass locally on our machines?

Let's raise an issue with [electron-spellchecker](https://github.com/electron-userland/electron-spellchecker) and point them to our failing tests and ask why they don't have a CI setup or if they do how they got it to work.

SimplyAhmazing:
Regarding point 1 from my previous comment, there's an open issue with `electron-spellchecker` not working on windows here: https://github.com/electron-userland/electron-spellchecker/issues/93

priyank-p:
Superseded by #350 closing.

Issue #310, opened 2017-10-03
Handle corrupted domain.json files gracefully


timabbott:
We got a report of a user having the Zulip desktop app just hang on startup.

It turned out the cause was their `domain.json` file was totally corrupted JSON and didn't open (it was filled with `\0` characters).

We should probably do some combination of the following.
(0) See if we have any bugs that might cause this corruption (e.g. if we open the file for writing before doing a network request and only after the network request returns, write to the file).
(1) Quickly look into whether there's a standard Electron storage library we should be using that does this sort of error-handling for us.  That might be the best plan.
(2) If not, make the code that reads `domain.json` handle errors in a reasonable way.  We probably want to show some sort of user-facing error like "The Zulip app's configuration file (path) seems to be corrupted; this is very rare and likely caused by problems with your filesystem.." with a button "Clear state" or something.  And we should log something to e.g. Sentry so we can track how often it happens.

akashnimare:
@geeeeeeeeek is it possible that reloading the DB while updating the saved server causing this issue. Do we really need to reload db here - 
https://github.com/zulip/zulip-electron/blob/0a1866abb58a1bba4200e790161eb357b39ac3e9/app/renderer/js/utils/domain-util.js#L49
https://github.com/zulip/zulip-electron/blob/0a1866abb58a1bba4200e790161eb357b39ac3e9/app/renderer/js/utils/domain-util.js#L218
https://github.com/zulip/zulip-electron/blob/master/app/renderer/js/utils/domain-util.js#L212-L221

akashnimare:
Short term solution - If `domain.json` or `setting.json` gets corrupted for some reason then re-create both the files so that users don't get the hanging app window.

Long term solution - Find the root cause of why the files are getting corrupted and fix that.

akashnimare:
Made some changes in error handling of corrupted config files. From now on, if the config files (settings.json, domain.json) gets corrupted for some reason then we re-initialize those files that way user won't get hanging app window. For an example if the domain.json gets corrupted -
* First we're showing the error dialog -

![image](https://user-images.githubusercontent.com/2263909/34342361-dbaab20c-e9d3-11e7-8483-9c10d43ead58.png)

* Deletes and re-init domain.json. Also logging the errors for better debugging.
FYI @timabbott.

timabbott:
"organization" is spelled incorrectly as "organisation" in one of the places there.  Also, "readd" isn't a word.  

@rishig can you help with the language here?

[...]

PR #351, opened 2017-12-15
handle corrupted domain.json

priyank-p:
Currently it just deletes the corrupted files if has `\0` which can't be parsed by `JSON.parse` until long term solution is implemented.

Refs: https://github.com/zulip/zulip-electron/issues/310

### DIFF
--- a/app/renderer/js/utils/domain-util.js
+++ b/app/renderer/js/utils/domain-util.js
@@ -5,6 +5,11 @@ const fs = require('fs');
 const path = require('path');
 const JsonDB = require('node-json-db');
 const request = require('request');
+const Console = require('../console');
+
+const console = new Console({
+	file: `domain-util.log`
+});
 
 let instance = null;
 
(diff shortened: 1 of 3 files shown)


app/renderer/js/utils/domain-util.js
```diff
@@ -225,7 +225,13 @@ class DomainUtil {
 	}
 
 	reloadDB() {
-		this.db = new JsonDB(app.getPath('userData') + '/domain.json', true, true);
+		const DOMAIN_JSON_PATH = path.join(app.getPath('userData'), '/domain.json');
```
akashnimare: camelcase please i.e. `domainJsonPath`. Also, we need to do the same thing for `settings.json`, [here](https://github.com/zulip/zulip-electron/blob/master/app/renderer/js/utils/config-util.js#L50).

akashnimare (changes requested):

app/renderer/js/utils/domain-util.js
```diff
@@ -225,7 +225,14 @@ class DomainUtil {
 	}
 
 	reloadDB() {
-		this.db = new JsonDB(app.getPath('userData') + '/domain.json', true, true);
+		const domainJsonPath = path.join(app.getPath('userData'), '/domain.json');
+		try {
+			const file = fs.readFileSync(domainJsonPath, 'utf8');
+			JSON.parse(file);
+		} catch (err) {
+			fs.unlinkSync(domainJsonPath);
```
akashnimare: @cPhost I think we should catch the err here and log using the console.js to make the debugging easier.

akashnimare (approved):

akashnimare:
Looks good, just need to take care of the failing travis tests.

akashnimare:
> Having some trouble with tests, the tests are failing for no reasons, When i run the test locally it says it try to make the log directory when it already exists and therefore fails. 
pasted image
Everything works file when you manually start the app, evenafter deleting the logs directory.

Hmm, the problem is that the test dir which is `ZulipTest` doesn't have the `Logs` directory. We use `/Users/USERNAME/Library/Application Support/ZulipTest` dir instead of `/Users/USERNAME/Library/Application Support/Zulip` so that the e2e tests data don't get mixed up with the Zulip app data. I think you probably need to make the `Logs` dir in the test config [here](https://github.com/zulip/zulip-electron/blob/master/tests/setup.js).
cc @SimplyAhmazing.

akashnimare:
@cPhost thanks for working on this. Merged in https://github.com/zulip/zulip-electron/commit/04083bfa8189f88c006bdc41ab2ce778b39f2760.

PR #353, opened 2017-12-19
default-util: Fix log dir errors.

priyank-p:


akashnimare:
I don't think it has to do with main vs remote app process. Getting the same error.

app/renderer/js/utils/domain-util.js
```diff
@@ -7,7 +7,7 @@ const JsonDB = require('node-json-db');
 const request = require('request');
 const Console = require('../console');
 
-const console = new Console({
+const logger = new Console({
 	file: `domain-util.log`
```
akashnimare: @cPhost let's add the timestamp here.

app/renderer/js/utils/default-util.js
```diff
@@ -1,9 +1,22 @@
...
+	app = require('electron').app;
+}
 
+const zulipDir = app.getPath('userData');
+const logDir = `${zulipDir}/Logs/`;
 const initSetUp = () => {
+	// if it is the first time the app is running
+	// create zulip dir in userData folder to
+	// avoid errors
+	if (!fs.existsSync(zulipDir)) {
```
akashnimare: The one thing that worries me here is that the app directory is ideally created by the electron. I hope this doesn't get messed up with that thing 😕

priyank-p: I confirmed that it will add all the files are added afterwards.

akashnimare: @cPhost so we don't need to worry, right?

priyank-p: Yep, to be 100% sure try an packaged app.

akashnimare:
Can you squash the commits here so that I can merge it?

akashnimare:
Merged, thanks @cPhost 🚀

PR #355, opened 2017-12-19
Add reply option to Notifications

priyank-p:
Will work well on all platforms.

To Test it out Toggle DevTools of the active tab and drop this in
```javascript
Notification.onreply = (reply) => {
  console.log('Reply form notification: ', reply);
}

new Notification('Testing Notification', {
  body: 'please reply'
})
```
 Note that you can also use `addEventListener`
```javascript
Notification.addEventListener('reply', (reply) => { 
  console.log('Reply form notification: ', reply);
});
```


priyank-p:
Seems like it does not work on windows. Found lots of Issues on node-notifier but they did not help seems like need to use alternative` electron-windows-notifications` modules.

Native Notifcation do now work either as i tried them.

priyank-p:
Ready for MacOs and Linux, need to figure something else out for windows since there are lots of issues open in Electron, node-notifier for notification not working for windows :/

app/renderer/js/notification.js
```diff
@@ -2,33 +2,139 @@
...
+			const notification = new DefaultNotification(title, opts);
+			notification.addEventListener('click', () => {
+				ipcRenderer.send('focus-app');
+			});
+			return;
+		}
+
+		// focus app incase notification is closed using
+		// close or reply button
+		ipcRenderer.send('focus-app');
```
akashnimare: I don't think we need this here.

package.json
```diff
@@ -53,7 +54,8 @@
         "zip",
         "AppImage"
       ],
-      "maintainer": "Akash Nimare <svnitakash@gmail.com>"
+      "maintainer": "Akash Nimare <svnitakash@gmail.com>",
+      "asarUnpack": "./app/node_modules/node-notifier/vendor/**"
```
akashnimare: Do we really need to unpack this on Linux and Windows?

priyank-p: Yep, The node-notifier says we need to unpack them.

package.json
```diff
@@ -42,7 +42,8 @@
     ],
     "copyright": "©2017 Kandra Labs, Inc.",
     "mac": {
-      "category": "public.app-category.productivity"
+      "category": "public.app-category.productivity",
+      "asarUnpack": "./app/node_modules/node-notifier/vendor/**"
```
akashnimare: @cPhost do we need to unpack this for `node-mac-notifier` also?

priyank-p: No we do need to.

app/renderer/js/notification/index.js
```diff
@@ -4,15 +4,18 @@ const {
   remote: { app }
 } = require('electron');
 
+const { appId } = require('./helpers');
```
priyank-p: Thanks :)

Issue #356, opened 2017-12-20
broken server icon 


akashnimare:
![image](https://user-images.githubusercontent.com/2263909/34221907-9576afcc-e5df-11e7-9270-bced6b9bfb12.png)

This happens when you reload the app and our backend fails to download the org icon from the server. In this type of cases we should set the default icon.

priyank-p:
@zulipbot claim.

PR #357, opened 2017-12-21
This add default icon if the image is not avaliable

priyank-p:
Fixes: https://github.com/zulip/zulip-electron/issues/356

### DIFF
--- a/app/renderer/js/main.js
+++ b/app/renderer/js/main.js
@@ -149,6 +149,13 @@ class ServerManagerView {
 			this.openSettings('General');
 		});
 
+		const $serverImgs = document.querySelectorAll('.server-icons');
+		$serverImgs.forEach($serverImg => {
+			$serverImg.addEventListener('error', () => {
+				$serverImg.src = 'img/icon.png';
+			});
+		});
+
 		this.sidebarHoverEvent(this.$addServerButton, this.$addServerTooltip);
 		this.sidebarHoverEvent(this.$settingsButton, this.$settingsTooltip);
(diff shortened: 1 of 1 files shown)


akashnimare:
I think it would be great if we can handle this in our `domain-util` since all the logic for domain/server/icon handling lies there.

priyank-p:
Yeah, i will do so, i did not know where to put so i put it here.

priyank-p:
@akashnimare rather leave it where it right now because `domain-util.js` solely handles downloading icon and add it to db. So we should leave it where it is now since this is where all the event listeners are handled.

priyank-p:
FYI, the tests are failing because, PR #353 is not merged yet.

akashnimare:
@cPhost can you confirm, if it fixes the issue? I still see the broken image icon.

app/renderer/js/main.js
```diff
@@ -149,6 +149,11 @@ class ServerManagerView {
 			this.openSettings('General');
 		});
 
+		const $serverImg = document.querySelector('.server-icons');
+		$serverImg.addEventListener('error', () => {
+			$serverImg.src = 'img/icon.png';
```
akashnimare: I don't think the path `img/icon.png` will work here. Try -
```
const APP_ICON = path.join(__dirname, '/img/icon.png');
```

akashnimare:
I should mention that the problem occurs sometimes when we reload the app. In the image below you can see the broken server image but if we do insect and check the src the image loads up successfully. Also, the same image is being used in the server section which surprisingly works correctly -

image

akashnimare:
Merged, thanks @cPhost.

PR #360, opened 2017-12-25
Add initial character of Realm Name instead of default icon

priyank-p:


akashnimare:
Merged. Thanks, @priyankp10 🚀

rishig:
A good thought, though we currently use the (default) realm icon consistently everywhere in the app. We can consider whether to change that, but I don't think we want to do something different on desktop than we do on web and mobile.

akashnimare:
> We can consider whether to change that, but I don't think we want to do something different on desktop than we do on web and mobile.

The default icon strategy doesn't affect the web users that much since the browser URL is visible all the time so they actually know in which realm there are in but in the case of the desktop app we need to differentiate. Consider a case - a user has added 5 organizations in the desktop app and those organizations have the same default icon in the left sidebar, in that case, it's difficult for a user to know the current realm (this is also based on few feedbacks I have got). To fix this issue we added the realm name in the tooltip in https://github.com/zulip/zulip-electron/commit/91f3afa8fe840ba629c85f9e627f8e7bc741fdaf and this PR is the follow-up improvements.


image

rishig:
ok, that's reasonable.
I think allowing people to upload custom realm icons would also help.

Issue #369, opened 2017-12-30
Don't allow non-zulip servers to be entered.


priyank-p:
---

- **Operating System**:
  - [X] Windows
  - [X] Linux/Ubutnu
  - [X] macOS

- **Clear steps to reproduce the issue**:
Enter a slack-server

priyank-p:
Found out why, Slack servers send out statusCode `403` which for some reason does not count as error for request module thereof allowed access

priyank-p:
@zulipbot claim.

PR #370, opened 2017-12-30
domain-util: fix checkDomain, so it checks all error codes.

priyank-p:
This fixes a issue where if server send non 404 error code such
as 403 forbidden we marked them as zulip server eventhough they are
not now it checks for 400 error range.

---

**What's this PR do?**
Fixes: #369 

**Screenshots?**
screen shot 2017-12-30 at 1 19 19 pm

**You have tested this PR on:**
  - [X] macOS

PR #375, opened 2018-01-06
Make `checkDomain` fail when a non-Zulip URL is entered.

marco:
Before, any URL without a status code response of 404 and without an
`error` in the `request` callback would fail. However, `error` doesn't
actually populate for error responses; it is populated when there is an
error with the request. Therefore, many websites that responded with
non-404 (but still error) status codes would be accepted. Instead, check
for all error status codes as well as the `request` error.

---

**What's this PR do?**

Properly checks that URLs are valid Zulip servers.

This was done for the *Issue #369 Don't allow non-zulip servers to be entered.* Google Code-in task.

**Screenshots?**

screen shot 2018-01-06 at 10 52 13 am

screen shot 2018-01-06 at 10 51 52 am

screen shot 2018-01-06 at 10 52 05 am

  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [x] macOS

### DIFF
--- a/app/renderer/js/utils/domain-util.js
+++ b/app/renderer/js/utils/domain-util.js
@@ -125,7 +125,20 @@ class DomainUtil {
 				const whitelistDomains = [
 					'zulipdev.org'
 				];
-				if (!error && response.statusCode !== 404) {
+
+				// If error is falsy, convert it to an empty string to prevent an error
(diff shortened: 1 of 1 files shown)


akashnimare:
@skunkmb it has been already fixed in #370. Thanks for working on this anyway.

Issue #378, opened 2018-01-10
Mute notification setting option not working properly


akashnimare:
It looks like app's "Mute all Sounds from Zulip" is not working as expected. The reason behind this could be webview's sound which is taking over the notification's silent parameter.

priyank-p:
Claiming issue to avoid confusion @zulipbot claim.

PR #380, opened 2018-01-10
silent: Reflect changes in webview for silent option.

priyank-p:
This silents the webview incase silent option is toggled, and
by default silent the webview when its create if needed.

Fixes: https://github.com/zulip/zulip-electron/issues/378

---

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [X] macOS

### DIFF
--- a/app/renderer/js/components/webview.js
+++ b/app/renderer/js/components/webview.js
@@ -11,7 +11,7 @@ const { shell, app } = require('electron').remote;
 
 const BaseComponent = require(__dirname + '/../components/base.js');
 
-const silentWebview = ConfigUtil.getConfigItem('silent');
+const shouldSilentWebview = ConfigUtil.getConfigItem('silent');
 class WebView extends BaseComponent {
 	constructor(props) {
 		super();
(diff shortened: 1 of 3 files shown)


app/renderer/js/components/webview.js
```diff
@@ -4,12 +4,14 @@ const path = require('path');
...
 
 const DomainUtil = require(__dirname + '/../utils/domain-util.js');
+const ConfigUtil = require(__dirname + '/../utils/config-util.js');
 const SystemUtil = require(__dirname + '/../utils/system-util.js');
 const LinkUtil = require(__dirname + '/../utils/link-util.js');
 const { shell, app } = require('electron').remote;
 
 const BaseComponent = require(__dirname + '/../components/base.js');
 
+const silentWebview = ConfigUtil.getConfigItem('silent');
```
akashnimare: **isWebviewSilent**.

priyank-p: Actually `shouldSilentWebview` would be better.

akashnimare: Yeah, that would be better :)

akashnimare:
Merged, thanks @cPhost 🚀

PR #390, opened 2018-01-19
Improve dev script 

### LINKED ISSUE #368: Improve gule-dev script
Currently, it does not do much. Few things we should add in this -
* Watch for all the JS/CSS/HTML changes in main and renderer process and then auto-reload app on changes
* Watch for linter also, this could be done using [nodemon](https://github.com/remy/nodemon)

abhigyank:
---

**What's this PR do?**
Fixes #368 
Add script in renderer html to trigger electron.reload in gulp to reload for all the JS/CSS/HTML changes in renderer process
Test for linter also using nodemon in app and renderer on changes in JS/CSS/HTML

**Screenshots?**

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS


akashnimare:
@abhigyank thanks for working on this. Two things -
* You need to add `nodemon` as a dev dependencies in package.json.
* Please follow our commit message guidelines for more details: http://zulip.readthedocs.io/en/latest/contributing/version-control.html

abhigyank:
@akashnimare  Thanks, will keep that in mind from now on. Meanwhile, have pushed a commit adding ```nodemon``` in devDependencies.

priyank-p:
@abhigyank you should also commit the `package-lock.json` so next time when somebody commits they don't have to.

abhigyank:
@cPhost Commited the package-lock.json

abhigyank:
@akashnimare good to merge?

akashnimare:
@abhigyank does this work if you make changes in renderer process? The script doesn't show changes on setting page. For an example if you open setting page and make changes in [this](https://github.com/zulip/zulip-electron/blob/master/app/renderer/js/pages/preference/new-server-form.js) file it won't auto-reload 😕

abhigyank:
@akashnimare  fixed that. Have a look now.

app/renderer/main.html
```diff
@@ -41,4 +41,6 @@
 </body>
 <script src="js/main.js"></script>
 
+<!-- To trigger electron.reload on changes in renderer from gulp -->
+<script>require('electron-connect').client.create()</script>
```
cPhost: As i said in chat. Try this
```javascript
window.addEventListener('load', () => {
  if (isDev) {
    require('electron-connect').client.create();
  }
});
```

priyank-p: As i said in chat. Try this
```javascript
window.addEventListener('load', () => {
  if (isDev) {
    require('electron-connect').client.create();
  }
});
```

akashnimare: @abhigyank we don't need to run this script in production so please use the code which @cPhost has suggested.

akashnimare (changes requested):

abhigyank:
@akashnimare @cPhost moved the script to Dev.

akashnimare (approved):

akashnimare:
Works great. I'll take care of the failing tests. Merged, thanks @abhigyank, @cPhost!

app/renderer/main.html
```diff
@@ -41,4 +41,13 @@
 </body>
 <script src="js/main.js"></script>
 
+<!-- To trigger electron.reload on changes in renderer from gulp -->
+<script>
+  window.addEventListener('load', () => {
+    const isDev = require('electron-is-dev');
+    if (isDev) {
+      require('electron-connect').client.create();
```
akashnimare, line 49: @cPhost, @abhigyank this also called when you run `npm start` which we don't want :/

abhigyank: I am not sure why ```isDev``` is getting set to ```true```, I think it is because ```process.execPath``` contains the ```node_modules``` (the executable that started the Node.js process in dev). 

The one main [line](https://github.com/sindresorhus/electron-is-dev/blob/master/index.js#L5) of the ```electron-is-dev``` module suggests something like this.

priyank-p: That is weird.

abhigyank: I don't think this will be the case in packaged or distributed version. Since, ```npm start``` is basically just ```electron app```, it is possibly treated as development.

akashnimare:
@cPhost, @abhigyank yeah, in the packaged app it won't trigger because of `isDev` check. We need to find a way to disable this when running `npm start`.

Issue #393, opened 2018-01-22
Error on toggling the left sidebar using keyboard shortcut


akashnimare:
The error came from #362. Repro using -
* Delete all servers
* Open add new server page
* CMD+S
* Check console -

```javascript
preference.js:80 Uncaught TypeError: Cannot set property 'checked' of null
    at EventEmitter.ipcRenderer.on (preference.js:80)
    at emitTwo (events.js:106)
    at EventEmitter.emit (events.js:194)
```

akashnimare:
@cPhost quick fix?

priyank-p:
Making a PR in just a bit

PR #394, opened 2018-01-22
preference: Only toggle the state if element is present.

priyank-p:
Fixes: https://github.com/zulip/zulip-electron/issues/393


akashnimare:
Merged, thanks @cPhost 🚀

PR #395, opened 2018-01-25
notifications: Adds bots mention support to reply option.

abhigyank:
Fixes: zulip/zulip-desktop#391 

---

**What's this PR do?**
Adds bots mention support to reply option in notifications.
Make a  ```getJSON``` request to ```/json/users/``` to get list of members and  filter out bots using ```is_bots``` property.

### DIFF
--- a/app/renderer/js/notification/helpers.js
+++ b/app/renderer/js/notification/helpers.js
@@ -3,6 +3,39 @@ const { remote } = require('electron');
 // Do not change this
 const appId = 'org.zulip.zulip-electron';
 
+const botsList = [];
+let botsListLoaded = false;
+
+// this function load list of bots from the server
+// sync=True for a synchronous getJSON request
+// in case botsList isn't already completely loaded when required in parseRely
+function loadBots(sync = false) {
+	const { $ } = window;
+	botsList.length = 0;
+	if (sync) {
+		$.ajaxSetup({async: false});
+	}
+	$.getJSON('/json/users')
+		.done(data => {
+			const members = data.members;
+			members.forEach(membersRow => {
+				if (membersRow.is_bot) {
+					const bot = `@${membersRow.full_name}`;
+					const mention = `@**${bot.replace(/^@/, '')}**`;
+					botsList.push([bot, mention]);
+				}
+			});
+			botsListLoaded = true;
+		})
+		.fail(error => {
+			console.log('Request failed: ', error.responseText);
+			console.log('Request status: ', error.statusText);
+		});
+	if (sync) {
+		$.ajaxSetup({async: true});
+	}
+}
+
 function checkElements(...elements) {
 	let status = true;
 	elements.forEach(element => {
@@ -41,7 +74,6 @@ const webContentsId = webContents.id;
 function focusCurrentServer() {
 	currentWindow.send('focus-webview-with-id', webContentsId);
 }
-
 // this function parses the reply from to notification
 // making it easier to reply from notification eg
 // @username in reply will be converted to @**username**
@@ -83,6 +115,20 @@ function parseReply(reply) {
 		reply = reply.replace(regex, streamMention);
 	});
 
+	// If botsList isn't completely loaded yet, make a synchronous getJSON request for list
+	if (botsListLoaded === false) {
+		loadBots(true);
+	}
+
+	// Iterate for every bot name and replace in reply
+	// @botname with @**botname**
+	botsList.forEach(([bot, mention]) => {
+		if (reply.includes(bot)) {
+			const regex = new RegExp(bot, 'g');
+			reply = reply.replace(regex, mention);
+		}
+	});
+
 	reply = reply.replace(/\\n/, '\n');
 	return reply;
 }
@@ -98,5 +144,6 @@ module.exports = {
 	customReply,
 	parseReply,
 	setupReply,
-	focusCurrentServer
+	focusCurrentServer,
+	loadBots
 };
--- a/app/renderer/js/notification/index.js
+++ b/app/renderer/js/notification/index.js
@@ -5,7 +5,7 @@ const {
 } = require('electron');
 
 const DefaultNotification = require('./default-notification');
-const { appId } = require('./helpers');
+const { appId, loadBots } = require('./helpers');
 
 // From https://github.com/felixrieseberg/electron-windows-notifications#appusermodelid
 // On windows 8 we have to explicitly set the appUserModelId otherwise notification won't work.
@@ -16,3 +16,7 @@ if (process.platform === 'darwin') {
 	const DarwinNotification = require('./darwin-notifications');
 	window.Notification = DarwinNotification;
 }
+
+window.addEventListener('load', () => {
+	loadBots();
+});


app/renderer/js/notification/helpers.js
```diff
@@ -83,6 +83,36 @@ function parseReply(reply) {
 		reply = reply.replace(regex, streamMention);
 	});
 
+	window.$.ajaxSetup({async: false});
```
priyank-p: Right here blocking the thread using synchronous request is not the right approach here

app/renderer/js/notification/helpers.js
```diff
@@ -83,6 +83,36 @@ function parseReply(reply) {
...
+
+	window.$.getJSON('/json/users', data => {
+		const members = data.members;
+		const bots = [];
+		members.forEach(membersRow => {
+			if (membersRow.is_bot) {
+				botsList.push(membersRow);
```
priyank-p: Right here instead of pushing `membersRow` here you should push `memberRow.full_name` and the mention syntax so you do not need to loop through it again.

priyank-p (changes requested):
The feature work correctly here but this needs a bit of code refactoring and the approach or requesting data on each notification is not correct here instead what we want to do is:

Load the bots already outside the parseReply function and asyncronusly.

app/renderer/js/notification/helpers.js
```diff
@@ -83,6 +83,36 @@ function parseReply(reply) {
 		reply = reply.replace(regex, streamMention);
 	});
 
+	window.$.ajaxSetup({async: false});
+
+	window.$.getJSON('/json/users', data => {
```
priyank-p: One more minor comment you should deconstruct the `$` variable up top.
`const { $ } = window;`

abhigyank:
@cPhost The problem with loading the bots already outside the ```parseReply``` function would be that the list would be static. After the list is generated, if someone creates a new bot, then that bot can't be mentioned in the reply since that list wouldn't have that bot. Hence, despite knowing that blocking the thread using synchronous request is not the a good way, I have done that. I feel we need to get the list from the server every time ```parseReply``` is called to get new bots if there are any.

priyank-p:
Do get your point but we don't care if new bot was added or not this is like a extra convince feature we provide people are not likely to be mentioning bots specifically new bots + in a mention so we shouldn't worry about it too much.

abhigyank:
Oh, in that case it won't be a issue. I'll make the changes and push it.

app/renderer/js/notification/helpers.js
```diff
@@ -83,6 +101,30 @@ function parseReply(reply) {
...
 
+	if (botsListLoaded) {
+		botsListLoaded.forEach(([bot, mention]) => {
+			const regex = new RegExp(bot, 'g');
+			reply = reply.replace(regex, mention);
+		});
+	} else {
+		$.ajaxSetup({async: false});
+
+		$.getJSON('/json/users', data => {
```
priyank-p: It would be better call `loadBots()` here instead here.

app/renderer/js/notification/helpers.js
```diff
@@ -92,6 +134,10 @@ function setupReply(id) {
 	narrow.by_subject(id, { trigger: 'notification' });
 }
 
+window.addEventListener('load', () => {
```
priyank-p: You should't put this under `load` because the while purpose of the function is to load bots before hand not after load event where notification could be created.

priyank-p (commented):
It would be better to call `loadsBots` instead of repeating the code and in `loadBots` reintialze the botList array like `botsList = []` in callback so no duplicates are being added.

And remove the function out of `load` event.

app/renderer/js/notification/helpers.js
abhigyank: The problem here is that, before ```load```, ```$``` is not defined.

priyank-p: Oh, yeah i did not think about it you can ignore this then.

priyank-p (approved):
@abhigyank this LGTM. requesting final review @akashnimare

app/renderer/js/notification/helpers.js
akashnimare: @abhigyank, @cPhost  works great. I think the `load` listener should be added in `notification/index.js`.  Can you do that?

app/renderer/js/notification/helpers.js
```diff
@@ -3,6 +3,31 @@ const { remote } = require('electron');
 // Do not change this
 const appId = 'org.zulip.zulip-electron';
 
+const botsList = [];
+let botsListLoaded = false;
+
+function loadBots(sync = false) {
```
akashnimare: @abhigyank would be great if you can add few comments here about the function?  In general, it's better to document your code so that other devs can see  what's going on :)

app/renderer/js/notification/helpers.js
```diff
@@ -83,6 +107,15 @@ function parseReply(reply) {
 		reply = reply.replace(regex, streamMention);
 	});
 
+	if (botsListLoaded === false) {
+		loadBots(true);
+	}
+
+	botsList.forEach(([bot, mention]) => {
```
akashnimare: here as well.

app/renderer/js/notification/helpers.js
```diff
@@ -92,6 +125,10 @@ function setupReply(id) {
 	narrow.by_subject(id, { trigger: 'notification' });
 }
 
+window.addEventListener('load', () => {
```
akashnimare: Put this in `/notification/index.js`.

akashnimare (changes requested):

abhigyank:
@akashnimare Amended commit with comments at specified places and moved ```load``` to ```index.js``` .  Request review.

akashnimare (approved):

akashnimare:
Before merging, I would want you to handle following cases-

* if the user is not logged in -
```sh
jquery.js?61dd:9566 GET https://chat.zulip.org/json/users 401 ()
```
* In case of muliple servers -
```sh
jquery.js?61dd:9566 GET https://playground.zulipdev.org/json/users 401 (Unauthorized)
﻿
```

abhigyank:
@akashnimare The 401() error would be thrown on the console by jQuery. How do you wish to handle these cases?

akashnimare:
For now, we can just log them. Maybe use something like -
```javascript
function loadBots(sync = false) {
	const { $ } = window;
	botsList.length = 0;
	if (sync) {
		$.ajaxSetup({ async: false });
	}
	$.getJSON('/json/users')
		.done((data) => {
			const members = data.members;
			members.forEach(membersRow => {
				if (membersRow.is_bot) {
					const bot = `@${membersRow.full_name}`;
					const mention = `@**${bot.replace(/^@/, '')}**`;
					botsList.push([bot, mention]);
				}
			});
			botsListLoaded = true;
			if (sync) {
				$.ajaxSetup({ async: true });
			}
		})
		.fail((error) => {
			console.log("Request Failed: ", error.responseText);
			console.log(`Request status: ${error.status + error.statusText}`);
		});
}
```
(dummy code  feel free  to refactor)

akashnimare:
@abhigyank merged, thanks 🚀

PR #400, opened 2018-01-29
proxy: Change proxy input placeholder

abhigyank:
This follows from the discussion [here](https://chat.zulip.org/#narrow/stream/electron/topic/Proxy).

**What's this PR do?**
Change proxy input placeholder to http_proxy:80;ftp_proxy:2121 to better suggest usage of multiple proxies

### DIFF
--- a/app/renderer/js/pages/preference/network-section.js
+++ b/app/renderer/js/pages/preference/network-section.js
@@ -27,7 +27,7 @@ class NetworkSection extends BaseSection {
 						</div>
 						<div class="setting-row" id="proxy-rules-option">
 							<span class="setting-input-key">Proxy rules</span>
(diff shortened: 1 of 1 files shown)


akashnimare:
Looks good. Can you update the docs in `help.md`? Also, can you paste the steps here which followed to cross-test this?

abhigyank:
I read the ````help.md``` guide to proxy, and understood what actually the proxy field expects. I tested again in the previous format properly, and it worked. So it basically works in both ways. The original placeholder way as well as the one I have pushed. 
 
So I don't think is PR is necessary. Hence, closing.

akashnimare:
@abhigyank no worries, thanks for testing that out 😄

Issue #396, opened 2018-01-29
Add a loading indicator when user clicks on new server button


akashnimare:
We need to show a loading indicator or change the text of "Add" button to "Adding..." when user clicks on  **Add** button.

![image](https://user-images.githubusercontent.com/2263909/35507450-d3db9442-0512-11e8-8eca-934d8a1fa461.png)

abhigyank:
I am on it.

PR #401, opened 2018-01-29
preference page: Loading indication for new server button.

abhigyank:
**What's this PR do?**
Changes the text of "Add" button to "Adding..." when user clicks on Add button for adding new server by chanding the ```innerHTML``` of the ```span``` dom element responsible for the text "Add". 
Fixes: #396 

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/pages/preference/new-server-form.js
+++ b/app/renderer/js/pages/preference/new-server-form.js
@@ -43,11 +43,13 @@ class NewServerForm extends BaseComponent {
 	}
 
 	submitFormHandler() {
+		this.$saveServerButton.children[1].innerHTML = 'Adding...';
 		DomainUtil.checkDomain(this.$newServerUrl.value).then(serverConf => {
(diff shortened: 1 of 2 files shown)


akashnimare:
@abhigyank works well in all the cases except when you try to add a server which is already added.

abhigyank:
@akashnimare  In that case, this [line](https://github.com/zulip/zulip-electron/blob/master/app/renderer/js/pages/preference/new-server-form.js#L46) throws up an error, because of the ```return``` [here](https://github.com/zulip/zulip-electron/blob/master/app/renderer/js/utils/domain-util.js#L103). Check the developer console for the error.

abhigyank:
@akashnimare  I have amended the commit, changing the ```return``` statement to return a ```Promise.reject()``` object. Works correctly now.

akashnimare:
@abhigyank merged. Thanks 🚀

Issue #398, opened 2018-01-29
Send a notification when a new update available [Debian]


akashnimare:
This could be done easily using the `update-available` event of [autoupdater](https://www.electron.build/auto-update) but we need to first check whether it's supported by electron-updater or not.

akashnimare:
I had a chat with the developer of electron-builder and he said there is no plan to add this feature in near future so we gotta add this on our own. 

Potential solution - Add a script to get the latest release from GitHub and then compare it to the installed version.

abhigyank:
I'd like to work on this. Where should the script be added?
I am thinking of working on the solution you suggest. With github API we can get the latest release version and compare it with `app.getVersion()` .

akashnimare:
I think it should be added somewhere in `app/main/`.

abhigyank:
Isn't auto updates already being done [here](https://github.com/zulip/zulip-electron/blob/master/app/main/autoupdater.js) using `electron-updater`?

priyank-p:
@abhigyank yep its done there but the real issue is that on debian electron-updater does not work and that way we need you do it use a github api. Does that make sense?

Generally if have other question post them at [#electron](https://chat.zulip.org/#narrow/stream/electron) it work better there, since you can more in detail answer for what you looking for :)

PR #408, opened 2018-02-08
updates: Pop a notification when a new update available.

abhigyank:
**What's this PR do?**
Send a notification when a new update available [Debian]

Fixes: #398.

### DIFF
--- a/app/main/autoupdater.js
+++ b/app/main/autoupdater.js
@@ -11,6 +11,12 @@ function appUpdater() {
 		return;
 	}
 
+	if (process.platform === 'linux' && !process.env.APPIMAGE) {
+		const { debianUpdateNotification } = require('./debianupdater');
+		debianUpdateNotification();
+		return;
+	}
+
 	// Create Logs directory
 	const LogsDir = `${app.getPath('userData')}/Logs`;
 
--- a/app/main/autoupdater.js
+++ b/app/main/autoupdater.js
@@ -12,8 +12,8 @@ function appUpdater() {
 	}
 
 	if (process.platform === 'linux' && !process.env.APPIMAGE) {
-		const { debianUpdateNotification } = require('./debianupdater');
-		debianUpdateNotification();
+		const { linuxUpdateNotification } = require('./linuxupdater');
+		linuxUpdateNotification();
 		return;
 	}
 
--- a/app/main/debianupdater.js
+++ b/app/main/debianupdater.js
@@ -1,0 +1,45 @@
+const { app } = require('electron');
+const { Notification } = require('electron');
+
+const request = require('request');
+const semver = require('semver');
+const ConfigUtil = require('../renderer/js/utils/config-util');
+
+function debianUpdateNotification() {
+	let url = '';
+
+	if (ConfigUtil.getConfigItem('betaUpdate')) {
+		url = 'https://api.github.com/repos/zulip/zulip-electron/releases';
+	} else {
+		url = 'https://api.github.com/repos/zulip/zulip-electron/releases/latest';
+	}
+	const options = {
+		url,
+		headers: {'User-Agent': 'request'}
+	};
+
+	request(options, (error, response, body) => {
+		if (error) {
+			console.log('Error:', error);
+			return;
+		}
+		if (response.statusCode < 400) {
+			const data = JSON.parse(body);
+			let latestVersion = '';
+			if (ConfigUtil.getConfigItem('betaUpdate')) {
+				latestVersion = data[0].tag_name;
+			} else {
+				latestVersion = data.tag_name;
+			}
+			if (semver.gt(latestVersion, app.getVersion())) {
+				new Notification({title: 'Zulip Update', body: 'A new version ' + latestVersion + ' is available. Please update using your package manager.'}).show();
+			}
+		} else {
+			console.log('Status:', response.statusCode);
+		}
+	});
+}
+
+module.exports = {
+	debianUpdateNotification
+};
--- a/app/main/debianupdater.js
+++ b/app/main/debianupdater.js
@@ -32,7 +32,11 @@ function debianUpdateNotification() {
 				latestVersion = data.tag_name;
 			}
 			if (semver.gt(latestVersion, app.getVersion())) {
-				new Notification({title: 'Zulip Update', body: 'A new version ' + latestVersion + ' is available. Please update using your package manager.'}).show();
+				const notified = ConfigUtil.getConfigItem(latestVersion);
+				if (notified === null || notified === false) {
+					new Notification({title: 'Zulip Update', body: 'A new version ' + latestVersion + ' is available. Please update using your package manager.'}).show();
+					ConfigUtil.setConfigItem(latestVersion, true);
+				}
 			}
 		} else {
 			console.log('Status:', response.statusCode);
--- a/app/main/linuxupdater.js
+++ b/app/main/linuxupdater.js
@@ -4,6 +4,7 @@ const { Notification } = require('electron');
 const request = require('request');
 const semver = require('semver');
 const ConfigUtil = require('../renderer/js/utils/config-util');
+const LinuxUpdateUtil = require('../renderer/js/utils/linux-update-util');
 
 function linuxUpdateNotification() {
 	let	url = 'https://api.github.com/repos/zulip/zulip-electron/releases';
@@ -24,10 +25,10 @@ function linuxUpdateNotification() {
 			const latestVersion = ConfigUtil.getConfigItem('betaUpdate') ? data[0].tag_name : data.tag_name;
 
 			if (semver.gt(latestVersion, app.getVersion())) {
-				const notified = ConfigUtil.getConfigItem(latestVersion);
+				const notified = LinuxUpdateUtil.getUpdateItem(latestVersion);
 				if (notified === null) {
 					new Notification({title: 'Zulip Update', body: 'A new version ' + latestVersion + ' is available. Please update using your package manager.'}).show();
-					ConfigUtil.setConfigItem(latestVersion, true);
+					LinuxUpdateUtil.setUpdateItem(latestVersion, true);
 				}
 			}
 		} else {
--- a/app/main/linuxupdater.js
+++ b/app/main/linuxupdater.js
@@ -5,14 +5,10 @@ const request = require('request');
 const semver = require('semver');
 const ConfigUtil = require('../renderer/js/utils/config-util');
 
-function debianUpdateNotification() {
-	let url = '';
+function linuxUpdateNotification() {
+	let	url = 'https://api.github.com/repos/zulip/zulip-electron/releases';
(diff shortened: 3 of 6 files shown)


app/main/autoupdater.js
```diff
@@ -48,6 +50,41 @@ function appUpdater() {
 	autoUpdater.checkForUpdates();
 }
 
+function updateNotification() {
+	const options = {
```
cPhost: You should use the `request` module here. This will make it more readable here.

app/main/autoupdater.js
```diff
@@ -48,6 +50,41 @@ function appUpdater() {
...
+		host: 'api.github.com',
+		path: '/repos/zulip/zulip-electron/releases/latest',
+		headers: {'User-Agent': 'request'}
+	};
+
+	https.get(options, res => {
```
priyank-p: You should use the `request` module here. This will make it more readable here and code shorter.

priyank-p (commented):
@abhigyank can you use request module instead of https here.

app/main/autoupdater.js
```diff
@@ -52,35 +52,27 @@ function appUpdater() {
...
-				}
-			} else {
-				console.log('Status:', res.statusCode);
+	request(options, (error, response, body) => {
+		if (error) {
+			console.log('Error:', error);
+			return;
+		}
+
+		if (response.statusCode === 200) {
```
priyank-p: We should use `response.statusCode < 400` here to avoid some problem in future. Sorry i missed this before.

akashnimare:
We only need to trigger the notification if:
   - Platform is Debian and the installer is not an AppImage (since the auto-updates works in this format).

app/main/autoupdater.js
```diff
@@ -48,6 +50,33 @@ function appUpdater() {
 	autoUpdater.checkForUpdates();
 }
 
+function updateNotification() {
```
akashnimare: Can you rename it to something like - `updateDebianNotification` and put it in a different file.

app/main/index.js
```diff
@@ -162,6 +162,7 @@ app.on('ready', () => {
 
 	page.once('did-frame-finish-load', () => {
 		// Initate auto-updates on MacOS and Windows
+		updateNotification();
```
akashnimare: We don't need to trigger this on Windows and macOS.

package.json
```diff
@@ -161,5 +161,8 @@
       "browser",
       "mocha"
     ]
+  },
+  "dependencies": {
```
akashnimare: All the app dependencies should go in `/app/package.json`. This is only for dev dependencies. Also, we use exact working version instead of `^`.

akashnimare (changes requested):

app/main/autoupdater.js
abhigyank: What should be the file name?

abhigyank:
@akashnimare redid the commit adding a new file. Review.

app/main/debianupdater.js
```diff
@@ -0,0 +1,40 @@
+const { app } = require('electron');
+const { Notification } = require('electron');
+
+const request = require('request');
+const semver = require('semver');
+const ConfigUtil = require('../renderer/js/utils/config-util');
+
+function updateDebianNotification() {
```
priyank-p: Can we use `debianUpdateNotification`?

priyank-p (approved):

app/main/debianupdater.js
abhigyank: Is that necessary? Kind of the same, I feel.

priyank-p: Totally, depends on how you guys fell about it :)

abhigyank: ping @akashnimare .

akashnimare: Yeah, `debianUpdateNotification` sounds good :)

akashnimare:
@abhigyank we don't need to trigger this function when the app is running from the AppImage format. You can use `process.env.APPIMAGE` to check that app running using AppImage or not.

app/main/debianupdater.js
```diff
@@ -0,0 +1,40 @@
...
+	};
+
+	request(options, (error, response, body) => {
+		if (error) {
+			console.log('Error:', error);
+			return;
+		}
+		if (response.statusCode < 400) {
+			const data = JSON.parse(body);
+			if (semver.gt(data.tag_name, app.getVersion())) {
```
akashnimare: In case of pre-release i.e. `https://api.github.com/repos/zulip/zulip-electron/releases`, it will be `data[1].tag_name`.

app/main/index.js
```diff
@@ -163,6 +164,9 @@ app.on('ready', () => {
 	page.once('did-frame-finish-load', () => {
 		// Initate auto-updates on MacOS and Windows
 		appUpdater();
+		if (process.platform === 'linux') {
+			updateDebianNotification();
```
akashnimare: I think this should go in `autoupdater.js`.

app/package.json
```diff
@@ -25,6 +25,14 @@
     "Desktop app",
     "InstantMessaging"
   ],
+  "build": {
```
akashnimare: This should go into the development package.json i.e. - `zulip-electron/package.json`. Read more about the structure [here](https://www.electron.build/tutorials/two-package-structure).

akashnimare (changes requested):

akashnimare:
@abhigyank it would be great if you cross-check your solution before pushing here.

app/main/index.js
```diff
@@ -5,6 +5,7 @@ const windowStateKeeper = require('electron-window-state');
 const isDev = require('electron-is-dev');
 const appMenu = require('./menu');
 const { appUpdater } = require('./autoupdater');
+const { updateDebianNotification } = require('./debianupdater');
```
priyank-p: You need to move this inside the `if` statement in line 167, otherwise the process will crash since this is a main process.

akashnimare: @cPhost I don't think that would be enough.

abhigyank: It is working fine now though without crashing. Why would it crash?

priyank-p: @akashnimare i think that enough since semver is used in `debianupdater.js` and checking is done in this file. Does that make sense?

@abhigyank we remove the semver package in platform is not linux since you require a modules that not present it will throw an error and therefore crash the app.

(This is getting messy here, I'd rather prefer discussing stuff on chat.zulip.org)

app/main/debianupdater.js
abhigyank: Works suitably on `data[0].tag_name` . Pushing that.

app/main/debianupdater.js
```diff
@@ -0,0 +1,45 @@
...
+		if (response.statusCode < 400) {
+			const data = JSON.parse(body);
+			let latestVersion = '';
+			if (ConfigUtil.getConfigItem('betaUpdate')) {
+				latestVersion = data[0].tag_name;
+			} else {
+				latestVersion = data.tag_name;
+			}
+			if (semver.gt(latestVersion, app.getVersion())) {
+				new Notification({title: 'Zulip Update', body: 'A new version ' + data.tag_name + ' is available. Please update using your package manager.'}).show();
```
priyank-p: `latestVersion` instead of `data.tag_name` here

abhigyank: Sorry! Thanks. Amended the commit. :)

akashnimare:
@abhigyank can you resolve the merge conflicts?

akashnimare (approved):

abhigyank:
@akashnimare rebased and squashed.

akashnimare:
We should create `debianupdater.json` in `app.getPath('userData')` directory as all the app data, config files lies there.

priyank-p:
@akashnimare Yep, and the json structure need some more work too.

akashnimare:
Is this ready to review?

priyank-p:
@akashnimare yep it is.

app/main/debianupdater.js
```diff
@@ -0,0 +1,49 @@
...
+			const data = JSON.parse(body);
+			let latestVersion = '';
+			if (ConfigUtil.getConfigItem('betaUpdate')) {
+				latestVersion = data[0].tag_name;
+			} else {
+				latestVersion = data.tag_name;
+			}
+			if (semver.gt(latestVersion, app.getVersion())) {
+				const notified = ConfigUtil.getConfigItem(latestVersion);
+				if (notified === null || notified === false) {
```
priyank-p: Minor nit: remove checking it for false, it would never be false, since we don't set it to be false.

app/main/debianupdater.js
```diff
@@ -0,0 +1,49 @@
...
+const ConfigUtil = require('../renderer/js/utils/config-util');
+
+function debianUpdateNotification() {
+	let url = '';
+
+	if (ConfigUtil.getConfigItem('betaUpdate')) {
+		url = 'https://api.github.com/repos/zulip/zulip-electron/releases';
+	} else {
+		url = 'https://api.github.com/repos/zulip/zulip-electron/releases/latest';
+	}
```
priyank-p: This whole if/else stuff could be simplified:
```javascript
let url = 'https://api.github.com/repos/zulip/zulip-electron/releases';
url = ConfigUtil.getConfigItem('betaUpdate') ? url + '/latest' : url
```

app/main/debianupdater.js
```diff
@@ -0,0 +1,49 @@
...
+			return;
+		}
+		if (response.statusCode < 400) {
+			const data = JSON.parse(body);
+			let latestVersion = '';
+			if (ConfigUtil.getConfigItem('betaUpdate')) {
+				latestVersion = data[0].tag_name;
+			} else {
+				latestVersion = data.tag_name;
+			}
```
priyank-p: ditto

priyank-p (commented):
Let some comment that could make code more simple/clear to follow

akashnimare:
@abhigyank can you make these changes? Also, I think we should rename `debian` to `linux` as it's for Linux in general.
> debianupdater.js = linuxupdater.js
> debianUpdateNotification = linuxUpdateNotification

priyank-p:
@abhigyank thanks this now lgtm.

app/main/linuxupdater.js
```diff
@@ -26,14 +22,11 @@ function debianUpdateNotification() {
 		if (response.statusCode < 400) {
 			const data = JSON.parse(body);
 			let latestVersion = '';
```
akashnimare: I think you can remove this and re-write this as - 
```javascript
const latestVersion = ConfigUtil.getConfigItem('betaUpdate') ? data[0].tag_name : data.tag_name;
```

akashnimare:
This looks good though we need to refactor `linux-util.js` and `config-util.js` as it's almost the same code. Merging it for now. Thanks, @abhigyank and @cPhost 🚀

PR #411, opened 2018-02-12
electron-connect: Disable electron-connect when it's ran from npm start.

priyank-p:


akashnimare:
@cPhost merged, thanks!

Issue #208, opened 2017-07-12
Add back button


gnprice:
In a browser, I find hitting alt-left (the keyboard shortcut for the back button) super handy and do it all the time.

We should have a back button (and maybe a forward button? but not sure where to put it), and also the hotkeys alt-left/right to go back and forward. Or Cmd-left/right, or whatever the appropriate thing is per platform.

timabbott:
Probably worth separating this issues.  `alt+left` and `alt+right` we should be able to dispatch quickly without making any decisions, but some design work needs to be done to figure out where to put the back/forward buttons.

akashnimare:
> In a browser, I find hitting alt-left (the keyboard shortcut for the back button) super handy and do it all the time.

@gnprice added `Alt+Left` and `Alt+Right`. 

macOS -

![image](https://user-images.githubusercontent.com/2263909/28189163-77c2d354-6842-11e7-940a-e72c51edecd3.png)

Ubuntu - 

![image](https://user-images.githubusercontent.com/2263909/28189454-9aae9000-6843-11e7-80c8-7bb479feb485.png)

timabbott:
Nice!  We may want to close this and open a new, somewhat lower priority issue for where to put a more visible back button (or are we still waiting on Windows?).

gnprice:
@akashnimare Sweet! Thanks for the fast action :)

On macOS, will Cmd-left and Cmd-right work, as well as Cmd-[ and Cmd-]? I don't currently have a Mac to try this on, but I used them for years, and I could not have told you that Cmd-[ / Cmd-] were browser shortcuts for back/forward; the shortcuts I used were I'm pretty sure Cmd-left and Cmd-right.

akashnimare:
Yeah, normally browsers have both the shortcuts available `Cmd-[/Cmd-]` and `Cmd-Left/Cmd-right`. I have changed it to `Cmd-Left/Cmd-Right` since it's the expected shortcuts. @timabbott on windows it works fine.

![shortcut](https://user-images.githubusercontent.com/2263909/28241148-78dfb926-69ac-11e7-96ee-c3555958c1ac.PNG)

timabbott:
Great.  If we can make both versions on Mac work to match browser behavior, that'd be nice, but probably not super important.

akashnimare:
@timabbott done.

abhigyank:
I think the back button can go on the sidebar, above the icons.
[...]

PR #413, opened 2018-02-13
design: Add back button.

abhigyank:
Fixes: #208

Discussions related to design is [here](https://chat.zulip.org/#narrow/stream/design/topic/Use.20of.20back.20button). 

The back button moves back the active tab's webview. 

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/menu.js
+++ b/app/main/menu.js
@@ -368,7 +368,7 @@ class AppMenu {
 
 	static sendAction(action, ...params) {
 		const win = BrowserWindow.getAllWindows()[0];
-
+	
 		if (process.platform === 'darwin') {
 			win.restore();
 		}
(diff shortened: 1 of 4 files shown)


akashnimare:
Merged, thanks @abhigyank!

app/main/menu.js
```diff
@@ -368,7 +368,7 @@ class AppMenu {
 
 	static sendAction(action, ...params) {
 		const win = BrowserWindow.getAllWindows()[0];
-
+
```
priyank-p, line 371: @abhigyank any reason to add this, linter fails locally?

abhigyank: That was mistake. Sorry!
But it didn't fail lint test while commiting though.

priyank-p: That's weird, no problem anymore since got the CI working again so it fails on linting error which it should have done.

Issue #312, opened 2017-10-06
Zulip client (most of the time) won't reconnect after waking from hibernation


zingmars:
Hello, 
As title says, after waking up the computer from hibernation the client won't reconnect to the server automatically - it gives the ["Zulip can't connect"](https://i.imgur.com/H82EoHn.png) screen when you open (or 'Focus') the client. Clicking the "Try now" button will of course work, but you might not think to check it after booting up and might miss a private message or two.

-
Zulip 1.4.0
Electron 1.6.11
win32 x64 10.0.15063

timabbott:
@zingmars how long did you wait before clicking the button?  The code we have will automatically try to reconnect  if you're off the Internet, with exponential backoff, so I wonder whether what happened here is that you first went off the Internet, then you hibernated, and so the exponential backoff had already gotten to its peak of "once a minute".

It's possible the issue could be something totally different, though.

zingmars:
I think about 30 minutes? I'm not sure, I never actually timed it.

Is there a way to debug this? I.e. to see how long until it tries to reconnect again? rocket.chat just had a giant message that overlaid the window that said "Reconnecting in ... seconds".

timabbott:
OK, that definitely is outside the space of plausible with the theories I had.

We probably should open an issue over in zulip/zulip for doing that sort of countdown; it'd be easy enough to do.  But since there's a cap on the timeout, what's happening here is something else.  

If you see this again and want to help debug, ping me on chat.zulip.org -- there's some stuff we can do on the JS console to see where the failure is.

zingmars:
Alright, will do.

zingmars:
Per request - screenshot of the developer's console when the app is in the stuck state - 
![image](https://user-images.githubusercontent.com/917126/31348384-2884379e-ad28-11e7-978a-42ef8ee32fdd.png)

timabbott:
@akashnimare the reason I was confused here is that we have two "Try now" buttons for reconnecting; one for when you're already connected to the server in that session (totally within the webapp code), and this one, totally within the desktop app's code.  It's not obvious to me from the screenshot or directly referenced code why retries aren't working, but hopefully you can figure it out.

As a sidenote, if we can figure out how to make electron include timestamps in its console logging, that would probably be a really good idea.  We could think about whether `static/js/blueslip.js` in Zulip is helpful.

priyank-p:
We can now log stuff with timestamps here are simple instruction to do so. 
```javascript
[...]

PR #415, opened 2018-02-14
reconnect: Check wheather internet is working before reloading.

priyank-p:
When online event is triggered check weather the internet is atually working or not.
Commonly on windows it turns out that internet takes couple of seconds to boot up after
connecting to internet or in some cases this might be they have to sign in to internet service
portal in order to access internet.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -32,6 +32,7 @@
     "electron-spellchecker": "1.1.2",
     "electron-updater": "2.18.2",
     "electron-window-state": "4.1.1",
+    "is-online": "7.0.0",
     "node-json-db": "0.7.3",
     "request": "2.81.0",
     "wurl": "2.5.0"
--- a/app/renderer/js/main.js
+++ b/app/renderer/js/main.js
@@ -11,6 +11,7 @@ const WebView = require(__dirname + '/js/components/webview.js');
 const ServerTab = require(__dirname + '/js/components/server-tab.js');
 const FunctionalTab = require(__dirname + '/js/components/functional-tab.js');
 const ConfigUtil = require(__dirname + '/js/utils/config-util.js');
+const ReconnectUtil = require(__dirname + '/js/utils/reconnect-util.js');
 
 class ServerManagerView {
 	constructor() {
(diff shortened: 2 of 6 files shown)


app/package.json
```diff
@@ -32,6 +32,7 @@
     "electron-spellchecker": "1.1.2",
     "electron-updater": "2.18.2",
     "electron-window-state": "4.1.1",
+    "is-online": "^7.0.0",
```
akashnimare: Should bt the exact working version here.

priyank-p: Done

app/renderer/js/main.js
```diff
@@ -488,9 +489,14 @@ class ServerManagerView {
 
 window.onload = () => {
 	const serverManagerView = new ServerManagerView();
+	const reconnectUtil = new ReconnectUtil(serverManagerView);
 	serverManagerView.init();
 	window.addEventListener('online', () => {
-		serverManagerView.reloadView();
+		reconnectUtil.pollInternetAndReload();
```
akashnimare, line 495: Is there a need for adding a console log, just to make the debugging better (for issues like https://github.com/zulip/zulip-electron/issues/312) -
`console.log('You're back online.')`
`console.log('No internet connection, you are offline.')`

priyank-p: Ah yeah, so its easier to figure out.

app/renderer/js/utils/reconnect-util.js
```diff
@@ -0,0 +1,52 @@
...
+	_checkAndReload() {
+		return new Promise(resolve => {
+			if (!this.alreadyReloaded) { // eslint-disable-line no-negated-condition
+				isOnline()
+				.then(online => {
+					if (online) {
+						if (!this.alreadyReloaded) {
+							this.serverManagerView.reloadView();
+						}
+						console.log("You're back online.");
```
akashnimare: `console.log('You're back online.');`

akashnimare: @cPhost this.

priyank-p: Ah. i forgot to `git add` this :)

priyank-p: I though if you had an `'` inside a string that needs to escaped we can use double quotes.

Issue #298, opened 2017-09-15
Add desktop notifications indicator in left sidebar


grahamegrieve:
We need the ability to easily suppress notifications from all chats when presenting - and you don't want to do that on a stream-by-stream.  And there needs to be a visual indicator (e.g. on the icon) so you get reminded to unsuppress when finished

timabbott:
There's actually a feature recently merged in the desktop app that does exactly this, in the desktop app settings page.  It might have been added since the last release; you can try running from Git to verify...

grahamegrieve:
is the desktop app = the electron app?

lmckenzi:
Easy = a button or a simple right-click option.  It shouldn't be necessary to go into settings and a specific menu area.  It needs to be super-easy so you can quickly do it when things start popping up when sharing your screan with others.

timabbott:
I very much agree with the sentiment.

@lmckenzi do you have a suggestion for where you'd like such a button to be?  We'd certainly be happy to make this more visible.  We can probably use a standard icon for this.

akashnimare:
@timabbott this is what I was talking about :)
@grahamegrieve you can use the `Show Desktop Notifications` option for this (set it off).

![image](https://user-images.githubusercontent.com/2263909/30496868-621e598e-9a6e-11e7-8c16-2775894f642e.png)

grahamegrieve:
I have the latest client (1.4.0) and I don't have that setting. I think Lloyd would like it to be a single keystroke, or a setting from the app icon.

lmckenzi:
If you can afford the real-estate, a small icon top-left right under the menu (and above the icons for the servers you're connected to) that you can just toggle would be ideal - where the icon signifies whether notifications are muted or not.  That gives you both a quick obvious place to click and a visual indicator of your current state.

akashnimare:
@grahamegrieve got it. This feature is in the latest master and has not pushed to production yet. You'll get it in the next release (1.5.0).

akashnimare:
@lmckenzi another option is to add it in the menu item. Something like this - 

![image](https://user-images.githubusercontent.com/2263909/30497673-53eba1d4-9a71-11e7-91cf-d5b1db461f10.png)

timabbott:
We should probably spell that "Show Desktop Notifications", but yeah, that's not a bad idea.

lmckenzi:
[...]

PR #417, opened 2018-02-15
feature: Add DND button in left sidebar.

jainaman224:
Fixes: #298 

**What's this PR do?**
This adds the do not disturb button to the left sidebar.

**Screenshots?**
ezgif com-video-to-gif 2

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/menu.js
+++ b/app/main/menu.js
@@ -7,6 +7,7 @@ const { app, shell, BrowserWindow, Menu, dialog } = require('electron');
 const fs = require('fs-extra');
 
 const ConfigUtil = require(__dirname + '/../renderer/js/utils/config-util.js');
+const DNDUtil = require(__dirname + '/../renderer/js/utils/dnd-util.js');
 
 const appName = app.getName();
 
@@ -218,6 +219,13 @@ class AppMenu {
 			}, {
 				type: 'separator'
 			}, {
+				label: 'Toggle Do Not Disturb',
+				accelerator: 'Command+Shift+M',
+				click() {
+					const dndUtil = DNDUtil.toggle();
+					AppMenu.sendAction('toggle-dnd', dndUtil.dnd, dndUtil.newSettings);
+				}
+			}, {
 				label: 'Reset App Settings',
 				accelerator: 'Command+Shift+D',
 				click() {
@@ -319,6 +327,13 @@ class AppMenu {
 				}
 			}, {
 				type: 'separator'
+			}, {
+				label: 'Toggle Do Not Disturb',
+				accelerator: 'Ctrl+Shift+M',
+				click() {
+					const dndUtil = DNDUtil.toggle();
+					AppMenu.sendAction('toggle-dnd', dndUtil.dnd, dndUtil.newSettings);
+				}
 			}, {
 				label: 'Reset App Settings',
 				accelerator: 'Ctrl+Shift+D',
--- a/app/renderer/js/main.js
+++ b/app/renderer/js/main.js
@@ -11,6 +11,7 @@ const WebView = require(__dirname + '/js/components/webview.js');
 const ServerTab = require(__dirname + '/js/components/server-tab.js');
 const FunctionalTab = require(__dirname + '/js/components/functional-tab.js');
 const ConfigUtil = require(__dirname + '/js/utils/config-util.js');
+const DNDUtil = require(__dirname + '/js/utils/dnd-util.js');
 const ReconnectUtil = require(__dirname + '/js/utils/reconnect-util.js');
 
 class ServerManagerView {
@@ -23,12 +24,14 @@ class ServerManagerView {
 		this.$settingsButton = $actionsContainer.querySelector('#settings-action');
 		this.$webviewsContainer = document.getElementById('webviews-container');
 		this.$backButton = $actionsContainer.querySelector('#back-action');
+		this.$dndButton = $actionsContainer.querySelector('#dnd-action');
 
 		this.$addServerTooltip = document.getElementById('add-server-tooltip');
 		this.$reloadTooltip = $actionsContainer.querySelector('#reload-tooltip');
 		this.$settingsTooltip = $actionsContainer.querySelector('#setting-tooltip');
 		this.$serverIconTooltip = document.getElementsByClassName('server-tooltip');
 		this.$backTooltip = $actionsContainer.querySelector('#back-tooltip');
+		this.$dndTooltip = $actionsContainer.querySelector('#dnd-tooltip');
 
 		this.$sidebar = document.getElementById('sidebar');
 
@@ -87,7 +90,12 @@ class ServerManagerView {
 			showNotification: true,
 			betaUpdate: false,
 			silent: false,
-			lastActiveTab: 0
+			lastActiveTab: 0,
+			dnd: false,
+			dndPreviousSettings: {
+				showNotification: true,
+				silent: false
+			}
 		};
 
 		// Platform specific settings
@@ -95,6 +103,7 @@ class ServerManagerView {
 		if (process.platform === 'win32') {
 			// Only available on Windows
 			settingOptions.flashTaskbarOnMessage = true;
+			settingOptions.dndPreviousSettings.flashTaskbarOnMessage = true;
 		}
 
 		for (const i in settingOptions) {
@@ -155,6 +164,11 @@ class ServerManagerView {
 	}
 
 	initActions() {
+		this.initDNDButton();
+		this.$dndButton.addEventListener('click', () => {
+			const dndUtil = DNDUtil.toggle();
+			ipcRenderer.send('forward-message', 'toggle-dnd', dndUtil.dnd, dndUtil.newSettings);
+		});
 		this.$reloadButton.addEventListener('click', () => {
 			this.tabs[this.activeTabIndex].webview.reload();
 		});
@@ -179,6 +193,12 @@ class ServerManagerView {
 		this.sidebarHoverEvent(this.$settingsButton, this.$settingsTooltip);
 		this.sidebarHoverEvent(this.$reloadButton, this.$reloadTooltip);
 		this.sidebarHoverEvent(this.$backButton, this.$backTooltip);
+		this.sidebarHoverEvent(this.$dndButton, this.$dndTooltip);
+	}
+
+	initDNDButton() {
+		const dnd = ConfigUtil.getConfigItem('dnd', false);
+		this.toggleDNDButton(dnd);
 	}
 
 	getTabIndex() {
@@ -324,6 +344,15 @@ class ServerManagerView {
 				}
 			});
 		});
+
+		ipcRenderer.on('toggle-dnd', (event, state, newSettings) => {
+			this.toggleDNDButton(state);
+			ipcRenderer.send('forward-message', 'toogle-silent', newSettings.silent);
+			const selector = 'webview:not([class*=disabled])';
+			const webview = document.querySelector(selector);
+			const webContents = webview.getWebContents();
+			webContents.send('toggle-dnd', state, newSettings);
+		});
 	}
 
 	destroyTab(name, index) {
@@ -388,6 +417,12 @@ class ServerManagerView {
 		} else {
 			this.$sidebar.classList.add('sidebar-hide');
 		}
+	}
+
+	// Toggles the dnd button icon.
+	toggleDNDButton(alert) {
+		this.$dndTooltip.textContent = (alert ? 'Turn Off' : 'Enable') + ' Do Not Disturb';
+		this.$dndButton.querySelector('i').textContent = alert ? 'notifications_off' : 'notifications';
 	}
 
 	registerIpcs() {
--- a/app/renderer/js/utils/dnd-util.js
+++ b/app/renderer/js/utils/dnd-util.js
@@ -1,0 +1,41 @@
+'use strict';
+
+const ConfigUtil = require(__dirname + '/config-util.js');
+
+function toggle() {
+	const dnd = !ConfigUtil.getConfigItem('dnd', false);
+	const dndSettingList = ['showNotification', 'silent'];
+	if (process.platform === 'win32') {
+		dndSettingList.push('flashTaskbarOnMessage');
+	}
+
+	let newSettings;
+	if (dnd) {
+		const oldSettings = {};
+		newSettings = {};
+
+		// Iterate through the dndSettingList.
+		for (const settingName of dndSettingList) {
+			// Store the current value of setting.
+			oldSettings[settingName] = ConfigUtil.getConfigItem(settingName);
+			// New value of setting.
+			newSettings[settingName] = (settingName === 'silent');
+		}
+
+		// Store old value in oldSettings.
+		ConfigUtil.setConfigItem('dndPreviousSettings', oldSettings);
+	} else {
+		newSettings = ConfigUtil.getConfigItem('dndPreviousSettings');
+	}
+
+	for (const settingName of dndSettingList) {
+		ConfigUtil.setConfigItem(settingName, newSettings[settingName]);
+	}
+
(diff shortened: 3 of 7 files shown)


priyank-p:
@jainaman224 this PR is totally broken the button still does't work have you tested this?

jainaman224:
@cPhost Yes, it is working for me.

jainaman224:
@cPhost What were the initial settings of `Show Notification` and `Mute all sounds`?

app/main/menu.js
```diff
@@ -366,9 +378,19 @@ class AppMenu {
 		}];
 	}
 
+	static doNotDisturb() {
+		const newValue = !((ConfigUtil.getConfigItem('showNotification', true) ||
```
priyank-p: I don't think this is the right approch here the whole conditional on one line does seem like a good idea its very bug prone. 

Can you instead put the list of all the setting in and array like `var settingList  = ['showNotification', ..]` and then loop through it using `.map` in one liner? 
(Do let me if this is unclear)

app/renderer/js/main.js
```diff
@@ -152,6 +154,35 @@ class ServerManagerView {
 	}
 
 	initActions() {
+		this.initDNDButton();
+		this.$dndButton.addEventListener('click', () => {
+			const dnd = ConfigUtil.getConfigItem('dnd', false);
```
priyank-p: Once again this is very hard to review an trace back the where the bug is happing if the situation arises. This needs a lot of code refactoring. And this might explain why is not working for me locally.

priyank-p (changes requested):
@jainaman224 thanks for working on it, but i think the code need to better refactored and this could be made very simple. Currently is very hard to review and know whats going on and why? I do get the points of function but you can do better here.

priyank-p:
@jainaman224 the weird errors we were facing, were because you did not add `dnd` key in `ConfigUtil` so you can't get retrive it hence causing all the errors. You need to add default setting here https://github.com/zulip/zulip-electron/blob/master/app/renderer/js/main.js#L77

app/renderer/js/main.js
```diff
@@ -87,7 +89,14 @@ class ServerManagerView {
...
 			betaUpdate: false,
 			silent: false,
-			lastActiveTab: 0
+			lastActiveTab: 0,
+			flashTaskbarOnMessage: true,
+			dnd: false,
+			oldSettings: {
+				showNotification: false,
+				silent: true,
+				flashTaskbarOnMessage: false
```
priyank-p: `oldSettings` by default should be same as default.

app/main/menu.js
```diff
@@ -399,6 +411,11 @@ class AppMenu {
 		});
 	}
 
+	static doNotDisturb() {
```
priyank-p: Can you move this into `click` handler above

app/main/menu.js
```diff
@@ -211,6 +211,26 @@ class AppMenu {
...
+				click() {
+					const dnd = !ConfigUtil.getConfigItem('dnd', false);
+					const settingList = ['showNotification', 'silent', 'flashTaskbarOnMessage'];
+					let oldSettings;
+					if (dnd) {
+						const currentSettings = {};
+						oldSettings = {};
+						for (const settingName of settingList) {
+							currentSettings[settingName] = ConfigUtil.getConfigItem(settingName);
+							oldSettings[settingName] = (settingName === 'silent');
```
priyank-p: This right here seems bit hacky, so i'd add a comment here to explain what's going on for future maintaing.

priyank-p (commented):
So for functionality wise looks good @jainaman224! But can you move the handling of the dnd into its own util at `js/utils`. And also a minor comment.

jainaman224:
@cPhost Please review.

app/renderer/js/utils/dnd-util.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+const ConfigUtil = require(__dirname + '/config-util.js');
+
+function dnd() {
```
priyank-p: `dnd` -> `toogle` so `DNDUtil.toogle`

app/renderer/js/utils/dnd-util.js
```diff
@@ -0,0 +1,34 @@
...
+	this.initDNDButton();
```
priyank-p: This is from the old commit which should be removed, or i'd think you do need to implement the old thing here>

app/renderer/js/utils/dnd-util.js
```diff
@@ -0,0 +1,34 @@
...
+	const dnd = !ConfigUtil.getConfigItem('dnd', false);
+	const settingList = ['showNotification', 'silent', 'flashTaskbarOnMessage'];
+	let oldSettings;
```
priyank-p: Remove this from here and initialize variable inside the `if`/`else` statement.

app/renderer/js/utils/dnd-util.js
```diff
@@ -0,0 +1,34 @@
...
+			currentSettings[settingName] = ConfigUtil.getConfigItem(settingName);
+			// Desired value of setting.
+			oldSettings[settingName] = (settingName === 'silent');
+		}
+		// Store current value in oldSettings.
+		ConfigUtil.setConfigItem('oldSettings', currentSettings);
+	} else {
+		oldSettings = ConfigUtil.getConfigItem('oldSettings');
+	}
+	for (const settingName of settingList) {
```
priyank-p: new line here!

app/renderer/js/utils/dnd-util.js
```diff
@@ -0,0 +1,34 @@
...
+		ConfigUtil.setConfigItem(settingName, oldSettings[settingName]);
+	}
+	ConfigUtil.setConfigItem('dnd', dnd);
```
priyank-p: Ditto!

app/renderer/js/utils/dnd-util.js
```diff
@@ -0,0 +1,34 @@
...
+const ConfigUtil = require(__dirname + '/config-util.js');
+
+function dnd() {
+	this.initDNDButton();
+	const dnd = !ConfigUtil.getConfigItem('dnd', false);
+	const settingList = ['showNotification', 'silent', 'flashTaskbarOnMessage'];
+	let oldSettings;
+	if (dnd) {
+		const currentSettings = {};
+		oldSettings = {};
```
priyank-p: I'd do change the `currentSetting` -> `oldSetting` and `oldSettings` to `newSettings` or `dndSetting`.  They are used in way seems confusing.

priyank-p (commented):
Left a bunch of comments again

app/renderer/js/utils/dnd-util.js
jainaman224: I think, we need to initialize the button.

priyank-p: Yeah, so you need to do this here :)

app/renderer/js/utils/dnd-util.js
jainaman224: I have to initialize it outside if/else. Due to scoping issue. It is inaccessible outside the block if I initialize it inside the conditional block.

priyank-p: Ignore this, i was think of change of variable and stuff :)

priyank-p (approved):

jainaman224:
@akashnimare Please review.

app/renderer/js/utils/dnd-util.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+const ConfigUtil = require(__dirname + '/config-util.js');
+
+function toggle() {
+	const dnd = !ConfigUtil.getConfigItem('dnd', false);
+	let settingList;
```
akashnimare: We can re-write this as - 
```javascript
const settingList = ['showNotification', 'silent'];
    settingList.push('flashTaskbarOnMessage')
}
```
maybe rename `settingList` to `dndSettingList`, cc @cPhost.

akashnimare (changes requested):

app/renderer/js/main.js
```diff
@@ -87,7 +90,12 @@ class ServerManagerView {
 			showNotification: true,
 			betaUpdate: false,
 			silent: false,
-			lastActiveTab: 0
+			lastActiveTab: 0,
+			dnd: false,
+			oldSettings: {
+				showNotification: true,
```
akashnimare: I think this should inherit from the main settings, if it's not null. Something like this - 
```javascript
oldSettings: {
	showNotification: ConfigUtil.getConfigItem('showNotification') || true,
	silent: ConfigUtil.getConfigItem('silent') || false
}

```

akashnimare (changes requested):

app/renderer/js/main.js
jainaman224: @akashnimare What are the advantages?

I think it should be like this.
```js
oldSettings: {
	showNotification: (ConfigUtil.getConfigItem('showNotification') === null) || true,
	silent: (ConfigUtil.getConfigItem('silent') === null) || false
}
```

app/renderer/js/utils/dnd-util.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+
+const ConfigUtil = require(__dirname + '/config-util.js');
+
+function toggle() {
+	const dnd = !ConfigUtil.getConfigItem('dnd', false);
+	let dndSettingList;
+	if (process.platform === 'win32') {
```
priyank-p: This if else block can be simplified to be
```javascript
const dndSettingList = ['showNotification', 'silent'];
if (process.platfrom == 'win32') {
  dndSettingList.push('flashTaskbarOnMessage');
}
```

app/renderer/js/utils/dnd-util.js
```diff
@@ -0,0 +1,43 @@
...
+		dndSettingList = ['showNotification', 'silent', 'flashTaskbarOnMessage'];
+	} else {
+		dndSettingList = ['showNotification', 'silent'];
+	}
+	let newSettings;
+
```
priyank-p: Remove the space and add it above!

priyank-p (commented):
Can you fix this basic styling issues with your code thanks!

app/renderer/js/main.js
```diff
@@ -87,7 +90,12 @@ class ServerManagerView {
 			showNotification: true,
 			betaUpdate: false,
 			silent: false,
-			lastActiveTab: 0
+			lastActiveTab: 0,
+			dnd: false,
+			oldSettings: {
```
akashnimare: @cPhost, @jainaman224 can we make it more readable? I was thinking to use something like "dndState", "dndOldState", "dndPreviousSettings".

akashnimare (changes requested):

app/renderer/js/main.js
```diff
@@ -87,7 +90,12 @@ class ServerManagerView {
...
+				showNotification: ConfigUtil.getConfigItem('showNotification') || true,
```
jainaman224: @cPhost @akashnimare This statement gives true always. So, we should go for conditional statement.

```js
showNotification: ConfigUtil.getConfigItem('showNotification') === null ? true : ConfigUtil.getConfigItem('showNotification'),
```

app/renderer/js/main.js
priyank-p: Yeah, we can this make's it more clear. (Not that it was not clear before, but yeah this is great)

app/renderer/js/main.js
priyank-p: It shouldn't return true if its disabled in settings. It returns `true` since the default it true.

priyank-p: To be clear this doesn't need a check!

app/renderer/js/main.js
jainaman224: I have changed it to `dndPreviousSettings`.

app/renderer/js/main.js
priyank-p: Err! The way it is is fine.

akashnimare (approved):

akashnimare:
@jainaman224 great work so for. We just need to add two more things here -
* Document the shortcuts (CTRL+SHIFT+M) on our shortcuts page.
* Update the content of tooltip depending upon the state of DND.

jainaman224:
@akashnimare I have made required changes. Please review this PR.

jainaman224:
@abhigyank It is used for the maintaining similarity across codebase. Have a look at https://github.com/zulip/zulip-electron/blob/master/app/renderer/js/main.js#L307.

priyank-p:
@jainaman224 the commit-msg needs to be fixed, `feature` instead of `design` and you need to remove the testing description from commit msg, generally, you should be describing what the commit is about unless it a test.

jainaman224:
@cPhost @akashnimare I have made required changes. Please review.

app/renderer/js/utils/dnd-util.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
+const ConfigUtil = require(__dirname + '/config-util.js');
+
+function toggle() {
+	const dnd = !ConfigUtil.getConfigItem('dnd', false);
+	const dndSettingList = ['showNotification', 'silent'];
+	if (process.platform === 'win32') {
+		dndSettingList.push('flashTaskbarOnMessage');
```
cPhost: The last thing to be done here is we also need to do this at -  https://github.com/zulip/zulip-electron/blob/master/app/renderer/js/main.js#L95 otherwise the flash feature is broken on windows.

jainaman224:
@cPhost Made the required changes. Please review.

priyank-p:
@jainaman224 thanks! @akashnimare can you cross-check that this fixes the flashing on windows.

akashnimare:
@jainaman224 there is something wrong with the DND functionality. The settings (Desktop notifications, silent) are changing automatically if you change it and then switch back to a server. 

dnd

akashnimare:
I can reproduce this bug in the master as well.

abhigyank:
I test this on Ubuntu 16.04. 
This doesn't work for me.  
1. I think  "Show Desktop Notifications" is broken in the master itself. Turning off "Show Desktop Notifications" still gets me notifications even on the master.
2. The DND button's functionality doesn't work for me anyway. It changes the settings for me alright in the settings window, but doesn't changes the functionality. It still works on the previous functionality. 
With "Mute all sounds" off initially, turning DND on, I still get sound, while with "Mute all sounds" on initially, turning DND on, I don't get sound. So it seems, it shows the previous behavior. 
3. I think if someone manually changes the settings to the DND position, the DND button should change its icon to mute.
Can others replicate this?

jainaman224:
Point 1, it was fixed by Akash as far as I know.
Point 3, It was discussed with team and we are planning to work on it next.
Point 2, I will test it again. Thanks for testing

akashnimare:
@jainaman224 can you rebase and fix the merge conflicts?

akashnimare:
This is what I'm experiencing on macOS -
* Fresh install + added `czo` server
* Clicking the DND button activates the DND mode but I still get the sound no notification though
* After hard reloading, it works as expected (no notification and no sound)
* To disable the DND mode I still need to hard reload otherwise (notification + no sound)

abhigyank:
I did a Fresh install + added 'czo' server too.
* No longer having the "Show Desktop Notifications" issue.
* I can corroborate @akashnimare's experience on Ubuntu 16.04 too.
* Adding to it, with "Mute all sounds" off initially, turning DND on, I still get sound (no notification though), while with "Mute all sounds" on initially, turning DND on, I don't get sound (no notification). So it seems, it shows the previous behavior for sound.
* Same with disable. Hard reload solves both issues.

akashnimare:
Merged in https://github.com/zulip/zulip-electron/commit/537fbe8f9e0f0dbb573ac4ffe140c0a580d17f24 after minor change.

Issue #418, opened 2018-02-15
design: Do not create a new instance of setting tab.


akashnimare:
On clicking the `+` and `Setting` icon we should not create a new instance of "setting icon" instead we should activate and focus the bottom setting icon.

![image](https://user-images.githubusercontent.com/2263909/36255426-fb983e4e-1274-11e8-9ec6-33810bb0feef.png)

abhigyank:
@zulipbot claim

PR #422, opened 2018-02-18
design: Not create a new instance of setting tab

abhigyank:
Fixes: #418
The functionality works as expected, need review on approach.
Approach - For Settings functional tab, it is not appended to the sidebar and to the `windows menu`. The settings icon at bottom is given an active tab which is removed when some other tab is activated.
 
  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/css/main.css
+++ b/app/renderer/css/main.css
@@ -113,6 +113,14 @@ body {
     color: #6c8592;
 }
 
+.action-button.active {
+    background-color: rgba(255, 255, 255, 0.25);
+}
+
+.action-button.active i {
+    color: #eee;
+}
+
 .tab:first-child {
     margin-top: 8px;
 }
--- a/app/renderer/js/main.js
+++ b/app/renderer/js/main.js
@@ -119,6 +119,8 @@ class ServerManagerView {
 			}
 			// Open last active tab
 			this.activateTab(ConfigUtil.getConfigItem('lastActiveTab'));
+			// Remove focus from the settings icon at sidebar bottom
+			this.$settingsButton.classList.remove('active');
 		} else {
 			this.openSettings('Servers');
 		}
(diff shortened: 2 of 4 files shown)


app/renderer/css/main.css
```diff
@@ -101,6 +101,14 @@ body {
     color: #98a9b3;
 }
 
+.action-button.active {
+    background-color: rgba(255, 255, 255, 0.25);
+}
+
+.action-button.active i{
```
priyank-p: nit: space after `i`

priyank-p (commented):
Work great locally, left one comment also can we change the commit msg up a bit this does not do anything to design in any way i would rather prefer the the commit msg have something like `settings-tab: ` than design

priyank-p:
This works out great, and the one thing i like about this PR is that the settings web view is not destroyable so the the time it takes to reopen the webview is instant which IMO is great. Although i wish we could somehow move the setting btn/icon on top (i think we just need to show the btn on top since we can't move it).

akashnimare:
@abhigyank can you resolve the merge conflicts? Also, is this ready for review?

abhigyank:
@akashnimare Done! Yes it is ready for review.

akashnimare:
Works great. Just noticed a bug. When you add or remove a server the `setting` page should not be active.
 
image

akashnimare:
Looks good. The other bug I have noticed is when you click on setting page, `Window` menu shows the first server as checked but at this time it shouldn't.
Repro - 
* Add two Zulip servers
* Open the second server
* Click on settings
* Check `Window` menu 

image

app/renderer/js/main.js
```diff
@@ -280,6 +282,9 @@ class ServerManagerView {
 			if (this.activeTabIndex === index) {
 				return;
 			} else if (hideOldTab) {
+				if (this.tabs[this.activeTabIndex].props.role === 'function' && this.tabs[this.activeTabIndex].props.name === 'Settings') {
```
akashnimare: Can you document a bit here and in `menu.js` about the code? Other than this it's good to merge after squashing and making that checkbox change.

akashnimare:
@abhigyank merged, thanks 🚀

PR #424, opened 2018-02-20
CPU Usage optimization (basic)

priyank-p:
This can be merged without much testing @akashnimare this does some basic require optimization.

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -1,9 +1,7 @@
 'use strict';
 const path = require('path');
 const electron = require('electron');
-const windowStateKeeper = require('electron-window-state');
 const isDev = require('electron-is-dev');
-const appMenu = require('./menu');
 const { appUpdater } = require('./autoupdater');
 const { crashHandler } = require('./crash-reporter');
 
@@ -50,6 +48,8 @@ const iconPath = () => {
 };
 
 function createMainWindow() {
+	const windowStateKeeper = require('electron-window-state');
+
 	// Load the previous state with fallback to defaults
 	const mainWindowState = windowStateKeeper({
 		defaultWidth: 1000,
@@ -145,6 +145,7 @@ app.on('activate', () => {
 });
 
 app.on('ready', () => {
+	const appMenu = require('./menu');
 	appMenu.setMenu({
 		tabs: []
 	});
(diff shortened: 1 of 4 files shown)


akashnimare:
Do we have any before and after data of CPU usage and performance? Let me check if this approach makes any difference.

timabbott:
@cPhost for context, https://github.com/zulip/zulip/pull/6965 has some interesting data and research, which we need to test more.  

It seems likely the actual problem is something in how we have Electron configured; the app itself doesn't do a lot of work beyond what the Zulip webapp does (and the Zulip webapp in Chrome doesn't have these perf problems).

priyank-p:
@timabbott https://github.com/zulip/zulip/pull/6965 that does provide good debugging but i don't know i want to take that now, it could be easily that the chrome version was old (Electron is always 1-2 versions behind because its hard to upgrade.)

1. Currently efforts are for optimizing node's require since file operations are very expensive specially in old computer.

2. Optimize the electron webapp wrapper like avoiding upating DOM too much etc.

(But yeah animations/css optimization are on my list)

timabbott:
@cPhost I think it's an actual bug in Electron (or maybe its configuration).  Like, closing the hotkeys modal takes 600ms in the Electron app on my very fast laptop, and a negligible amount of time in Chrome on my laptop.  They should be basically the same browser.  

I don't think working on little optimizations is the right approach here; there's something qualitatively wrong that needs investigation.

priyank-p:
> I think it's an actual bug in Electron (or maybe its configuration)

Yep, totally agree with that (its not just our app every major electron app like vscode, atom, slack).

> Like, closing the hotkeys modal takes 600ms in the Electron app on my very fast laptop, and a negligible amount of time in Chrome on my laptop. They should be basically the same browser.

We can report this to upstream and get it fixed. 

> I don't think working on little optimizations is the right approach here; there's something qualitatively wrong that needs investigation.

This right here is hard, this optimization are for making the app faster and are best practices (of web in particular) we should follow then to squash out every bit of performance we can. This is not being debug anywhere i think there are many issues opened for this never saw any updates on any of them 

Anyway if you are interested here is what electron issue https://github.com/electron/electron/issues/5672 and this comment https://github.com/electron/electron/issues/5672#issuecomment-222053637

Issue #421, opened 2018-02-18
Only show the back button if it needed


priyank-p:
Right now clicking on the go-back button sometimes does nothing so we can hide the button when not in need you can check of you can go back on an active webview by `webview.canGoBack()`.

This is a followup of #413, @abhigyank can you do this?

cPhost:
@zulipbot label "Priority: Medium", "Type: Enhancement"

abhigyank:
Sure, I can take this up. 
I think showing and hiding the back button, would not look good from design perspective. Rather I feel, we can change the shade of color of the back button to suggest inactiveness. Like how it is handled in browsers like chrome. What do you think?

priyank-p:
@abhigyank i do prefer hiding the button than disabling it since the icon is too big so its not working i rather not have it in my way as a user but feel free to get other opinion on it.

abhigyank:
I get your view. Maybe  open a thread on this in design stream?

akashnimare:
I don' think we need to hide the button, just disabaling the cursor and `0.6` opacity will work.

![image](https://user-images.githubusercontent.com/2263909/36382581-746b229a-15af-11e8-85e3-304bb2767871.png)

priyank-p:
Do we need it up though and not be functional, specially with a new button about to be added and this button being completely being useless most of the time it just adds to clutter for no reason.

akashnimare:
I'm 100% sure that this button will be visible most of the time if a user is using the app. We are planning to add a `hide/show` kind of button to deal with the clutter thing. Something like - 

<img src="https://d2ffutrenqvap3.cloudfront.net/items/281S2Q381l123N3S0w0e/Screen%20Recording%202018-02-13%20at%2002.39%20AM.gif?v=bca79b43" />

But let's try the first approach i.e. show it only if `webview.canGoBack()` and see how it looks.

PR #426, opened 2018-02-21
Only show the back button if it needed.

abhigyank:
Need review.

Everything works well, except a small issue described [here](https://chat.zulip.org/#narrow/stream/16-electron/topic/zulip-electron.20.2F.20master).

Address [issue 421](https://github.com/zulip/zulip-electron/issues/421)

### DIFF
--- a/app/renderer/js/components/webview.js
+++ b/app/renderer/js/components/webview.js
@@ -69,6 +69,18 @@ class WebView extends BaseComponent {
 			this.props.onTitleChange();
 		});
 
+		this.$el.addEventListener('did-navigate-in-page', event => {
+			const isSettingPage = event.url.includes('renderer/preference.html');
+			if (isSettingPage) {
+				return;
+			}
+			this.canGoBackButton();
+		});
+
+		this.$el.addEventListener('did-navigate', () => {
+			this.canGoBackButton();
+		});
+
 		this.$el.addEventListener('page-favicon-updated', event => {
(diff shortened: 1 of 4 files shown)


abhigyank:
@zulipbot label "needs review"

priyank-p:
@abhigyank you need to used [`did-navigate-in-page`](https://electronjs.org/docs/api/webview-tag#event-did-navigate-in-page) and also [`did-navigate`](https://electronjs.org/docs/api/webview-tag#event-did-navigate) too event of web-view to track of the state, to know weather its time to update the button or not checking it on first time does work.

Edit: Just don't use page-title-update event.

app/renderer/js/components/webview.js
```diff
@@ -69,6 +69,17 @@ class WebView extends BaseComponent {
 			this.props.onTitleChange();
 		});
 
+		this.$el.addEventListener('did-navigate-in-page', event => {
+			if (event.url.includes('renderer/preference.html')) {
```
akashnimare: We could use it like - 
```
const isSettingPage = event.url.includes('renderer/preference.html');
if (isSettingPage) {
   return;
}
```

akashnimare (changes requested):

akashnimare (approved):

akashnimare:
Merged. Thanks, @abhigyank.

PR #428, opened 2018-02-25
electron-bridge: Implement electron bridge.

priyank-p:


priyank-p:
Side-effect of force-pushing, reopening..

app/main/index.js
```diff
@@ -224,6 +224,10 @@ app.on('ready', () => {
 	ipcMain.on('toggleAutoLauncher', (event, AutoLaunchValue) => {
 		setAutoLaunch(AutoLaunchValue);
 	});
+
+	ipcMain.on('realm-icon-changed', (event, serverURL, iconURL) => {
+		mainWindow.webContents.send('update-realm-icon', serverURL, iconURL);
```
akashnimare: `mainWindow.webContents` should be `page`.

app/renderer/js/electron-bridge.js
```diff
@@ -0,0 +1,44 @@
+const events = require('events');
+const { ipcRenderer } = require('electron');
+
+// we have and will have some non camelcase stuff
+// while working with zulip so just turning the rule off
+// for the wole file.
```
akashnimare: Typo - *whole

akashnimare (changes requested):

akashnimare:
This looks good. Can you fix above minor changes and resolve conflicts?

priyank-p:
Fixed, and rebased.

akashnimare:
@priyankp10 I'm getting following errors -

image

This doesn't let me update the icon for the first time only, later I can see the changes.

Reproduce -
* Run Zulip server
* Start Zulip app
* Add http://localhost:9991
* Try to update the realm icon

akashnimare:
I think the bug lies in the index of the tab -
```js
const serverImgSelector = `.tab[data-tab-id="${index}"] .server-icons`;`
```
The `data-tab-id` will keep changing as you navigate in the app. Maybe get the server tab using the active class?

priyank-p:
Yeah, using `.active` is the right solution here. Although the `data-user-id` changing is a bug we would need to fix right?

abhigyank:
@priyankp10  Hey, Akash and me met at the Zulip retreat so we decided to complete the work on this PR so that we can make use of the `electron_bridge` feature. Thanks for implementing this! :rocket:

priyank-p:
Done in #516.

PR #435, opened 2018-03-02
Menu: Improve Help submenu

abhigyank:
**What's this PR do?**

- Change `Show App Logs` path to navigate to `Logs` subfolder in `userData`.
- Add `What's New` to changelog of current release version.

### DIFF
--- a/app/main/menu.js
+++ b/app/main/menu.js
@@ -121,6 +121,12 @@ class AppMenu {
 				enabled: false
 			},
 			{
+				label: `What's New...`,
+				click() {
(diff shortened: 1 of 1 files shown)


akashnimare:
I kept the `userData` path intentionally instead of `/Logs` since we don't have much data in logs and we are not logging more often. That help item is mainly for those users who quickly want to share app config files etc. We can probably revisit this in future but for now, let's use the `userData` path only.

abhigyank:
Okay.  I'll remove the logs commit. Is the other commit of "What's new" fine?

akashnimare:
Merged. Thanks, @abhigyank.

PR #436, opened 2018-03-03
[WIP] Add i18n support

### LINKED ISSUE #407: i18n support

ihsavru:
Fix #407

### DIFF
--- a/app/translations/en.json
+++ b/app/translations/en.json
@@ -1,4 +1,46 @@
 {
+	"File": "File",
+	"About Zulip": "About Zulip",
+	"Desktop App Settings": "Desktop App Settings",
+	"Keyboard Shortcuts": "Keyboard Shortcuts",
+	"Reset App Settings": "Reset App Settings",
+	"Log Out": "Log Out",
+	"Quit": "Quit",
+
     "Edit": "Edit",
-    "Undo": "Undo"
+    "Redo": "Redo",
+    "Undo": "Undo",
+    "Cut": "Cut",
+    "Copy": "Copy",
+    "Paste": "Paste",
+    "Paste and Match Style": "Paste and Match Style",
+    "Delete": "Delete",
+    "Select All": "Select All",
+
+    "View": "View",
+    "Reload": "Reload",
+    "Hard Reload": "Hard Reload",
+    "Toggle Full Screen": "Toggle Full Screen",
+    "Zoom In": "Zoom In",
+    "Zoom Out": "Zoom Out",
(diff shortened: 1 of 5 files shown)


app/translations/en.json
```diff
@@ -0,0 +1,46 @@
+{
```
akashnimare, line 1: @ihsavru I think we should short this strings alphabetically so that it becomes easier to add/update translation in future.

akashnimare (changes requested):

akashnimare:
One important thing we need to add here is the ability to update the menu items depending on the server language. The menu items don't change/update when you switch servers so for an example if the server (a) has the server language as 'en' the menu items should be translated to 'en' and when user switches to the server (b) which has 'de' as server lang, menu items should be translated to 'de' but in current PR it remains same and doesn't change. We probably need to re-create the menu items when user changes the server language changes.

app/translations/i18n.js
```diff
@@ -0,0 +1,32 @@
...
+const ConfigUtil = require(__dirname + '/../renderer/js/utils/config-util.js');
+
+class i18n {
+	constructor() {
+		this.app = electron.app ? electron.app : electron.remote.app;
+		this.appLanguage = ConfigUtil.getConfigItem('spellcheckerLanguage');
+		this.appLocale = path.join(__dirname, this.appLanguage + '.json');
+
+		if (fs.existsSync(this.appLocale)) {
+			this.loadedLanguage = JSON.parse(fs.readFileSync(this.appLocale, 'utf8'));
```
priyank-p, line 14: Right here if we add a check for `isDev` and if not in development we `require` the json it would be cached by node and optimized by electron which will be faster.

cc @akashnimare

akashnimare:
@ihsavru you could have just rebased and fixed the merge conflicts :)

ihsavru:
@akashnimare I think it's not required now that #437 is merged because it already includes the changes i made :sweat_smile:

akashnimare:
No, it only provides the translation files. We still need to add the i18n script and then update `menu.js`.

ihsavru:
@akashnimare oh alright

akashnimare:
Yeah, so the things you need to do here -
* Rebased
* Fix merge conflicts
* update menu.js with i18n
* Update settings locales 
* Find a way to update the menu items translation (and setting) when a user switches servers.

akashnimare:
@ihsavru let me know if you're interested in finishing this PR.

PR #441, opened 2018-03-11
add missing debian/ubuntu dependancy

blinfoldking:
---

**Adding libgconf-2-4**

simply adding libgconf-2-4 into the debian/ubuntu dependancies install script

### DIFF
--- a/development.md
+++ b/development.md
@@ -25,7 +25,7 @@ manager (see [here][nodesource-install] for more on the first command):
 
 ```sh
 $ curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash -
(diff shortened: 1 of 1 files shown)


akashnimare:
@BlinfoldKing I'm not sure if we need this dependency.

blinfoldking:
i've got an error libconf-2-4 error when running the development install, i've also try this in other machine.

my machine runs ubuntu 17.10

it's also been refrencing in https://github.com/electron/electron-quick-start/issues/186

akashnimare:
Hmm, correct. Looks like this dependency is required for `Ubuntu > 16.10`.

blinfoldking:
it seems so

akashnimare:
@BlinfoldKing merged. Thanks!

PR #452, opened 2018-03-22
report-issue: Add report issue UX using send-feedback electron element.

priyank-p:
**What's this PR do?**
This uses @electron-elements/send-feedback package to easily implementUX for reporting issues.

**Screenshots?**
screen shot 2018-03-21 at 8 36 15 pm

### DIFF
--- a/app/main/menu.js
+++ b/app/main/menu.js
@@ -1,5 +1,4 @@
 'use strict';
-const os = require('os');
 const path = require('path');
 
 const { app, shell, BrowserWindow, Menu, dialog } = require('electron');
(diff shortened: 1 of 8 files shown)


abhigyank:
One problem here could be that `send-feedback` module uses [`promisify`](https://nodejs.org/api/util.html#util_util_promisify_original) which has been introduced in node v8.0.0 but currently we have downgraded to node6 in this 028bc02, this could break the app.

priyank-p:
@abhigyank that's not a problem electron uses its own internal embedded node version which should be greater than node 7 since electron `v1.8.0` and we are currently on `v1.8.2` in here I think.

akashnimare:
I think we need to clear the contents once the feedback is sent.

PR #455, opened 2018-03-22
security-fix: Prevent drag and drop events.

akashnimare:
This PR fixes a security bug which was caused by drag and drop events. Anyone can execute a remote code if a user drag and drops that link to the app. This PR prevents both the events.

### DIFF
--- a/app/renderer/main.html
+++ b/app/renderer/main.html
@@ -44,4 +44,5 @@
   </div>
 </body>
 <script src="js/main.js"></script>
+<script>require('./js/shared/preventdrag.js')</script>
 </html>
--- a/app/renderer/about.html
+++ b/app/renderer/about.html
@@ -1,45 +1,50 @@
 <!DOCTYPE html>
 <html lang="en">
-	<head>
-		<meta charset="UTF-8">
-		<link rel="stylesheet" href="css/about.css">
-	</head>
-	<body>
+
+<head>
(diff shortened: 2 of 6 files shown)


app/renderer/main.html
```diff
@@ -44,4 +44,5 @@
   </div>
 </body>
 <script src="js/main.js"></script>
+<script>require('./js/shared/preventdrag.js')</script>
 </html>
```
abhigyank, line 48: New line at end of file.

abhigyank:
Everything looks good. It works correctly as expected.
One thing I would like to point here is that, drag and drop in compose box works fine, all other drag and drop has been disabled i.e. it now won't work for dropping url at "Add Organisation", text at Search box, email/password at log in and search people.   
I think this is fine considering the trade off was a major security issue.

akashnimare:
> it now won't work for dropping url at "Add Organisation", text at Search box, email/password at log in and search people.

I think all these are acceptable and the use case is very very low so we are good here (If this works for compose box we can make it work for search box etc also this might need few tweaks though).

priyank-p:
To make it works for search box or anything all we need is we need to check `event.target` and only prevent default if it's not search box or whatever have you.

akashnimare:
I checked it again, the `drag` is not enabled in the `search-box`.

PR #460, opened 2018-03-22
"Organizations" settings page fixes and tweaks

### LINKED ISSUE #456: Tweaks to the "Organizations" settings page
* ~~We should have a link to "Add organization", in case users don't see the `+` icon; probably at the bottom is fine.~~ (Decided not to do this)
* ~~It's not great that the hover CSS moves the row (e.g. the "Disconnect" button jumps)~~
* ~~I think we want the "Disconnect" button to be at the far right, so that longer organization URLs don't mess up the table~~:

image

Update - I think we could still make that page responsive. 

image

abhigyank:
* Remove right transition on hover on settings cards.
* Move disconnect button to extreme right using flex.
* Text-overflow for server urls using ellipsis.

Partly fixes #456 .

### DIFF
--- a/app/renderer/css/preference.css
+++ b/app/renderer/css/preference.css
@@ -198,7 +198,11 @@ img.server-info-icon {
 
 .server-info-right {
     margin-top: 4px;
-    min-width: 55%;
+    width: 55%;
+    display: flex;
+    align-items: baseline;
+    justify-content: space-between;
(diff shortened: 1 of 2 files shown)


akashnimare:
Merged. Thanks, @abhigyank. I'm not closing the issue since we still need to work on the responsive part.

PR #461, opened 2018-03-23
org-settings: Responsive UI for connected orgs in smaller window sizes.

abhigyank:
For window sizes smaller than 800px UI changes for list of connected organisation

Fixes: #456.

Screenshot : -
screenshot from 2018-03-23 20-09-11


akashnimare:
The UI breaks down at ~900 px so we should check the width against the same.

image

akashnimare:
@abhigyank this looks good for now but we should revisit this UI later. Merged.

Issue #459, opened 2018-03-22
Shortcuts menu should also advertise Zulip webapp's built-in hotkeys


timabbott:
I think we could just handle this by having a tip-section at the top or bottom of the in-app shortcuts page linking to:

https://zulipchat.com/help/keyboard-shortcuts

And by "tip" I mean something styled nicely lik https://zulipchat.com/help/view-the-markdown-source-of-a-message

abhigyank:
What should be the message in the tooltip?
I was thinking "You can view the webapp built-in hotkeys [here](https://zulipchat.com/help/keyboard-shortcuts)".

timabbott:
Maybe something more like "These hotkeys are in addition to <a>Zulip's web keyboard shortcuts</a>"?

We may want to do a bit of editing after seeing what it looks like implemented, so post a screenshot on the PR.

PR #468, opened 2018-03-29
shortcuts-menu: Add a tip-section to link to webapp's built-in hotkeys.

abhigyank:
Fixes: #459
**Screenshot**
screenshot from 2018-03-30 00-22-38

### DIFF
--- a/app/renderer/css/preference.css
+++ b/app/renderer/css/preference.css
@@ -557,6 +557,23 @@ input.toggle-round:checked+label:after {
     background: #329588;
 }
 
+.tip {
+    width: 70%;
+    background-color: hsl(46,63%,95%);
+    border: 1px solid hsl(46,63%,84%);
+    border-radius: 4px;
+    padding: 12px 30px;
+    margin: 10px 0 20px 0;
+}
+
+.md-14 {
+    font-size: 14px;
+}
+
+#open-hotkeys-link {
+    text-decoration: underline;
+    cursor: pointer;
+}
 
 /* responsive grid */
 
--- a/app/renderer/js/pages/preference/shortcuts-section.js
+++ b/app/renderer/js/pages/preference/shortcuts-section.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const BaseSection = require(__dirname + '/base-section.js');
+const shell = require('electron').shell;
 
 class ShortcutsSection extends BaseSection {
 	constructor(props) {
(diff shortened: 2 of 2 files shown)


timabbott:
Visually, looks great to me (though maybe we should try putting it at the top and see how it goes).  I would call them "Keyboard shortcuts", not "hotkeys", for consistency.

Tagging @rishig for an opinion on the label text; I'm not sure it's quite right.

rishig:
Yeah, I think putting on top makes sense. Though would it possibly make sense just to duplicate the web shortcuts information here?

rishig:
Or maybe we can frame it as: "Zulip's shortcuts are primarily documented [here]. In addition, .."

akashnimare:
@rishig would it make sense to append app's shortcut into the web app keyboard shortcut's popup (and remove the shortcut page from app's setting page). We could introduce a tab like **Desktop shortcut** or something. This is because we have two shortcuts within the app one is in the Menu item which opens the webapp shortcut popup and the other one is in the desktop app shortcut's page which only shows the app shortcuts. We just don't want user to get confused between both the shortcuts. We could also add all the webapp shortcuts (that popup) into the app's shortcut page.

timabbott:
We can't duplicate the webapp shortcuts here, because the user may be connected to multiple organizations which might have different Zulip versions with different sets of hotkeys.  So I think it's better to not try to inline that data into the desktop app.

rishig:
I see. Would it be possible to go in the other direction, as Akash suggested? E.g. dynamically embed the desktop shortcuts into the webapp keyboard shortcuts help? 

As a quick solution though, we can put this at the top:
 "Zulip's shortcuts are primarily documented [here]. A few additional shortcuts are documented below."

Also, most of the current shortcuts (cut/copy/paste, zoom, etc) probably work in the webapp too, for all common browsers? So arguably we could add them to the webapp shortcut list, though I think we'd need a better UI for displaying them.

timabbott:
I think we should not block integrating the link here on doing some more complicated with trying to duplicate shortcuts.  I think it's actually kinda a pain to do something nice with that, and I think 90% of the value is just having it be clear that there's two sets (especially since they basically are for different things -- e.g. one is for controlling the app in terms of things like what org you're in).

For the text, I think this is fine:

"These desktop app shortcuts extend the <a>Zulip webapp's keyboard shortcuts</a>"

akashnimare:
@timabbott make sense. @abhigyank can you update the text so that we can merge this.

akashnimare:
Merged in https://github.com/zulip/zulip-electron/commit/ceaf13dee2fbae479b51f528359f6d58bc39ba1c.

PR #471, opened 2018-04-05
internal-links: open non-image links in hidden webview.

priyank-p:
This will make sure that the current server webview will
not reload due to URL change.

  - [X] macOS

### DIFF
--- a/app/renderer/js/components/handle-external-link.js
+++ b/app/renderer/js/components/handle-external-link.js
@@ -16,7 +16,7 @@ function handleExternalLink(event) {
 	if (isWhiteListURL) {
 		event.preventDefault();
 
-    // only open the pdf in hidden webview since opening the
+    // only open the  pdf, mp3, mp4 etc.. in hidden webview since opening the
     // image in webview will do nothing and will not save it
     // whereas the pdf will be saved to user desktop once openened in
     // in the hidden webview and will not trigger webview reload
--- a/app/renderer/js/components/handle-external-link.js
+++ b/app/renderer/js/components/handle-external-link.js
@@ -1,0 +1,36 @@
+const { shell } = require('electron').remote;
+const LinkUtil = require('../utils/link-util');
+const DomainUtil = require('../utils/domain-util');
+const hiddenWebView = require('../components/hidden-webview');
+
+function handleExternalLink(event) {
+	const { url } = event;
+	const domainPrefix = DomainUtil.getDomain(this.props.index).url;
+
+  // Whitelist URLs which are allowed to be opened in the app
+	const {
+    isInternalUrl: isWhiteListURL,
+    isUploadsUrl: isUploadsURL
+  } = LinkUtil.isInternal(domainPrefix, url);
+
+	if (isWhiteListURL) {
+		event.preventDefault();
+
+    // only open the pdf in hidden webview since opening the
+    // image in webview will do nothing and will not save it
+    // whereas the pdf will be saved to user desktop once openened in
+    // in the hidden webview and will not trigger webview reload
+		if (!LinkUtil.isImage(url) && isUploadsURL) {
+			hiddenWebView.loadURL(url);
+			return;
+		}
+
+    // open internal urls inside the current webview.
+		this.$el.loadURL(url);
+	} else {
+		event.preventDefault();
+		shell.openExternal(url);
+	}
+}
+
+module.exports = handleExternalLink;
--- a/app/renderer/js/components/webview.js
+++ b/app/renderer/js/components/webview.js
@@ -3,13 +3,12 @@
 const path = require('path');
 const fs = require('fs');
 
-const DomainUtil = require(__dirname + '/../utils/domain-util.js');
 const ConfigUtil = require(__dirname + '/../utils/config-util.js');
 const SystemUtil = require(__dirname + '/../utils/system-util.js');
-const LinkUtil = require(__dirname + '/../utils/link-util.js');
-const { shell, app, dialog } = require('electron').remote;
+const { app, dialog } = require('electron').remote;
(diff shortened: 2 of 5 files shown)


app/renderer/js/components/webview.js
```diff
@@ -55,9 +64,17 @@ class WebView extends BaseComponent {
 				url === domainPrefix + '/' ||
 				url.includes(domainPrefix + '/user_uploads/'); // URL containing the file attachments
 
+      const isImageURL = /\.(png|gif|bmp|webp|jpeg|jpg))/;
 			if (isWhiteListURL) {
 				event.preventDefault();
-				this.$el.loadURL(url);
+
+        // only open the pdf in hidden webview since opening the
```
abhigyank: This works not only for pdf but also for other formats like mp3, doc, mp4 etc.

abhigyank:
This works not only for pdf but also for other formats like mp3, doc, mp4 etc.

priyank-p:
Yeah, everything not an image, i forgot to update the commit message.
Since image does not get downloaded when clicked.

app/renderer/js/components/webview.js
```diff
@@ -12,6 +12,15 @@ const { shell, app, dialog } = require('electron').remote;
 const BaseComponent = require(__dirname + '/../components/base.js');
 
 const shouldSilentWebview = ConfigUtil.getConfigItem('silent');
+
+// this hidden webview will be used to open pdf url and
+// save it to user's computer without triggering a reload
+// when navigating to pdf url to download it.
+const hiddenWebView = document.createElement('webview');
```
akashnimare: @cPhost should we make a component for this like `hidden-webview.js` and then init it?

priyank-p: Depends: i don't see this needing more thing added to it so, nor it's too complicated, so i think it fine.

akashnimare: @cPhost should we make a component for this like `hidden-webview.js` and init it?

app/renderer/js/components/webview.js
```diff
@@ -55,9 +64,17 @@ class WebView extends BaseComponent {
 				url === domainPrefix + '/' ||
 				url.includes(domainPrefix + '/user_uploads/'); // URL containing the file attachments
 
+			const isImageURL = /\.(png|gif|bmp|webp|jpeg|jpg)/;
```
akashnimare: I think we should use the function here - `LinkUtil.isImage(url)`. I'll merge the other PR first so that you just need to rebase and use this function.

priyank-p: Yeah, makes sense.

app/renderer/js/components/webview.js
priyank-p: I will also put the function there which would make this more simpler.

app/renderer/js/utils/link-util.js
```diff
@@ -19,7 +19,11 @@ class LinkUtil {
 		const currentDomain = wurl('hostname', currentUrl);
 		const newDomain = wurl('hostname', newUrl);
 
-		return (currentDomain === newDomain) && newUrl.includes('/#narrow');
+		const sameDomainUrl = (currentDomain === newDomain || newUrl === currentUrl + '/');
+		const isInternalUrl = newUrl.includes('/#narrow') ||
+      newUrl.includes(currentUrl + '/user_uploads/');
```
akashnimare: @cPhost do we need this here? Since we already added this in the `handle-external-link.js`.

akashnimare (changes requested):

app/renderer/js/utils/link-util.js
priyank-p: I think i left it here since i think it makes sense here, ie if we use this somewhere else this should be standalone

priyank-p: `isInternal` sound like it should support this so, i am tempted to add it here.

app/renderer/js/utils/link-util.js
```diff
@@ -19,7 +19,14 @@ class LinkUtil {
 		const currentDomain = wurl('hostname', currentUrl);
 		const newDomain = wurl('hostname', newUrl);
 
-		return (currentDomain === newDomain) && newUrl.includes('/#narrow');
+		const sameDomainUrl = (currentDomain === newDomain || newUrl === currentUrl + '/');
+		const isUplodsUrl = newUrl.includes(currentUrl + '/user_uploads/');
```
akashnimare: Typo here too.

app/renderer/js/components/handle-external-link.js
```diff
@@ -0,0 +1,36 @@
...
+  // Whitelist URLs which are allowed to be opened in the app
+	const {
+    isInternalUrl: isWhiteListURL,
+    isUploadsUrl: isUploadsURL
+  } = LinkUtil.isInternal(domainPrefix, url);
+
+	if (isWhiteListURL) {
+		event.preventDefault();
+
+    // only open the pdf in hidden webview since opening the
```
akashnimare: Gotta update this to add other file types also i.e - pdf, mp3, mp4 etc.

app/renderer/js/utils/link-util.js
```diff
@@ -19,7 +19,14 @@ class LinkUtil {
...
+		const isInternalUrl = newUrl.includes('/#narrow') || isUplodsUrl;
+
+		return {
+			isInternalUrl: sameDomainUrl && isInternalUrl,
+			isUplodsUrl
```
akashnimare: Typo.

akashnimare (changes requested):

akashnimare (approved):

Issue #469, opened 2018-04-02
Security/UI Issue: Images and Attachments Must Open in Electron


stone212:
**Using:** Electron 1.9 for LInux

**To Reproduce:**

When you click on an image or uploaded file link (*not* the image thumbnail) within the Electron Client, electron parses this as a link.  Therefore it tries to open in the Browser.

This is a problem.  There is a reason I am not using my browser to view electron.  Also this is a potential security risk if you are using Electron through a proxy but use a browser through a different proxy.  It is also a security risk if your browser is less secure (many reasons a browser may be insecure including Apps, open pages, running javascript etc.) than electron client.

This is also a UI issue because I have to wait for the browser to open, or I have to go to a different monitor or I have to stop doing the browsing that I am doing in order to open an image/attachment.  So this is an annoyance also.

**Preferred Outcome:** 

Solution: Electron should allow you to save or open the file, using your system default application for viewing that type of application.

EDIT: The image thumbnail works great.  The image *link* and file *link* could work the same and that would solve this problem.

abhigyank:
If the image url in ends in an image format like `jpg` it'll indeed open in the app itself. However if it is a non image url it is bound to open in a browser. 
Can you share a sample url you're referring to? In case it is an image that doesn't have a image format url, we can address this in the webapp itself to treat that url also as an image.

akashnimare:
This is the intentional behaviour. We do open files in the app. Instead of clicking on an image link try clicking on the image itself. There you will get the option to download, open link etc.

<img src="https://d2ffutrenqvap3.cloudfront.net/items/0p2o330x3U2C3c0K2t0Q/Screen%20Recording%202018-04-02%20at%2004.10%20PM.gif?v=53d776c9"/>

stone212:
@akashnimare 

I may not share the URL since it is to a private system.  But!

If I click on the *thumbnail* then what happens is what you show in your gif.  However I never thought to do that.  The ticket is for clicking on the *link* so I was clicking on the word "filename.jpg" above the thumbnail.  (I didn't even think of the difference until I saw your gif.)

Please use the example in your gif.  Do not click on the thumbnail.  Please click on the filename "zulip-terminal.gif" and I think Electron will open firefox or some other image viewer to view the image.  I will update my original Issue post.

stone212:
@akashnimare Unrelated but your example GIF shows a CLI-based zulip client.  I need this in my life.  How do I get it?

stone212:
[...]

PR #472, opened 2018-04-05
internal-links: open image link in webapp lightbox.

abhigyank:
This will open the image in the webapp lightbox, shows the same behavior
that happens when clicking on the image preview.
Implemented almost the same way webapp handles opening lightbox on clicking image preview.

Fixes: #469.

Facing two lint errors:
1.`lightbox  is not defined.`
2. `$ is not defined.`
  - [x] Linux/Ubuntu

### DIFF
--- a/app/renderer/js/preload.js
+++ b/app/renderer/js/preload.js
@@ -4,6 +4,7 @@ const { ipcRenderer } = require('electron');
 const SetupSpellChecker = require('./spellchecker');
 
 const ConfigUtil = require(__dirname + '/utils/config-util.js');
+const LinkUtil = require(__dirname + '/utils/link-util.js');
 
 // eslint-disable-next-line import/no-unassigned-import
 require('./notification');
@@ -55,6 +56,24 @@ document.addEventListener('DOMContentLoaded', () => {
 			ipcRenderer.send('forward-message', 'reload-viewer');
 		});
 	}
+
+	// Open image attachment link in the lightbox instead of opening in the default browser
+	const { $, lightbox } = window;
+
+	$('#main_div').on('click', '.message_content p a', function (e) {
+		const url = $(this).attr('href');
+
+		if (LinkUtil.isImage(url)) {
+			const $img = $(this).parent().siblings('.message_inline_image').find('img');
+
+			// prevent the image link from opening in a new page.
+			e.preventDefault();
+			// prevent the message compose dialog from happening.
+			e.stopPropagation();
+
+			lightbox.open($img);
+		}
+	});
 });
 
 // Clean up spellchecker events after you navigate away from this page;
--- a/app/renderer/js/utils/link-util.js
+++ b/app/renderer/js/utils/link-util.js
@@ -21,6 +21,12 @@ class LinkUtil {
 
 		return (currentDomain === newDomain) && newUrl.includes('/#narrow');
 	}
+
+	isImage(url) {
+		// test for images extension as well as urls like .png?s=100
+		const isImageUrl = /\.(bmp|gif|jpg|jpeg|png|webp)\?*.*$/i;
+		return isImageUrl.test(url);
+	}
 }
 
 module.exports = new LinkUtil();


app/renderer/js/utils/link-util.js
```diff
@@ -21,6 +21,13 @@ class LinkUtil {
 
 		return (currentDomain === newDomain) && newUrl.includes('/#narrow');
 	}
+
+	isImage(url) {
+		const extension = url.substring(url.lastIndexOf('.') + 1).toLowerCase();
+		const imageExtensions = ['bmp', 'gif', 'jpg', 'jpeg', 'png', 'webp'];
```
priyank-p: You didn't need to think about each and every possible image type here zulip only supports four types of image uploads see:

https://github.com/zulip/zulip/blob/5740af27d668b2a2f9942da170183779b7ae3b28/static/js/upload_widget.js#L7-L12

But this is probably fine :)

priyank-p:
> Facing two lint errors:
1.lightbox is not defined.
> 2. $ is not defined.

Deconstruct them from window object, it should get rid of the errors

app/renderer/js/utils/link-util.js
abhigyank: Ah! I was searching that. I took from here https://github.com/zulip/zulip/blob/5740af27d668b2a2f9942da170183779b7ae3b28/static/js/markdown.js#L19 .
Can you verify which ones should be included?

priyank-p: Those are for markdown, upload_widget only accepts four types but it does not hurt to add them anyways so leave them there in fact i think i will be adding those in my PR so in case if we add some more image types in future we are safe.

abhigyank: Sure that'd be good.

app/renderer/js/preload.js
```diff
@@ -55,6 +56,23 @@ document.addEventListener('DOMContentLoaded', () => {
 			ipcRenderer.send('forward-message', 'reload-viewer');
 		});
 	}
+
+	const { $, lightbox } = window;
```
akashnimare: @cPhost, @abhigyank Not sure but should we use the JS here instead of the jQuery?

akashnimare: @abhigyank can you add few more lines here about what's going on here? Something like -
> Open image attachment link in the lightbox instead of opening in the default browser

akashnimare (changes requested):

app/renderer/js/preload.js
abhigyank: @akashnimare  Does that make any major difference?

I feel jQuery is fine. Makes it easier to implement and is in parallel to the webapp.

app/renderer/js/utils/link-util.js
priyank-p: Can you make this a regex check this right here is not some we would want: 
```javascript
const extension = url.substring(url.lastIndexOf('.') + 1).toLowerCase();
```

You can just do this instead:
```javascript
const isImageUrl = /\.(png|...)/;
return isImageUrl.test(url);
```

app/renderer/js/utils/link-util.js
```diff
@@ -21,6 +21,13 @@ class LinkUtil {
 
 		return (currentDomain === newDomain) && newUrl.includes('/#narrow');
 	}
+
+	isImage(url) {
```
priyank-p, line 25: This should be `isImageURL` right?
cc @akashnimare @abhigyank

priyank-p (changes requested):

app/renderer/js/utils/link-util.js
abhigyank: One issue with this is that, if `.png` etc will come even in the middle of the url (which could be a rare case) then  it would return true.

priyank-p: can you provide one example of such url?

app/renderer/js/utils/link-util.js
abhigyank: @cPhost  It is already a `LinkUtil` file, so that part is understoond I suppose. I wrote it similar to already existing `isInternal`.

priyank-p: Ah, yeah makes sense. `LinkUtil.isImage` sounds good (but  `LinkUtil.isImageURL` also does); You can ignore this then.

app/renderer/js/preload.js
priyank-p: Yeah, I think jQuery would make this very easy whereas js we would need query all `.message_content p a` and add listener. And event bubbling would also be hard since the target would be `a` and we need to then check if its inside `.message_content p` sounds too much for doing this in js side.

Otherwise, i would have said it should be on js side for consistency

akashnimare (approved):

akashnimare:
Merged, thanks @abhigyank.

timabbott:
@akashnimare @abhigyank can you explain in what situations this changes the behavior?  I just want to make sure I understand the new model.

akashnimare:
So if you click on an image attachement link it will open the image in the lightbox. Basically we are re-using webapp's `lightbox.open()` function and passing the image to have the same effect.

PR #473, opened 2018-04-09
sidebar: Add scrollbar for list of organizations on overflow.

abhigyank:
Adds scrollbar for list of organizations on overflow in the sidebar. Scrollbar visible on hover.
Issue discussed [here](https://chat.zulip.org/#narrow/stream/16-electron/topic/No.20scrollbar.20for.20Organisations.20within.20Zulip.20Desktop.20App).

**Screenshots?**
peek 2018-04-09 13-26

### DIFF
--- a/app/renderer/css/main.css
+++ b/app/renderer/css/main.css
@@ -45,6 +45,28 @@ body {
     transition: all 0.6s ease-out;
 }
 
+#view-controls-container {
+	height: calc(100% - 156px);
+	overflow-y: hidden;
+}
+
+#view-controls-container:hover {
+	overflow-y: overlay;
+}
+
+#view-controls-container::-webkit-scrollbar {
+	width: 4px;
+}
+
+#view-controls-container::-webkit-scrollbar-track {
+    -webkit-box-shadow: inset 0 0 6px rgba(0,0,0,0.3);
+}
+
+#view-controls-container::-webkit-scrollbar-thumb {
+  background-color: darkgrey;
+  outline: 1px solid slategrey;
+}
+
 @font-face {
     font-family: 'Material Icons';
     font-style: normal;
--- a/app/renderer/css/main.css
+++ b/app/renderer/css/main.css
@@ -59,7 +59,7 @@ body {
 }
 
 #view-controls-container::-webkit-scrollbar-track {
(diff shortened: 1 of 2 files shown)


akashnimare:
You also need to handle the position of the tooltip.

image

priyank-p:
I think we would also want the `+` add org button to be outside of scroll container other a user will need to scroll all the way down to add another org (Which we can avoid easily).

akashnimare:
Yeah, we could but I don't think we can put it anywhere else (it won't look good). Also, the use case for this button is pretty low so it won't hurt even if it's not visible.

timabbott:
Yeah, I think the button being at the bottom of the scrollable list is correct.

app/renderer/css/main.css
```diff
@@ -45,6 +45,28 @@ body {
...
+#view-controls-container:hover {
+	overflow-y: overlay;
+}
+
+#view-controls-container::-webkit-scrollbar {
+	width: 4px;
+}
+
+#view-controls-container::-webkit-scrollbar-track {
+    -webkit-box-shadow: inset 0 0 6px rgba(0,0,0,0.3);
```
priyank-p: This should be `box-shadow` :)

app/renderer/js/main.js
```diff
@@ -190,6 +190,8 @@ class ServerManagerView {
 	sidebarHoverEvent(SidebarButton, SidebarTooltip) {
 		SidebarButton.addEventListener('mouseover', () => {
 			SidebarTooltip.removeAttribute('style');
+			const top = SidebarButton.getBoundingClientRect().top;
```
priyank-p: `const { top } = SidebarButton.getBoundingClientRect()` here.

app/renderer/js/main.js
```diff
@@ -199,6 +201,8 @@ class ServerManagerView {
 	onHover(index, serverName) {
 		this.$serverIconTooltip[index].innerHTML = serverName;
 		this.$serverIconTooltip[index].removeAttribute('style');
+		const top = this.$serverIconTooltip[index].parentElement.getBoundingClientRect().top;
```
priyank-p: Ditto.

PR #480, opened 2018-04-28
proxy-setting: Feature to use system proxy settings.

### LINKED ISSUE #296: Add setting option to use system proxy settings

abhigyank:
This commit uses resolveProxy to to read system proxy settings by requesting to a url namely `google.com` using the 4 protocols `http`, `https`, `ftp`, `socks`, this causes the main process to return the system used proxy for this particular protocol. We store this proxies in proper proxy format string using ConfigUtil into settings.json as `systemProxySettings`.
On turning on the use system proxy settings this value of `systemProxySettings`is written into the proxy rules and the webview session and is reloaded as done previously with only manual settings.
It removes the previous use proxy option and replaces it with use system proxy and manual proxy
options.

Fixes: #296
  - [x] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -13,6 +13,7 @@ const { app, ipcMain } = electron;
 
 const BadgeSettings = require('./../renderer/js/pages/preference/badge-settings.js');
 const ConfigUtil = require('./../renderer/js/utils/config-util.js');
+const ProxyUtil = require('./../renderer/js/utils/proxy-util.js');
 
 // Adds debug features like hotkeys for triggering dev tools and reload
 // in development mode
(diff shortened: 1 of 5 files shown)


app/main/index.js
```diff
@@ -151,6 +151,54 @@ app.on('ready', () => {
 
 	const page = mainWindow.webContents;
 
+	const ses = page.session;
+
+	const resolveProxyUrl = 'chat.zulip.org';
```
akashnimare: @abhigyank is this hardcorded?

abhigyank: The main aim is make a request to a website using the mainWindow that can access the system proxy. This `resolveProxyUrl` can be anything, just about any website. Here I have taken `chat.zulip.org` but it could be `google.com` as well.

app/main/index.js
```diff
@@ -151,6 +151,54 @@ app.on('ready', () => {
 
 	const page = mainWindow.webContents;
 
+	const ses = page.session;
+
```
akashnimare: I think we could put this into a `proxy-util.js`.

abhigyank: Do you mean the entire thing from here till the [end](https://github.com/zulip/zulip-electron/pull/480/files#diff-b7ea71851f2bf6f64dbcec17f174a921R201)? 
I will try to do that. We can use it as `proxyUtil.setSystemProxy()` ?

abhigyank: I  think `storeSystemProxy`/`resolveSystemProxy` sounds appropirate.

abhigyank:
@akashnimare Could you also test this on your macOS? If you're not on a proxy server, you can just try putting some random proxy server in system settings and just test if the correct ones are being written on to the `settings.json`.

akashnimare:
Merged. Thanks, @abhigyank. The only thing remaining here is to update the system proxy settings without restarting the app. The more advance set up would be update the settings etc when user changes/updates these settings from the system configs. Let's also look into `http-proxy` if it gives better results :)

PR #485, opened 2018-05-03
travis: Upload build artifacts of PR's.

priyank-p:


priyank-p:
@akashnimare If we still want this I think let's first move to circleci. For background on this PR, we had problem with correctly uploading the artifacts (file names were the issue I think).

Issue #482, opened 2018-05-01
Fix broken loading indicator


akashnimare:
We need to remove the loading indicator once the server is loaded.

<img src="https://chat.zulip.org/user_uploads/2/2d/aVDSO0-ok9_5HJQ9NPcYTd7D/pasted_image.png"/>

abhigyank:
@zulipbot claim

PR #486, opened 2018-05-04
loading-indicator: Fix to hide loading indicator when server is loaded

abhigyank:
Moved the loading indicator to `webviews-container` as a pseudo element to it. The indicator is hidden once the active tabs `DOM` is loaded. In case of loading of functional tab and adding/removing servers(`DestroyView`) is explicitly handled in `main.js`.

Fixes: #482.

### DIFF
--- a/app/renderer/css/main.css
+++ b/app/renderer/css/main.css
@@ -13,9 +13,6 @@ body {
 #content {
     display: flex;
     height: 100%;
-    background: #fff url(../img/ic_loading.gif) no-repeat;
-    background-size: 60px 60px;
-    background-position: center;
 }
 
 .toggle-sidebar {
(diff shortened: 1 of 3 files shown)


app/renderer/css/main.css
```diff
@@ -273,6 +270,25 @@ body {
     width: 100%;
 }
 
+/*Pseudo element for loading indicator*/
+#webviews-container:before {
```
priyank-p: Can you use `::before` here?

`::` is for indicating its a pseudo element whereas `:` is for pseudo-classes, this was something new is CSS3, browsers support the old `:` before for backward compatibility reasons.

app/renderer/css/main.css
```diff
@@ -273,6 +270,25 @@ body {
...
+    z-index: 1;
+    background: #fff url(../img/ic_loading.gif) no-repeat;
+    background-size: 60px 60px;
+    background-position: center;
+    width: 100%;
+    height: 100%;
+}
+
+/*When the active webview is loaded*/
+#webviews-container.loaded:before {
```
priyank-p: ditto.

priyank-p (commented):
Just left a nitpicking comment.

Issue #126, opened 2017-03-08
Add validation for custom/self-signed certificates


penCsharpener:
Zulip 1.5.1 running on Ubuntu 16.04.2 in hyper-V 
the ubuntu machine is called zulip with a self-signed certificate
the settings.py has as external_host = 'zulip'
we only use it internally without access from the outside of the LAN
in the browser https://zulip works or https://localipaddress
in the electron app however this doesn't work (nor with the IP)
![image](https://cloud.githubusercontent.com/assets/26190934/23706530/dab8aa08-040e-11e7-80af-cca062c7c6b4.png)

Lplenka:
@vbNETonIce  Can you give a screenshot of opening this "https://zulip" server in your browser , i wanted to know if this server you want to access is accessible in your browser or not. https in localhost is quite unlikely

vbNETonIce:
yes, it is accessible. I'm on a Win10 machine and use zulip that runs on the hyper-v VM with Ubuntu. It works in all browsers and also in the legacy app but not in the electron app.
![2017-03-08_15-12-15_chrome](https://cloud.githubusercontent.com/assets/26190934/23707218/d78911b2-0411-11e7-9b43-b46c0750bc1a.png)

Lplenka:
First thing-
Make sure your server is running locally when you are tyring to access it in elctron (i.e make sure it is running in your browser at the same time when you are accessing it in electron).

Second-
Please check if this file "zulip.ogg" exists in your server root directory in the following path
root_directory_of_your_server/static/audio/zulip.ogg

vbNETonIce:
sorry. I'm not sure where to look for the file
/root/static doesn't exist
/home/zulip/deployment/current doesn't show a path static either

Lplenka:
@vbNETonIce  open your zulip(server) folder in ur VM  and press `Ctl+ F` , in the search bar type `zulip.ogg` and tell the path in which it exists.

It must exist somewhere in that server folder , I am pretty sure

vbNETonIce:
Ctrl + F doesn't seem to work (I tried Kitty and VM Connection of Hyper-V)

vbNETonIce:
ah, locate did the trick 
deployment/date and time/prod-static/serve/audio/zulip.ogg

Lplenka:
Great! @vbNETonIce 
It will be easy to fix this
[...]

PR #494, opened 2018-05-20
settings: Add option to add custom certificates

abhigyank:
To validate custom/self-signed certificates for servers by saving the certificate location and then passing the certificate in the request used to validate the server when adding. The validation is done
by the [request](https://www.npmjs.com/package/request#tlsssl-protocol) module itself.

Fixes: #126 

This PR now stores the location of the cert when the user adds it, a better option could be if we copy the certificate to the `appData` folder of the user in a certificates subfolder. This would prevent errors due to removal of the original file and seems something more like a behavior that a user would expect. Thoughts?

### DIFF
--- a/app/renderer/js/pages/preference/add-certificate.js
+++ b/app/renderer/js/pages/preference/add-certificate.js
@@ -1,0 +1,92 @@
+'use-strict';
+
+const { dialog } = require('electron').remote;
+
+const BaseComponent = require(__dirname + '/../../components/base.js');
+const CertificateUtil = require(__dirname + '/../../utils/certificate-util.js');
+const DomainUtil = require(__dirname + '/../../utils/domain-util.js');
+
+class AddCertificate extends BaseComponent {
+	constructor(props) {
+		super();
+		this.props = props;
+		this._certFile = '';
+	}
+
+	template() {
+		return `			
+			<div class="settings-card server-center certificates-card">
+				<div class="certificate-input">
+					<div>Organization URL :</div> 
+					<input class="setting-input-value" autofocus placeholder="your-organization.zulipchat.com or zulip.your-organization.com"/>
+				</div>
+				<div class="certificate-input">
+					<div>Custom CA's certificate file :</div> 
+					<button id="add-certificate-button">Add</button>
+				</div>
+			</div>
+		`;
+	}
+
+	init() {
+		this.$addCertificate = this.generateNodeFromTemplate(this.template());
+		this.props.$root.appendChild(this.$addCertificate);
+		this.addCertificateButton = this.$addCertificate.querySelector('#add-certificate-button');
+		this.serverUrl = this.$addCertificate.querySelectorAll('input.setting-input-value')[0];
+		this.initListeners();
+	}
+
+	validateAndAdd() {
+		const certificate = this._certFile;
+		const serverUrl = this.serverUrl.value;
+		if (certificate !== '' && serverUrl !== '') {
+			const server = encodeURIComponent(DomainUtil.formatUrl(serverUrl));
+			const fileName = certificate.substring(certificate.lastIndexOf('/') + 1);
+			const copy = CertificateUtil.copyCertificate(server, certificate, fileName);
+			if (!copy) {
+				console.log('We encountered error while saving the certificate.');
+				return;
+			}
+			CertificateUtil.setCertificate(server, fileName);
+			dialog.showMessageBox({
+				title: 'Success',
+				message: `Certificate saved!`
+			});
+			this.serverUrl.value = '';
+		} else {
+			dialog.showErrorBox('Error', `Please, ${serverUrl === '' ?
+      'Enter an Organization URL' : 'Choose certificate file'}`);
+		}
+	}
+
+	addHandler() {
+		const showDialogOptions = {
+			title: 'Select file',
+			defaultId: 1,
+			properties: ['openFile'],
+			filters: [{ name: 'crt, pem', extensions: ['crt', 'pem'] }]
+		};
+		dialog.showOpenDialog(showDialogOptions, selectedFile => {
+			if (selectedFile) {
+				this._certFile = selectedFile[0] || '';
+				this.validateAndAdd();
+			}
+		});
+	}
+
+	initListeners() {
+		this.addCertificateButton.addEventListener('click', () => {
+			this.addHandler();
+		});
+
+		this.serverUrl.addEventListener('keypress', event => {
+			const EnterkeyCode = event.keyCode;
+
+			if (EnterkeyCode === 13) {
+				this.addHandler();
+			}
+		});
+	}
+}
+
+module.exports = AddCertificate;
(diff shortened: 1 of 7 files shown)


timabbott:
Does it make sense to include some automated tests for the error handling logic (maybe using badssl.com) so that it's easier to tweak this stuff in the future?

abhigyank:
It seems a good idea. I think I can probably add that test when I would be working on the unit tests.

abhigyank:
There is an issue here, i.e. unable to copy certs if the `certificate` folder isn't already existing in appData folder. I think this would solved by making that folder already exisiting in the build? @akashnimare

akashnimare:
Yeah, we can create the folder on app start. This might go in `initDefaultSettings` or similar.

app/renderer/js/pages/preference/add-certificate.js
```diff
@@ -0,0 +1,93 @@
...
+class AddCertificate extends BaseComponent {
+	constructor(props) {
+		super();
+		this.props = props;
+		this._certFile = '';
+	}
+
+	template() {
+		return `			
+			<div class="settings-card server-center" style="width:70%;">
```
akashnimare: Can you put this CSS in the `preference.css`?

app/renderer/js/utils/certificate-util.js
```diff
@@ -0,0 +1,91 @@
...
+	setCertificate(server, fileName) {
+		const filePath = this.generateFilePath(fileName);
+		this.db.push(`/${server}`, filePath, true);
+		this.reloadDB();
+	}
+	removeCertificate(server) {
+		this.db.delete(`/${server}`);
+		this.reloadDB();
+	}
+	generateFilePath(filename) {
```
akashnimare: I think we can put this in the `default-util.js`?

app/renderer/js/utils/certificate-util.js
```diff
@@ -0,0 +1,91 @@
...
+		const dir = `${app.getPath('userData')}/certificates`;
+
+		// Create 'certificates' directory if not existed
+		if (!fs.existsSync(dir)) {
+			fs.mkdirSync(dir);
+		}
+
+		return `${dir}/${filename}`;
+	}
+	reloadDB() {
```
akashnimare: This exactly looks like the one which we have in the config-util.

akashnimare (changes requested):

app/renderer/js/utils/certificate-util.js
abhigyank: Um yes I think so. It's probably used in some other util too. Maybe move it from all. In a different PR?

akashnimare: Yeah, make sense. We should put this in one file only.

app/renderer/js/utils/certificate-util.js
abhigyank: Yes, it's the same template that I've followed. Modified it for `certifcates.json`. Since we need to reload it after every access. Right?

app/renderer/js/utils/certificate-util.js
abhigyank: Cool. I'll make a separate PR for that. Or do you think, we should just add the part of creating `certificates` folder in `default-util.js` just like Logs?

app/renderer/js/utils/certificate-util.js
akashnimare: I think it's okay for now but we should find a better way to reload the db. We can make a function which basically takes a db file as an argument and then we can reuse and pass it in other utils also.

abhigyank: Yes that is good idea. Maybe add it in something like `db-util.js`  later.

app/renderer/js/pages/preference/add-certificate.js
```diff
@@ -0,0 +1,92 @@
...
+
+	template() {
+		return `			
+			<div class="settings-card server-center certificates-card">
+				<div class="certificate-input">
+					<div>Organization URL :</div> 
+					<input class="setting-input-value" autofocus placeholder="your-organization.zulipchat.com or zulip.your-organization.com"/>
+				</div>
+				<div class="certificate-input">
+					<div >Custom CA's certificate file :</div>
```
akashnimare: Typo -
`<div > === <div>`

app/renderer/js/utils/domain-util.js
```diff
@@ -106,7 +108,19 @@ class DomainUtil {
...
+		if (certificate) {
+			// To handle case where certificate has been moved from the location in certificates.json
+			try {
+				certificateLocation = fs.readFileSync(certificate);
+			} catch (err) {
+				console.log(err);
+			}
+		}
+
+		const checkDomain = (certificateLocation) ? ({url: domain + '/static/audio/zulip.ogg', ca: certificateLocation}) : domain + '/static/audio/zulip.ogg';
```
akashnimare: Can you add a comment just above this, explaining a bit about what's going on?

abhigyank: I'll add one in the next commit push.

app/renderer/js/pages/preference/add-certificate.js
```diff
@@ -0,0 +1,92 @@
...
+		};
+		dialog.showOpenDialog(showDialogOptions, selectedFile => {
+			if (selectedFile) {
+				this._certFile = selectedFile[0] || '';
+				this.validateAndAdd();
+			}
+		});
+	}
+
+	initListeners() {
```
akashnimare, line 77: We are using the same function in other places also, remember to put this in shared utils so that we can reuse this in other files also.

akashnimare (approved):

app/renderer/js/utils/certificate-util.js
```diff
@@ -0,0 +1,85 @@
...
+	}
+	// Function to copy the certificate to userData folder
+	copyCertificate(server, location, fileName) {
+		let copied = false;
+		const filePath = `${certificatesDir}/${fileName}`;
+		try {
+			fs.copyFileSync(location, filePath);
+			copied = true;
+		} catch (err) {
+			console.log(err);
```
akashnimare: We should replace or maybe add our logger here for better debugging.

akashnimare:
@abhigyank great work 👏  Finally merged 🚀 

This is a great feature to have. I'm sure this will definitely help a lot of users/orgs who are using Zulip with a custom/self-signed certificates etc. The next thing we should do now is to document this feature properly in our desktop installation guide - https://chat.zulip.org/help/desktop-app-install-guide

I'll be releasing a beta shortly just to cross test this feature. @penCsharpener, @epaul13, @Shayan-To can you folks test this?

epaul13:
@akashnimare, @abhigyank : If you give me specific steps I can do some test. For now, using https://chat.zulip.org/help/desktop-app-install-guide, the ubuntu/apt part, Zulip install itself correctly, I can connect to my local self-signed Zulip instance, while getting the "Do you trust ..." dialog. Then I found the Add custom certificate and by adding my server cert chain, disconnecting and reconnecting my organization, I had no problems and no dialog. 

If you tell me where certs are stored I can delete and add cert before connecting :)

Better, if I can script the installation of the certs with the installation of zulip on my users desktops, everything will be perfect :)

Anyway thanks for the work. !

PR #495, opened 2018-05-20
Update node modules.

priyank-p:
This fixes all the vulnerabilities of packages, as pointed out by `npm audit`. All of them are for devDependencies so the app is not affected.

The only module that updated is `gulp`, other are updated using `--depth` option so only `package-lock.json` is updated.


akashnimare:
Merged, thanks @priyankp10.

PR #496, opened 2018-05-21
add-server: Make adding server dynamic [WIP].

### LINKED ISSUE #488: Do not reload app while adding/removing a server
I think it would be great if we could implement this behaviour. It's bit tricky since we have to reload the db file, activate the last server etc but let's give it a try.
cc @priyankp10, @abhigyank.

akashnimare:
Fixes: #488.

This refactors the code for adding and removing the server so that the app doesn't reload when adding/removing a server.

### DIFF
--- a/app/renderer/js/main.js
+++ b/app/renderer/js/main.js
@@ -120,8 +120,7 @@ class ServerManagerView {
 		this.toggleSidebar(showSidebar);
 	}
 
-	initTabs() {
-		const servers = DomainUtil.getDomains();
+	initTabs(servers = DomainUtil.getDomains()) {
 		if (servers.length > 0) {
 			for (let i = 0; i < servers.length; i++) {
 				this.initServer(servers[i], i);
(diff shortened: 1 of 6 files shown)


priyank-p:
This is getting really close, to be done..., The issue needs to be addressed, is that when adding a server, previous we add them beforehand so we can assume that `this.tabs` are all sorted and does not have a `FunctionalTab` now if we add it dynamically there will be if not more than one functional tab for adding an org so we need to filter them out and update the list accordingly which I need to think what we can do.

app/renderer/js/main.js
```diff
@@ -572,6 +600,7 @@ window.onload = () => {
 	serverManagerView.init();
 
   window.serverManagerView = serverManagerView;
+  window.FunctionalTab = FunctionalTab;
```
priyank-p: These here, are just for testing, provides a great way see what happens underneath the hood using dev-tools.

Just leaving a comment so I know to remove them when cleaning up.

akashnimare:
@priyankp10 let's spend some time to finish this.

PR #499, opened 2018-05-25
documentation: Change #electron stream links to #desktop stream link.

abhigyank:
Updating the stream link and name from #electron to #desktop after the change in stream name as discussed [here](https://chat.zulip.org/#narrow/stream/16-desktop/topic/Stream.20name) .

I don't think there are any more mentions of it in this repo. If there is any I missed out please leave a comment I will update the PR.


akashnimare:
Thanks, @abhigyank.

PR #501, opened 2018-06-08
config-util: Add isConfigItemExists to check if a key exists.

abhigyank:
Add isConfigItemExists to check if a key exists in `ConfigUtil`.

### DIFF
--- a/app/renderer/js/utils/config-util.js
+++ b/app/renderer/js/utils/config-util.js
@@ -49,6 +49,16 @@ class ConfigUtil {
 		}
 	}
 
+	isConfigItemExists(key) {
+		this.reloadDB();
+		const value = this.db.getData('/')[key];
+		if (value === undefined) {
+			return false;
+		} else {
+			return true;
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


akashnimare:
Merged. Thanks, @abhigyank.

app/renderer/js/utils/config-util.js
```diff
@@ -49,6 +49,16 @@ class ConfigUtil {
...
+	isConfigItemExists(key) {
+		this.reloadDB();
+		const value = this.db.getData('/')[key];
+		if (value === undefined) {
+			return false;
+		} else {
+			return true;
+		}
+	}
+
```
akashnimare, line 61: This can be also be done like -
```js
	isConfigItemExists(key) {
		this.reloadDB();
		const value = this.db.getData('/')[key];
		return (value === undefined);
	};
```

abhigyank: Shouldn't it be 
```
return !(value === undefined);
```

PR #505, opened 2018-06-23
logger: Replace all instances of console.log with our logger-util.

abhigyank:
In this PR all instances of console.log in `app` are replaced with our logger-util. It writes all errors to some log file. `errors.log` file is used to write errors which are are caused in files like main,js, helpers.js etc.

### DIFF
--- a/app/renderer/js/spellchecker.js
+++ b/app/renderer/js/spellchecker.js
@@ -3,6 +3,12 @@
 const { SpellCheckHandler, ContextMenuListener, ContextMenuBuilder } = require('electron-spellchecker');
 
 const ConfigUtil = require(__dirname + '/utils/config-util.js');
+const Logger = require(__dirname + '/js/utils/logger-util.js');
+
+const logger = new Logger({
+	file: 'errors.log',
+	timestamp: true
+});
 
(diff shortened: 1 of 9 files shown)


app/renderer/js/spellchecker.js
```diff
@@ -3,6 +3,12 @@
 const { SpellCheckHandler, ContextMenuListener, ContextMenuBuilder } = require('electron-spellchecker');
 
 const ConfigUtil = require(__dirname + '/utils/config-util.js');
+const Logger = require(__dirname + '/js/utils/logger-util.js');
```
akashnimare: This path is wrong, we don't need to add `js` -
```
const Logger = require(__dirname + '/utils/logger-util.js');
```

akashnimare (changes requested):

app/renderer/js/spellchecker.js
abhigyank: Aah copy paste woes! I'll fix that.

akashnimare (approved):

akashnimare:
Merged, thanks @abhigyank.

PR #507, opened 2018-06-26
[WIP] config-util: Use package reload instead of custom reload.

abhigyank:
This is a WIP commit to fix the errors while saving settings that occur from reloadDB function.
Here I have changed using `reloadDB` to use the reload function offered by the `node-json-db` package itself. 
Also, have removed the reloading db where we are writing to the database and replaced it with saving the db. Since we are using `saveOnPush` already, this part is actually redundant and not necessary, but for safety purposes I have put in now.
Regarding reloading the db,I don't think it is necessary at all, because `config-util` is the only file where changes to `settings-json` is being done, meaning no exterior change to the database file, so the database variable in it would be up to date and no conflict would occur. Removing reload would reduce the chances of causing db errors significantly.

### DIFF
--- a/app/renderer/js/utils/config-util.js
+++ b/app/renderer/js/utils/config-util.js
@@ -39,7 +39,12 @@ class ConfigUtil {
 	}
 
 	getConfigItem(key, defaultValue = null) {
-		this.reloadDB();
+		try {
+			this.db.reload();
+		} catch (err) {
+			logger.error('Error while reloading settings.json: ');
+			logger.error(err);
+		}
 		const value = this.db.getData('/')[key];
 		if (value === undefined) {
 			this.setConfigItem(key, defaultValue);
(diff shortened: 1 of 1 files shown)


app/renderer/js/utils/config-util.js
```diff
@@ -39,7 +39,12 @@ class ConfigUtil {
 	}
 
 	getConfigItem(key, defaultValue = null) {
-		this.reloadDB();
+		try {
+			this.db.reload();
```
abhigyank, line 43: This can possibly be removed.

app/renderer/js/utils/config-util.js
```diff
@@ -50,19 +55,24 @@ class ConfigUtil {
 	}
 	// This function returns whether a key exists in the configuration file (settings.json)
 	isConfigItemExists(key) {
-		this.reloadDB();
+		try {
+			this.db.reload();
```
abhigyank, line 59: Same here.

akashnimare:
Can you try removing `this.db.reload()` in `getConfigItem` and `isConfigItemExists` and see if it makes any difference? I think we might not need it.

abhigyank:
Yes that's why I feel as well. Do you want me to remove that and push?

akashnimare:
Yeah, make the changes and see if everything works as expected and then push those changes.

akashnimare:
Merged into master. Thanks, @abhigyank 🚀

PR #508, opened 2018-06-29
config-files: Move all the configuration files into config folder.

abhigyank:
This PR moves all the configuration(.json) files into a config folder if the config folder doesn't already exists. It also updates the places where these files are being accessed with the new address.
The files are `updates.json`, `certificates.json`, `domain.json`, `settings.json`, `window-state.json`. `window-state.json` is moved by changing the path in the package `electron-window-state`.

### DIFF
--- a/app/main/menu.js
+++ b/app/main/menu.js
@@ -423,7 +423,7 @@ class AppMenu {
 		const resetAppSettingsMessage = 'By proceeding you will be removing all connected organizations and preferences from Zulip.';
 
 		// We save App's settings/configurations in following files
-		const settingFiles = ['window-state.json', 'domain.json', 'settings.json', 'certificates.json'];
+		const settingFiles = ['window-state.json', 'config/domain.json', 'config/settings.json', 'config/certificates.json'];
 
 		dialog.showMessageBox({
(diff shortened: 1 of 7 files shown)


app/main/menu.js
```diff
@@ -423,7 +423,7 @@ class AppMenu {
 		const resetAppSettingsMessage = 'By proceeding you will be removing all connected organizations and preferences from Zulip.';
 
 		// We save App's settings/configurations in following files
-		const settingFiles = ['window-state.json', 'domain.json', 'settings.json', 'certificates.json'];
+		const settingFiles = ['window-state.json', 'config/domain.json', 'config/settings.json', 'config/certificates.json'];
```
akashnimare: @abhigyank I think we should move this `window-state.json` file to `/config`.
https://github.com/mawie81/electron-window-state

abhigyank: `window-state.json` is created by this module. I don't think we can change where this module creates the file. If we move it to `/config`, it will recreate the file in the appData folder.

akashnimare:
Thanks, @abhigyank!

PR #509, opened 2018-07-04
error-logs: Zip files into Downloads folder when user clicks on Show App Logs

abhigyank:
This PR adds function that when a user clicks of "Show App Logs" from the menu, open the Downloads folder with a zip file containing all the Logs and User Configs. This will help the user to easily share logs with us.


akashnimare:
Thanks, @abhigyank. Let's work on implementing the sentry now.

PR #511, opened 2018-07-06
package.json: Move adm dependecy to app/package.json

abhigyank:
This fixes wrong placement of `adm` dependency (in #509) in `package.json` to `app/package.json` .


akashnimare:
Don't know how we missed this 😅

PR #512, opened 2018-07-06
Setup sentry for the repository

abhigyank:


app/renderer/js/utils/domain-util.js
```diff
@@ -155,6 +155,9 @@ class DomainUtil {
 							resolve(serverConf);
 						});
 					} else {
+						// Report error to sentry to get idea of posible certificate errors
+						// users get when adding the servers
+						logger.reportSentry(new Error(error));
```
akashnimare: @abhigyank will `logger.reportSentry(error);` work?

abhigyank: No not here, because error has been changed to a string. We can send the error then as a string with `captureMessage` of sentry, but it'd be more of a text string rather then having the error trace stack.

akashnimare: Ah, cool.

akashnimare (approved):

akashnimare:
Merged, thanks, @abhigyank!

PR #513, opened 2018-07-09
macOS: Add dock bouncing effect on macOS.

akashnimare:
---

**What's this PR do?**
Add a new setting option to control the dock boucing on macOS. 

**Screenshots?**

  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [X] macOS

### DIFF
--- a/app/renderer/js/pages/preference/general-section.js
+++ b/app/renderer/js/pages/preference/general-section.js
@@ -31,6 +31,10 @@ class GeneralSection extends BaseSection {
 						<div class="setting-description">Show app unread badge</div>
 						<div class="setting-control"></div>
 					</div>
+					<div class="setting-row" id="dock-bounce-option">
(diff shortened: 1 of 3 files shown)


app/renderer/js/pages/preference/general-section.js
```diff
@@ -31,6 +31,10 @@ class GeneralSection extends BaseSection {
 						<div class="setting-description">Show app unread badge</div>
 						<div class="setting-control"></div>
 					</div>
+					<div class="setting-row" id="dock-bounce-option">
```
abhigyank: Probably add style like ` style= "display:${process.platform === 'macOS' ? '' : 'none'}"` ? This'd show this option only in mac.

akashnimare: Oh yes!

PR #514, opened 2018-07-10
Download file attachments to a default location.

abhigyank:
  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS


akashnimare:
Merged, thanks @abhigyank. We should document this in the web app/usage guide somewhere.

PR #515, opened 2018-07-12
certificates: Make default option "No" on enter key press to reject certificate.

abhigyank:
This is done so that if a user just presses `enter` without reading the text, by default it'd  reject the text.


akashnimare:
Nice, merged. @gnprice thanks for the suggestion \m/

PR #516, opened 2018-07-13
electron-bridge: Implement electron bridge

abhigyank:
This PR completes the minor work left from #428 .

### DIFF
--- a/app/renderer/js/electron-bridge.js
+++ b/app/renderer/js/electron-bridge.js
@@ -1,0 +1,39 @@
+const events = require('events');
+const { ipcRenderer } = require('electron');
+
+// we have and will have some non camelcase stuff
+// while working with zulip so just turning the rule off
+// for the whole file.
+/* eslint-disable camelcase */
+class ElectronBridge extends events {
+	send_event(...args) {
+		this.emit(...args);
+	}
+
+	on_event(...args) {
+		this.on(...args);
+	}
+}
+
+const electron_bridge = new ElectronBridge();
+
+electron_bridge.on('total_unread_count', (...args) => {
+	ipcRenderer.send('unread-count', ...args);
+});
+
+electron_bridge.on('realm_name', (...args) => {
+	ipcRenderer.send('realm-name-changed', ...args);
+});
+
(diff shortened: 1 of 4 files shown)


priyank-p:
@abhigyank Thanks for finishing it off! LGTM.

app/renderer/js/electron-bridge.js
```diff
@@ -0,0 +1,44 @@
+const events = require('events');
+const { ipcRenderer } = require('electron');
+
+// we have and will have some non camelcase stuff
+// while working with zulip so just turning the rule off
+// for the whole file.
+/* eslint-disable camelcase */
+class ElectronBridge extends events {
+	// eslint-disable-next-line no-useless-constructor
+	constructor() {
```
priyank-p: Minor Thing, can you remove this thing, this is not required anymore.

priyank-p (approved):

app/renderer/js/electron-bridge.js
abhigyank: Thanks, done!

akashnimare:
Merged, thanks, @priyankp10, @abhigyank. We now have a working bridge between the webapp and the desktop app :construction:

timabbott:
So, this handles the icon changing immediately if the app was running when the admin changes it.

If the realm icon is changed while the app isn't running, do we handle that already?

akashnimare:
> If the realm icon is changed while the app isn't running, do we handle that already?

Yeah, we do have a [function](https://github.com/zulip/zulip-electron/blob/781883a898ed30f2ddc95928cbacb945230ab5ed/app/renderer/js/utils/domain-util.js#L248) which lazy loads the realm icon but this needs a bit of work. Current flow is -
* Someone updated the realm icon while app isn't running
* User start the app they'll see the old icon
* On reload/re-starting the app again they'll see the new icon

timabbott:
Cool.  If it were helpful for this sense, we could trigger an electron_bridge event on webapp startup; we'd just want to make sure the app is avoiding doing unnecessary work on every load if the icon URL hasn't changed.

PR #522, opened 2018-07-20
Fixes: #517

### LINKED ISSUE #517: page_params is not defined
We are getting this error because on some pages the `page_params`is not there so we gotta add a check for the same.

aashish-ak:
---

**What's this PR do?**
  It fixes #517 ,

 The error 'page_params is not defined' occured because when the third party authorization windows 
 are opened at the time of login, so i added a check that if the variable `page_params` exist only then load the DOM.

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/notification/index.js
+++ b/app/renderer/js/notification/index.js
@@ -20,8 +20,15 @@ if (process.platform === 'darwin') {
 
 window.addEventListener('load', () => {
 	// Call this function only when user is logged in
+	let webpageParams = null;
+	try {
+		// eslint-disable-next-line no-undef, camelcase
+		webpageParams = page_params;
+	} catch (err) {
+		webpageParams = null;
+	}
 	// eslint-disable-next-line no-undef, camelcase
-	if (page_params.realm_uri) {
+	if (webpageParams && page_params.realm_uri) {
 		loadBots();
 	}
 });
--- a/app/renderer/js/notification/index.js
+++ b/app/renderer/js/notification/index.js
@@ -4,6 +4,7 @@ const {
   remote: { app }
 } = require('electron');
 
+const params = require('../utils/params-util.js');
 const DefaultNotification = require('./default-notification');
 const { appId, loadBots } = require('./helpers');
 
@@ -19,16 +20,8 @@ if (process.platform === 'darwin') {
 }
 
 window.addEventListener('load', () => {
-	// Call this function only when user is logged in
-	let webpageParams = null;
-	try {
-		// eslint-disable-next-line no-undef, camelcase
-		webpageParams = page_params;
-	} catch (err) {
-		webpageParams = null;
-	}
 	// eslint-disable-next-line no-undef, camelcase
-	if (webpageParams && page_params.realm_uri) {
+	if (params.isPageParams() && page_params.realm_uri) {
 		loadBots();
 	}
 });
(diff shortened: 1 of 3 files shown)


abhigyank:
Hello @aashish-ak thanks for doing this!
However I think the correct approach could be to use a `try..catch` for the `serverLanguage` variable. I don't think hard-coding the websites is a good solution.

aashish-ak:
Okay will it be better if i fix the hostname to `*.zulip.*` or `*.zulipchat.*` ?? Will it work?? @abhigyank

abhigyank:
No we can't do that, as many people host Zulip on-premise and they have different domains.
Can you try something like?
```js
let serverLanguage = null;
try{
    serverLanguage = page_params.default_language;
} 
```

aashish-ak:
@abhigyank this isn't helping as the the whole code in that block should be loaded if the window is not of a third party authorisation method.

aashish-ak:
@abhigyank I've modified the code, merge it if you find the PR worthy :smile:

app/renderer/js/notification/index.js
```diff
@@ -20,8 +20,14 @@ if (process.platform === 'darwin') {
...
 	// Call this function only when user is logged in
-	// eslint-disable-next-line no-undef, camelcase
-	if (page_params.realm_uri) {
+	let webpageParams = null;
+	try {
+		// eslint-disable-next-line no-undef, camelcase
+		webpageParams = page_params;
+	} catch (err) {
+		webpageParams = null;
+	}// eslint-disable-next-line no-undef, camelcase
```
abhigyank: Can you move the comment to the new line?

aashish-ak: @abhigyank  Done :+1:

abhigyank:
@aashish-ak Looks good to me, left a comment about a small change.
Also can you squash both the commits into a single commit. You can look up `git rebase -i` for the same [here](https://zulip.readthedocs.io/en/latest/git/fixing-commits.html?highlight=squash#squashing-commits).
Also the can you change the commit message as per our guidelines [here](https://zulip.readthedocs.io/en/latest/contributing/version-control.html?highlight=commit%20message#commit-messages) . You can also look at other commit messages for references.

abhigyank (approved):

abhigyank (approved):
@akashnimare LGTM. Can be merged.

akashnimare:
@aashish-ak, @abhigyank  since we are using the same code, I think we can wrap it up in a function and export it wherever we need, something like this -

```js
const isPageParams = () => {
    let webpageParams = null;
    try {
        webpageParams = page_params;
    } catch (err) {
        webpageParams = null;
    }
    return webpageParams;
}


module.exports = {
    isPageParams
}

```

abhigyank:
Right, this should probably go in `default-util` right?

akashnimare:
There or maybe we could create `params-util.js`.

akashnimare:
@aashish-ak are you interested in finishing this PR?

aashish-ak:
Sorry been a little busy, yes i'll finish it soon :smile:

aashish-ak:
I'll complete the work till noon, you can merge easily after that ☺️

akashnimare:
Awesome, would be great if you can finish it up since we're planning to add this in a new release.

aashish-ak:
@akashnimare @abhigyank It's done, sorry for the late commit.

akashnimare:
Merged after fixing linting errors and commit message. Thanks, @aashish-ak.

Issue #82, opened 2016-11-02
Build an efficient process for testing and releasing new versions of the desktop apps


brainwane:
This is a TODO from https://zulip.readthedocs.io/en/latest/roadmap.html that Tim included in the Zulip roadmap when he first published it in 27b8e8b294e3c43566dea724d059ef32af16a8f3 .

akashnimare:
@brainwane Sorry for the late reply as I was on a little vacation 😄 .
I have not done much work on testing but releasing new versions ie auto updates etc will be covered in our first stable release.

geeeeeeeeek:
What kind of tests are you looking for? Since our app is directly fetched from the web, there is only few things we can test on Electron, like testing if each menu item works.

timabbott:
Yeah I think the things I'd have automated tests for are:
* Areas we've had a lot of bugs, e.g. the handling of the server URL entered on the first screen would be a good candidate for this
* That the various menu items work
* Basic testing of any dependencies we've had trouble with.

And then maybe part 2 of this (which probably deserves its own issue) is that we should have a beta channel in the auto-update system, so that we can push an update to only some people, wait a few days for issue reports, and then push it out to the rest.

akashnimare:
> Basic testing of any dependencies we've had trouble with.

@timabbott specially **Spellchecker**.

geeeeeeeeek:
I'll try to write a plan. And see what we can do on testing.

priyank-p:
@timabbott i can start working on test using mocha if you like.

akashnimare:
For testing, have a look [test-spellchecker](https://github.com/zulip/zulip-electron/tree/test-spellchecker) branch (it's WIP). What kind of test you want to work on? @SimplyAhmazing have added couple of e2e tests [here](https://github.com/zulip/zulip-electron/pull/337).

priyank-p:
I am working with Ahmed Abdalla on that one 
edit: with @SimplyAhmazing.

SimplyAhmazing:
@cPhost There 3 categories of testing for an electron application - testing code in the renderer process, testing code in the main process, and integration tests that runs the entire app with both processes live.

* `renderer process` tests - I just submitted [this PR](https://github.com/zulip/zulip-electron/pull/350) for writing tests that are executed in the renderer process
[...]

PR #526, opened 2018-07-24
Setup e2e and karma unit tests

abhigyank:
This PR sets up karma for executing renderer process tests in electron and adds one test file for testing spellchecker, refactors tests dir into tests/e2e & tests/unit. 
It picks up parts of #350 and improves upon them.

Address #82


tests/package.json
```diff
@@ -0,0 +1,185 @@
+{
+ "name": "zulip",
+ "productName": "ZulipTest",
+ "version": "2.3.3",
+ "main": "../app/main",
+ "description": "Zulip Desktop App",
```
priyank-p: For making the electron app use the `ZulipTest` app temp directory and the `../app/main` entry point only `productName`, `name` and `version` are required, you can remove other stuff.

abhigyank: I think this is being dynamically created in the `generateTestAppPackageJson` function from the current `package.json` of the main app.

abhigyank: @priyankp10  I think the a good idea would be to remove this file, and create it at the time of test start and then remove it when the test is over. What do you think?

abhigyank:
@priyankp10  @akashnimare Can you guys give this a test run on macOS?

tests/package.json
priyank-p: Yep, That is what was done is the last PR I think?

akashnimare:
@abhigyank can you rebase this so that I can review?

abhigyank:
@akashnimare rebased.
use `npm run test-unit`.

akashnimare:
@abhigyank is this ready to review?

abhigyank:
I'll test it once and update you.

abhigyank:
I think currently this may not work due to failure in the spectron package as we had discussed in the retreat - electron/spectron#301

abhigyank:
@priyank-p updated the PR. 
`npm run test-e2e` works as expected, `npm run test-unit` works, but would prefer someone to test it on macOS. If it doesn't we can remove the spellchecker unit test and keep the framework as unit test framework is working correctly.

priyank-p:
@abhigyank can add this to `.travis.yml` so these tests run in CI.

karma.conf.js
```diff
@@ -0,0 +1,21 @@
+module.exports = function (config) {
+	config.set({
+		basepath: '',
+		frameworks: ['jasmine'],
```
priyank-p: Let use mocha, instead of jasmine. I think mocha is widely used. The only change needed to do this is changing `test` to `it` or `describe`.

abhigyank: What assertion library should you suggest when using `mocha`, should I go with `chai`?

priyank-p: I don't think we will need a assertion library. We can just user the node.js `assert` module to do the assertions.

abhigyank:
@priyank-p can do a test on this locally? Both test works for me on my local, but seems to fail on the travis-ci. I am unable to understand the occurrence of the error on travis-ci, can you give a look?

priyank-p:
Isn't the spectron tests failing in the CI? I think that is just because of travis' hardware being slow so the tests are timing out. We should just get rid of them by porting them to karma or disable them just like before in CI.

tests/unit/spellchecker.js
```diff
@@ -1,5 +1,6 @@
 const ConfigUtil = require('js/utils/config-util.js');
 const SetupSpellChecker = require('js/spellchecker')
+var assert = require('assert');
```
priyank-p: Minor nit: use `const`

tests/unit/spellchecker.js
```diff
@@ -8,15 +9,16 @@ if(process.platform === 'darwin') {
 	  ConfigUtil.setConfigItem('spellcheckerLanguage', 'en');
 	  var dictionary = ['apple', 'pear', 'banana', 'pie'];
 		// console.log(dictionary);  
+
 	  SetupSpellChecker.init(dictionary);  // re-initialize after setting update
 	  const spellCheckHandler = SetupSpellChecker.SpellCheckHandler
 
 	  it('mark misspelled word', function () {
-	    expect(spellCheckHandler.isMisspelled('asdasdasdw')).toBe(true)
+	    assert(spellCheckHandler.isMisspelled('asdasdasdw') == true)
```
priyank-p: This can just be `assert(spellCheckHandler.isMisspelled('asdasdasdw'))`

tests/unit/spellchecker.js
```diff
@@ -8,15 +9,16 @@ if(process.platform === 'darwin') {
...
 	  })
 
 	  it('verify properly spelled word', function () {
-	    expect(spellCheckHandler.isMisspelled('help')).toBe(false)
+	    assert(spellCheckHandler.isMisspelled('help') == false)
```
priyank-p: Use `===`

abhigyank:
Removed the e2e spectron test from travis (passes locally). 
unit-test passes on macOS, on linux no unit test is present so fails.So, for now running unit-tests on macOS only in travis.

abhigyank:
@akashnimare I think its ready to merge.
@priyank-p give both the tests a try .

PR #530, opened 2018-07-30
attachments: Fix failing attached file downloads.

### LINKED ISSUE #523: File download not working/no dialog shown
referring to https://github.com/zulip/zulip-electron/commit/6787b8d5573b1118851a4bb91e1e32fccb12dd2a

- **Operating System**:
  - [x] Windows
  - [ ] Linux/Ubutnu
  - [ ] macOS
Zulip hosted on private server. Click on file attachment. nothing happens, just a short sound is played. No notification shown and file is not being downloaded.

abhigyank:
This commit fixes the failing download files that had occured due to session not being same in the browserWindow and the webview. This made the uploaded files unavailable to browserWindow for download. 
This fix adds the persist session to the browserWindow, and should now work for all servers. It'd still revert back to the dialog download in case the download fails for some other reason.
Fixes: #523

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -73,7 +73,8 @@ function createMainWindow() {
 		minHeight: 400,
 		webPreferences: {
 			plugins: true,
-			nodeIntegration: true
+			nodeIntegration: true,
+			partition: 'persist:webviewsession'
 		},
 		show: false
 	});


akashnimare:
Thanks, @abhigyank.

Issue #534, opened 2018-08-03
Add proxy details in request module while validating server


akashnimare:
If a user has turned on the system/manual proxy settings, we need to send the proxy parameters while validating a server [here](https://github.com/zulip/zulip-electron/blob/781883a898ed30f2ddc95928cbacb945230ab5ed/app/renderer/js/utils/domain-util.js#L132).

PR #535, opened 2018-08-05
proxy: Set environment proxy variables as per proxyRules.

abhigyank:
Fixes: #534.

---

**What's this PR do?**
This commit sets the http and https proxy environment variable as per proxyRules so that the request module can use these rules while sending a request.

`request` module is used in many places in the codebase. It respects the environment proxy variables so usually it used to work as whenever proxy is set in the system using GUI the environment variables are usually written as usual. However, sometimes the environment proxy variables maybe incorrect or not written due to some OS issues. Hence we are are writing the environment variables temporarily for the duration of app run with the proxy rules that we are using for the webview.

Fixes: #534 

  - [x] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/utils/proxy-util.js
+++ b/app/renderer/js/utils/proxy-util.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const url = require('url');
 const ConfigUtil = require('./config-util.js');
 
 let instance = null;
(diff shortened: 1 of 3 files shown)


abhigyank:
@akashnimare  Added a commit that passes proxy as a parameter. If it is fine, i'll remove the first commit and update.

app/renderer/js/utils/proxy-util.js
```diff
@@ -15,6 +16,39 @@ class ProxyUtil {
...
+		const proxyRules = ConfigUtil.getConfigItem('proxyRules', '').split(';');
+		// If SPS is on and system uses no proxy then request should not try to use proxy from
+		// environment. NO_PROXY = '*' makes request ignore all environment proxy variables.
+		if (proxyRules[0] === '') {
+			process.env.NO_PROXY = '*';
+			return;
+		}
+
+		const proxyRule = {};
+		if (uri.protocol === 'http:') {
```
akashnimare, line 31: @abhigyank we can refactor this function since it's the same for http + https, but let's leave it for now.

akashnimare:
Merged. Thanks, @abhigyank.

Issue #547, opened 2018-08-27
Add an option to view pdf files in the app


akashnimare:
I think we should add a basic pdf viewer within the desktop app until we fix https://github.com/zulip/zulip/issues/9081.

CarloLucibello:
It seems this feature has been removed, see https://github.com/zulip/zulip-desktop/issues/640, can we reopen the issue?

CarloLucibello:
Related issue is #664

PR #549, opened 2018-08-29
pdf-viewer: Show pdf files in a new window.

akashnimare:
This adds a feature of showing the pdf attachments in a new window so that a user can quickly view the same.

Fixes: #547.

**Screenshots?**

  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [x] macOS

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -196,6 +196,28 @@ app.on('ready', () => {
 		app.quit();
 	});
 
+	// Show pdf in a new BrowserWindow
+	ipcMain.on('pdf-view', (event, url) => {
+		// Paddings for pdfWindow so that it fits into the main browserWindow
+		const paddingWidth = 55;
+		const paddingHeight = 22;
+
+		// Get the config of main browserWindow
+		const mainWindowState = global.mainWindowState;
+
+		// Window to view the pdf file
+		const pdfWindow = new electron.BrowserWindow({
+			x: mainWindowState.x + paddingWidth,
+			y: mainWindowState.y + paddingHeight,
+			width: mainWindowState.width - paddingWidth,
+			height: mainWindowState.height - paddingHeight,
+			webPreferences: {
+				plugins: true
+			}
+		});
+		pdfWindow.loadURL(url);
+	});
+
 	// Reload full app not just webview, useful in debugging
 	ipcMain.on('reload-full-app', () => {
 		mainWindow.reload();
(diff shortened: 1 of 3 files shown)


akashnimare:
@abhigyank can you cross-test this?

abhigyank (commented):
The PDF opens in a new window as expected in a new window for CZO.
For organizations on zulipchat - asks me to login into the realm. I think we need to pass the session to the pdfWindow.
Clicking on download - pops up the dialog box to download to a folder. This is not connected to our download notification system hence no notification on download completion.

app/main/index.js
```diff
@@ -196,6 +196,28 @@ app.on('ready', () => {
...
+		const mainWindowState = global.mainWindowState;
+
+		// Window to view the pdf file
+		const pdfWindow = new electron.BrowserWindow({
+			x: mainWindowState.x + paddingWidth,
+			y: mainWindowState.y + paddingHeight,
+			width: mainWindowState.width - paddingWidth,
+			height: mainWindowState.height - paddingHeight,
+			webPreferences: {
+				plugins: true
```
abhigyank: I think we need to pass `partition: 'persist:webviewsession'` here for realm session to pass to this BrowserWindow.

abhigyank: I think we need to pass the `partition: 'persist:webviewsession'` parameter to transfer the realm session into this BrowserWindow.

akashnimare: Oh yes, I forgot. Thanks!

abhigyank:
@akashnimare couple of things in this PR - 
* Clicking on download - pops up the dialog box to download to a folder. This is not connected to our download notification system hence no notification on download completion. Is there a way that we could capture this download and show a notification upon completion?
* In the pdf window - I don't think we should have the top menu bar (that includes options of File,View etc).

akashnimare:
> Is there a way that we could capture this download and show a notification upon completion?

I doubt it and it's okay to not have the notification option I think. I looked into it a bit but couldn't find any way to trigger the notification event when a user clicks on the download button.

> In the pdf window - I don't think we should have the top menu bar (that includes options of File,View etc).

There is no menu bar in the macOS. Not sure how to hide it in the Windows.

abhigyank:
There is a menu bar in Linux as well. I'll see if it is possible to hide it.

akashnimare:
> There is a menu bar in Linux as well. I'll see if it is possible to hide it.

Yeah, setting the menu to be `null` does the job here -
 > `pdfWindow.setMenu(null)`

Issue #425, opened 2018-02-20
Organization name does not update on the sidebar


anurag1018:
---

- **Operating System**:
  - [ ] Windows
  - [ ] Linux/Ubutnu
  - [x] macOS
![zulipelectronbug](https://user-images.githubusercontent.com/17769227/36432456-40a80dac-1680-11e8-9a6b-3e5320c0cc2e.gif)

Upon changing the Organization's name and icon, they are not instantly updated on the sidebar. I tried logging out and back in but still the bug persists. 
The only way to get the name and icon updated is to delete and re add the organization from the app Settings>Servers page.

akashnimare:
Yeah, it takes time to load. You gotta reload app to see the changes. We basically fetch and save the icon URL on the first load/hard reload etc.

timabbott:
@akashnimare if we had a hook mechanism for webapp code calling into the Electron app, it'd be easy to make this work.  I think we should think about designing something like this.  E.g. the old desktop app had a `window.bridge` defined in the app.  For the Electron app, the webapp code could do e.g.
```
if (window.electron_bridge !== undefined && window_electron_bridge.update_org_name !== undefined) {
    window_electron_bridge.update_org_name(new_name);
}
```

inside the handler for the webapp receiving an event update from the server that the organization name had changed.  (Or maybe a slightly different API, but we need to deal with cases where either the webapp or the Electron app is newer, so that sort of feature detection is probably a good idea)

(And the Electron app would just have an object it would add to the window, `window.electron_bridge`, which we would attach functions to for features like this.

I think this is a good simple mechanism that we could do for the first use of this, but once we have it working, we can switch a lot of other things (unread count data, etc.) to use it rather than ugly things like parsing the title bar.

akashnimare:
I agree that we should use this kind of approach to pass the data from webapp to electron app.

timabbott:
OK.  @akashnimare do you want to try doing the exact interface I described above, and do PRs to this repo and the zulip/zulip one for this feature?  I think it's worth getting this protocol worked out soon, so that we can use it for the next N features.

timabbott:
@cPhost do you want to work on this using the `electron_bridge` approach?

priyank-p:
I am working on it. I was looking for code on webapp that does this but did get chance to dig more into the codebase.

abhigyank:
@zulipbot  claim

PR #553, opened 2018-09-02
realm name: Update sidebar realm name when it's changed in webapp.

abhigyank:
This PR updates the realm name in the sidebar tooltip and domains.json when it is updated in the server through electron_bridge.

This PR  also removes the server.alias parameter from the tooltip `onHover` function and inserts it in the the innerHTML of the tooltip DOM, This helps to easily update the tooltip text when the realm name is changed. Since this text is never removed this is a good way to do this as we now simply leverage `display:none` to hide/show the tooltip.

Fixes: #425

### DIFF
--- a/app/renderer/js/main.js
+++ b/app/renderer/js/main.js
@@ -4,6 +4,7 @@ const { ipcRenderer, remote } = require('electron');
 const isDev = require('electron-is-dev');
 
 const { session, app } = remote;
+const escape = require('escape-html');
 
 require(__dirname + '/js/tray.js');
 const DomainUtil = require(__dirname + '/js/utils/domain-util.js');
@@ -168,11 +169,12 @@ class ServerManagerView {
 		this.tabs.push(new ServerTab({
 			role: 'server',
 			icon: server.icon,
+			name: server.alias,
 			$root: this.$tabsContainer,
 			onClick: this.activateLastTab.bind(this, index),
 			index,
 			tabIndex,
-			onHover: this.onHover.bind(this, index, server.alias),
+			onHover: this.onHover.bind(this, index),
 			onHoverOut: this.onHoverOut.bind(this, index),
 			webview: new WebView({
 				$root: this.$webviewsContainer,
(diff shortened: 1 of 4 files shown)


app/renderer/js/ans,txt
```diff
@@ -0,0 +1,78 @@
+1. 127.0.0.1 0.047 ms
```
akashnimare: @abhigyank what's the purpose of this file?

akashnimare (changes requested):

app/renderer/js/ans,txt
abhigyank: This is some other file that came into the repository folder by mistake I guess. My bad.
Will remove and push.

akashnimare (approved):

app/renderer/js/main.js
```diff
@@ -530,6 +533,27 @@ class ServerManagerView {
 			webContents.send('toggle-dnd', state, newSettings);
 		});
 
+		ipcRenderer.on('update-realm-name', (event, serverURL, realmName) => {
+			DomainUtil.getDomains().forEach((domain, index) => {
+				if (domain.url.includes(serverURL)) {
+					const serverTooltipSelector = `.tab .server-tooltip`;
+					const serverTooltips = document.querySelectorAll(serverTooltipSelector);
+					serverTooltips[index].innerHTML = escape(realmName);
+					this.tabs[index].props.name = escape(realmName);
```
akashnimare, line 542: Do we really need to escape this and the `webview.props.name` ?

abhigyank: This was to maintain uniformity and be on the safe side in case.

abhigyank: I think we should remove it for the `webview.props.name` since it is being used by the Windows Menu.

akashnimare:
Thanks, @abhigyank.

timabbott:
Is there a zulip/zulip commit related to this that we need to merge as well?

abhigyank:
No, the required part in zulip/zulip was merged earlier.

Issue #554, opened 2018-09-03
Unescape server names in Window menu item


akashnimare:
This is follow up for #541.

PR #555, opened 2018-09-03
server-name: Unescape server name in window menu item.

akashnimare:
Escaping is necessary to avoid any security risk but we need
to unescape those strings in order to show them in the frontend
otherwise it will have ugly special characters.

We already escape server name in the db and unesacoe it in
the left-sidebar. This PR adds the decodeString function in
order to unescape strings in the menu items.

Fixes: #554.

---

**What's this PR do?**

**Screenshots?**

  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [x] macOS


abhigyank (commented):
This works. 
But similar to `escape-html` that we use, there is also [`unescape-html`](https://www.npmjs.com/package/unescape-html) that we can use directly or use inbuilt `unescape` instead, Is there is reason fdor going with this method?

akashnimare:
> directly or use inbuilt unescape instead

No, this doesn't work.

> Is there is reason for going with this method?

Not particularly but I feel like the browser's parsing method is better than the `unescape-html` here.

Issue #563, opened 2018-09-18
Auto hide menu bar on Windows/Linux


akashnimare:
It would be great to auto-hide the menu bar on Windows and Linux.

PR #564, opened 2018-09-18
menubar: Auto hide menubar on Windows/Linux.

akashnimare:
This PR adds a setting option to auto-hide the menubar.

Fixes: #563.

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -110,11 +110,6 @@ function createMainWindow() {
 
 	win.setTitle('Zulip');
 
-	// Auto-hide menu bar on Windows + Linux
-	if (process.platform !== 'darwin') {
-		win.setAutoHideMenuBar(ConfigUtil.getConfigItem('autoHideMenubar'));
-	}
-
 	win.on('enter-full-screen', () => {
 		win.webContents.send('enter-fullscreen');
 	});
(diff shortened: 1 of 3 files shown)


akashnimare:
As of now, toggling the setting doesn't update the menubar.

akashnimare:
Okay, added the IPC event to update the settings. It works as expected in all the cases except one -
when you turn off the setting it needs two clicks. Also, the other weird behaviour I have noticed is that if the setting is turned on and menubar is open and if you click anywhere the menubar gets hidden for some reason. Not sure what but something is wrong with this click event.

abhigyank (commented):
Linux (Gnome Environment) - 
* Works as expected. However, when I press Alt with the settings turned on, and I click anywhere (other than the menu) shouldn't the behavior should be that the menu bar hides? Currently it hides only when I press Alt again (which is the expected behaviour of the PR currently).

* We are calling `setMenuBarVisibility` on every toggle, we can maybe optimize it to call it only if it is hidden when the setting is turned off using `isMenuBarVisible`. It is a probably a very low level optimization that may not make much difference.

akashnimare:
> Works as expected. However, when I press Alt with the settings turned on, and I click anywhere (other than the menu) shouldn't the behavior should be that the menu bar hides? Currently it hides only when I press Alt again (which is the expected behaviour of the PR currently).

On Windows, this is what I'm getting -
* Turn on the auto-hide menubar setting
* Press `ALT` to toggle the menubar 
* If the menubar is visible and if I click anywhere in the app it hides the menubar

akashnimare:
> and I click anywhere (other than the menu) shouldn't the behavior should be that the menu bar hides? 

This creates a bit of problem when you trying to turn off the auto-hide menubar setting since the first click is hijacked by the alt thing (first click will hide the menubar) so you need to click twice in order to turn off this setting.

akashnimare:
I'm merging this for now. To-do is to find that double click bug and see if we can fix it. The one other thing I have noticed is that we only need the double clicking thing for the toggle switches and if you click on the `+` icon when the menubar is open it hides the menubar first and then fires the `+` event (in that case we don't need to double click).

Issue #244, opened 2017-08-01
Do another pass of revising the "About" page for the desktop app.


timabbott:
Here's what it currently looks like:

![image](https://user-images.githubusercontent.com/2746074/28846600-e81af384-76c0-11e7-9c24-95c7ef1076f7.png)

@rishig @brockwhittaker we should talk about what should go on this page and make a design.  I think probably 3 segments make sense:
* About Zulip -- probably just a link to zulipchat.com and/or zulipchat.com/about
* About the app -- probably just a nice paragraph explaining this app is open source software under the Apache 2 license, with a GitHub link
* How to get help and report bugs.  Which may belong on another more visible element.

I think going for clean and mostly linking to other resources is a good way to keep effort low.

akashnimare:
@rishig can you please give me the content so that I can just copy paste there 😄

rishig:
What do other apps do here? It's possible About pages have fallen out of favor; I can't find spotify's in their desktop app, and chrome's is buried two layers deep and just looks like
![image](https://user-images.githubusercontent.com/890911/33002059-9fc6a22a-cd66-11e7-8705-c94a97c680d4.png)

timabbott:
I would consider looking at apps that are more similar in purpose like Nylas or Slack.

priyank-p:
@zulipbot claim.

akashnimare:
Added in #346.

PR #567, opened 2018-09-25
about: Remove Found bug button.

rishig:
Haven't tested this out. Was going to open an issue, but realized opening a PR was just as fast.

I think the about page is not a typical place for a "report issue" button, and it looks sort of off to see it there.

### DIFF
--- a/app/renderer/about.html
+++ b/app/renderer/about.html
@@ -19,7 +19,6 @@
 				Available under the
 				<a onclick="linkInBrowser('license')">Apache 2.0 License</a>
 			</p>
-			<a class="bug" onclick="linkInBrowser('bug')" href="#">Found bug?</a>
 		</div>
 	</div>
 	<script>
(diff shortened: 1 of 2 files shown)


akashnimare:
@timabbott should we remove this? I think we decided to keep it in the page  -

https://github.com/zulip/zulip-electron/issues/244#issue-247199318

rishig:
We should move this to menu item to **Help** (independent of this change), and the **Help** menu already has a "Report an issue" menu item.

rishig:
`About Zulip` has now been moved to Help, so I think we can remove this button now. We can wait for Tim to confirm if you'd like.

For context, the Help menu is now:
```
Zulip Desktop v2.3.82
About Zulip
Help Center
Report an Issue
```

akashnimare:
Make sense. Merged, thanks @rishig.

PR #568, opened 2018-09-25
menus: Rename a few menu items.

rishig:
Have a few more suggestions to make, but these were faster to propose this way. Feel free to remove any changes you disagree with. 

The reason for the tray changes is that it currently just looks like the line spacing is broken.


akashnimare:
Merged after fixing minor issues. Thanks, @rishig.

Issue #533, opened 2018-08-03
Error: Invalid URI "../renderer/img/icon.png"


abhigyank:
This error occurs when we try to `saveServerIcon` with `../renderer/img/icon.png` (which is a fallback icon in case we are unable to get the icon while adding the server) as `url` parameter, since we then do  `request` to this file which leads to this error obviously. 
I think we can have a check not to do a request to the request if the url is this file. This will reduce this error.

abhigyank:
@zulipbot  label "good first issue" "Type: Bug"

aashish-ak:
@zulipbot claim

zulipbot:
Hello @aashish-ak, you have been unassigned from this issue because you have not updated this issue or any referenced pull requests for over 14 days.

You can reclaim this issue or claim any other issue by commenting `@zulipbot claim` on that issue.

Thanks for your contributions, and hope to see you again soon!

ShivangiKakkar:
@zulipbot claim

zulipbot:
Welcome to Zulip, @ShivangiKakkar! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-electron/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* Sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/), so that Zulip can use your code.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

zulipbot:
Hello @ShivangiKakkar, you have been unassigned from this issue because you have not updated this issue or any referenced pull requests for over 14 days.

You can reclaim this issue or claim any other issue by commenting `@zulipbot claim` on that issue.

Thanks for your contributions, and hope to see you again soon!

ghost:
@zulipbot claim

zulipbot:
[...]

PR #575, opened 2018-09-28
Altered path to default Icon

ghost:
**Altered the path of defaultIconUrl**

Issue in focus: #533 
The default image didn't seem to load on referring. I tried my level best to find the fault by backtracking through all the reference of the `defaultIconUrl` but got lost on my way. I realised this might have been caused as a result of the previous path `../renderer/img/icon.png` not going back to the image, probably due to some previous change in the structure. Now using `__dirname` we get the complete path of the file that allows us the access to image from any corner of the code base.
Also this was the only thing that I could track back to that may have caused this issue.

  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/utils/domain-util.js
+++ b/app/renderer/js/utils/domain-util.js
@@ -20,7 +20,7 @@ const logger = new Logger({
 
 let instance = null;
 
-const defaultIconUrl = '../renderer/img/icon.png';
+const defaultIconUrl = __dirname + '/../../img/icon.png';
 
 class DomainUtil {
 	constructor() {


app/renderer/js/utils/domain-util.js
```diff
@@ -20,7 +20,7 @@ const logger = new Logger({
 
 let instance = null;
 
-const defaultIconUrl = '../renderer/img/icon.png';
+const defaultIconUrl = __dirname.slice(0, -9) + '/img/icon.png';
```
akashnimare: @flowkraD can you document why you are using the `(0, -9)` range?

akashnimare (changes requested):

app/renderer/js/utils/domain-util.js
ghost: @akashnimare Would a comment above the code do the job or something additional is required?

ghost: I added a comment for now. Any changes or addition required can be mentioned and will be implemented at my earliest convenience.
Also i squashed the two commits - the changes to the code and the one with comments - into one.
Being new to Open-Source, I will highly value any advice you give me. Thank you (^_^)

app/renderer/js/utils/domain-util.js
```diff
@@ -20,7 +20,10 @@ const logger = new Logger({
 
 let instance = null;
 
-const defaultIconUrl = '../renderer/img/icon.png';
+// __dirname returns the path to the directory similar to: '/home/user/zulip-electron/app/renderer/js/utils'
+// Slicing it from 0 to -9 gives us just '/home/user/zulip-electron/app/renderer'
+// To this, we concatinate '/img/icon.png' with which now one can refer to the fallback image from anywhere in the repository
+const defaultIconUrl = __dirname.slice(0, -9) + '/img/icon.png';
```
abhigyank: I think the path was maybe inserted wrong or changed at some point of time.
An easier and should fix should be `const defaultIconUrl = '../../img/icon.png';` or `const defaultIconUrl ='__dirname + '../../img/icon.png';` . Slicing the substring, feels a bit hackish way of doing things.
What do you think @flowkraD ?

app/renderer/js/utils/domain-util.js
flowkraD: It does feel like a unthoughtful hack but using `__dirname` and slicing it, we get the whole path and one that is Independent of where you are accessing it from. You can refer to the `defaultIconUrl` from anywhere in the code base and it points to the same image. I thought this way would make it bullet proof for future in case this path is referred at any place that isn't coved by declaring it as just `../../img/icon.png`
I you know of any better way, I would love to know and make changes accordingly (^_^)

app/renderer/js/utils/domain-util.js
ghost: It does feel like a hack but using `__dirname` and slicing it, we get the whole path and one that is Independent of where you are accessing it from. You can refer to the `defaultIconUrl` from anywhere in the code base and it points to the same image. I thought this way would make it bullet proof for future in case this path is referred at any place that isn't coved by declaring it as just ../../img/icon.png which completely depends on the `pwd`.
The second approach looks good:`const defaultIconUrl ='__dirname + '../../img/icon.png'; ` Do you mind if I i implement the same?

akashnimare:
Quick note - whatever approach you follow, please try to test the same on Windows if possible.

app/renderer/js/utils/domain-util.js
abhigyank: Yeah, second approach is preferable according to me.

ghost:
I currently don't have a Windows device in sight but I'll make sure it is the first thing i do tomorrow

ghost:
Sorry for no update for so long but __dirname returns path with backslash like `c:\home\` and concatenating  as shown above will give a mixture of `\ /`. There is a way to get the path in UNIX format using `normalize()` but I think that will not help in fetching the image on windows.

priyank-p:
@flowkraD if you use node's `path` module this could be done in manner that could work cross-platform by using `path.resolve` method like: `path.resolve(__dirname, '../renderer/img/icon.png')`

ghost:
Thanks for the help @priyank-p. Will submit a new PR with the change. Just wanted to confirm, is it 100% cross-platform. I have very little access to windows machine so it's hard for me to check this.
Thank for your help, really appreciate it (^_^)

priyank-p:
Yes, It is cross-platform since the method is part of core node.js module.

ghost:
Thanks a lot for that confirmation and guidance  (^_^)

akashnimare:
@flowkraD feel free to send a PR and I can test the same on Windows.

PR #576, opened 2018-10-02
readme: Replace the word team with organizations.

abhigyank:


akashnimare:
@rishig we are using organizations everywhere instead of `team`, right?

rishig:
yeah. We might change it in the future, but it's currently all standardized on that.

Issue #585, opened 2018-10-17
settings: Change get beta update option on updating enable update setting


akashnimare:
If the `Enable auto updates` option is disabled then we should disable the beta updates option also. 
This case is already handled in the backend so even if the beta option is enabled and the auto-update option is disabled, the auto-updates won't work. We just need to make changes in the frontend.

kanishk98:
@zulipbot claim

kanishk98:
Okay, looks like ZulipBot has run into a problem. 
Regardless, I've looked into the issue, and I think I've solved the problem. Changed the 
`clickHandler()` in `autoUpdateOption()` in `app\renderer\js\pages\preference\general-section.js` to the following block:

```
clickHandler: () => {
	const newValue = !ConfigUtil.getConfigItem('autoUpdate');
	ConfigUtil.setConfigItem('autoUpdate', newValue);
	if (!newValue) {
		ConfigUtil.setConfigItem('betaUpdate', false);
		this.betaUpdateOption();
	}
	this.autoUpdateOption();
```
@akashnimare can you please have a look at this solution and tell me if I'm missing something before I submit a PR?

akashnimare:
GitHub is down right now so the comments are not being shown in the GitHub
(replying through the email). I would suggest playing with the solution locally
and testing if the values are getting updated as expected.

akashnimare:
Fixed in https://github.com/zulip/zulip-electron/pull/586.

PR #586, opened 2018-10-23
Disable beta updates if auto updates disabled

kanishk98:
---

**If a user disables automatic updates in Preferences > General, they will no longer receive beta updates either.**

**Fixes issue #585**

As of now, the default behavior in the General pane allows the following setting:

old

This PR disallows the above and enforces the following:

new

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/pages/preference/general-section.js
+++ b/app/renderer/js/pages/preference/general-section.js
@@ -233,6 +233,10 @@ class GeneralSection extends BaseSection {
 			clickHandler: () => {
 				const newValue = !ConfigUtil.getConfigItem('autoUpdate');
 				ConfigUtil.setConfigItem('autoUpdate', newValue);
+				if (!newValue) {
+					ConfigUtil.setConfigItem('betaUpdate', false);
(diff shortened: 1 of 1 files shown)


akashnimare:
Thanks for working on this. There is one more case we need to handle which is when the auto-update is off then the beta option should be disabled (cursor) so that user can't update this setting -

kanishk98:
@akashnimare Sorry for the trouble, should've taken care of that. I've pushed another commit, hope that solves the problem.

akashnimare:
Looks good now. Would it be possible to have the not-allowed cursor on the beta option when the update setting is off?

image

kanishk98:
Shouldn't be a problem. Will push another commit asap.

akashnimare:
Merged. Thanks, @kanishk98.

akashnimare:
> Would it be possible to have the not-allowed cursor on the beta option when the update setting is off?

@kanishk98 this can be added as a follow-up.

Issue #594, opened 2018-11-16
SSL Handshake Error


superbiche:
Whenever I try to use the Desktop client to login to my Zulip instance, I get the "{domain} does not appear to be a valid Zulip server". The Web based application works totally fine.

After debugging, the error seems related to SSL handshake failure.

@gnprice pointed that it could be related to https://github.com/nodejs/node/issues/16196, which it was: `request` is trying to use Node's default ECDH curve, which is `prime256v1` on Node > 8.5 and is `secp384r1` in any recent Nginx configuration following https://cipherli.st/ recommendations.  

For now, a quick & dirty workaround is to add `prime256v1` to `ssl_ecdh_curve` in nginx.conf like this: ` ssl_ecdh_curve prime256v1:secp384r1;` (adding it at in the server block didn't help so I had to add it to the http block)  

Also, adding `ecdhCurve: "auto"` option to `request` call works, and avoids the need to add the 
`prime256v1` curve to nginx.conf: 
```javascript
// app/renderer/js/utils/domain-util.js:129
const checkDomain = {
    url: domain + '/static/audio/zulip.ogg',
    ca: (certificateLocation) ? certificateLocation : '',
    proxy: proxyEnabled ? ProxyUtil.getProxy(domain) : '',
    ecdhCurve: 'auto' // Fix
};
```
---
- **Operating System**: macOS 10.14.1

See https://chat.zulip.org/#narrow/stream/31-production-help/topic/nginx-let's.20encrypt-desktop.20client

akashnimare:
@superbiche sorry for the slow reply (just recovering from the surgery I recently had). I went through with that node issue on GH and the discussion on chat.zulip.org and it seems that using the `auto` flag of `ecdhCurve` could fix this issue. Feel free to open a PR with this fix :) Also, is there any quick way to reproduce this issue in the desktop app so that I can cross-verfiy the solution?

superbiche:
Hey @akashnimare no problem, hope you're gonna quickly and fully recover 👍 
I could make a PR, maybe tonight or tomorrow.

To reproduce, you should try connecting to https://zulip.superbiche.co with the desktop app in a few minutes. I'm going to remove `prime256v1` so you can reproduce - and so I'm forced to do the fix PR

superbiche:
@zulipbot claim

zulipbot:
Hello @superbiche!

[...]

PR #595, opened 2018-11-18
network: Fix `request` ecdhCurve mismatch errors

superbiche:
Fix `request` ecdhCurve mismatch errors by passing `ecdhCurve: 'auto'` to options

As investigated in #594, adding a self-hosted organisation fails if the app is behing a [Nginx >= 1.10.0 ](http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_ecdh_curve) reverse proxy with config `ssl_ecdh_curve secp384r1;` which is [the recommended ecdh curve](https://cipherli.st/) and if Node > 8.5 is used. 

Setting `ecdhCurve` to `"auto"` leaves the ecdh curve choice and its security outcomes on the hosting server admin.  
I suppose this could lead to weak ecdh curves used - I don't know a bit about whether this exists and is supported - so maybe we could try to detect this, but I don't know if we can, and how.
So I guess the best option is to ensure secure ecdh curves are properly supported thanks to the 'auto' option, without breaking for other users using weaker curves.

Tested on 2.3.82 Mac OS Mojave 10.14.1, running smooth, added, removed and re-added my org without any issue.

### DIFF
--- a/app/main/linuxupdater.js
+++ b/app/main/linuxupdater.js
@@ -21,7 +21,8 @@ function linuxUpdateNotification() {
 	const options = {
 		url,
 		headers: {'User-Agent': 'request'},
-		proxy: proxyEnabled ? ProxyUtil.getProxy(url) : ''
+		proxy: proxyEnabled ? ProxyUtil.getProxy(url) : '',
+		ecdhCurve: 'auto'
 	};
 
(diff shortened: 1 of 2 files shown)


akashnimare:
Merged. @superbiche, thanks for working on this 🚀

Issue #561, opened 2018-09-15
Add a feature to configure DND mode


akashnimare:
We have a basic DND feature in the left sidebar which activates the DND mode and to deactivate user need to update it manually. It would be great to have something like -

* Enable DND mode for X minutes/hrs/day

and then automatically handles the notification setting after the DND time gets over.

Two things we need to do here -
**a)** UI/UX - I think we can show the time within the tooltip itself
**b)** How to calculate the DND time since there could be multiple cases like the laptop is switched off/sleep etc.

![image](https://user-images.githubusercontent.com/2263909/45583972-096f9b00-b8e9-11e8-8986-61c8df68c750.png)

priyank-p:
To implement this I have/had a rough plan in mind:

  * First we should create the UX/UI in a new branch or that only shows up in development mode.

  * Then, once we have the UI, we can set up the turn on/off thing per input provided by the user.

  * Lastly, we should save in `JsonDB` the end-time of DND time, so it just a one if statement on the start of the app to turn it off or schedule a timer if it not done yet.

For step two, we can use something like `node-schedule` to do turn off the DND button after the specified time, in which case if the app is closed or device shut-down it will not work. Which we will need to cover in step three.

lukasIO:
I know this is the repository for the electron app, but can you clarify why you see DND as an app setting rather than a profile-setting?
Imho it leads to confusion and unwanted behaviour if there is a DND layer above a profile setting.
If I don't want to be disturbed, then I do so in my online status which should then reflect across all apps.

similarly if i want to receive notifications again, I should be able to do so on mobile and the changes should reflect on desktop. 
that's why I think this should be followed further (only) as a profile setting and not as an app setting.

akashnimare:
@lukasIO please see the relevant conversations here - https://chat.zulip.org/#narrow/stream/16-desktop/topic/DND.20button

lukasIO:
Thanks for the link. As it seems the main reasonis this statement by @timabbott 

> I think it's actually probably useful to have a desktop app feature for this, because if you're in multiple organizations, you kida want to be able to turn it off for all of them at once.

I agree that this is a very good point, which i didn't consider before. 
I would still prefer if this kind of status-change would reflect consistently over web, desktop and mobile.
How to communicate over multiple organizations a common state - I have no idea.
An approach to change an additonal notification-status in each logged in organization would be an option, maybe.

akashnimare:
[...]

PR #602, opened 2018-12-01
adds support for time dependent DND mode

harkishen:
---

**What's this PR do?**
This PR solves #561 . It allows the User to set a time limit for DND mode, which would be switched off, once the time exceeds the limit.
This checks every min the current time, and compares with the set of conditions, whether to turn off the DND mode or not.
**Screenshots?**
GIF ( 9 slides )
webp net-gifmaker

Though, I have tested this feature in Fedora 28, but it is platform independent.


abhigyank:
Hello @Harkishen-Singh , thank you for working on this. Before reviewing this could you please do the following - 
* Remove the package-lock file from the PR (it shouldn't be required in this PR).
* functions `showDNDTimeLeft` and `checkDNDstate` are repeated in two files. I'd suggest to put it in one file, preferaby in dnd-util and use from there (I haven't read their workings but I feel it shouldn't be require in both the places).
* Also could you give a brief workflow of how you have implemented the feature? Specially I am curious how you are handling if a user shuts the app down during a dnd time period.

harkishen:
@abhigyank , sure, i will put the two functions in a single file.

**Work Flow**

On start of the application, the database is check with respect to the **dndSwitchOff** collection, and if no record is found, a null is returned, which skips the further checks, hence the UI part is not prepended to the actions-container. If the record is found, then it check the nature of the record, which would satisfy if it contains the object containing the 3 parameters: hr, min, carry. 

The variables hr, min have the function as their names go. However, the carry parameter is included for a particular case; if the user select a time out for the DND mode which lies after the 12 am or 0:00, then the carry is added by 1, and the hour is saved as ( hour - 24 ). The reason or increasing the carry to one is, to pass all the time checks, which are occurring every minute by _checkDNDstate_ function. As soon as the time passes 12am or 0:00, the check.carry is check that if it has the value greater than 0 and the day has changed, then re-initialized back to 0 for normal functional of the program.

For the part _"handling if a user shuts the app down during a dnd time period"_ , whenever the user starts the application, a check is made from the **main.js** renderer process on whether the previous DND mode existed ( will return an object of params: hr, min, carry ) or not ( will return null ). This clarifies, the decision on whether the tag of **time left** and the **per minute time count** would be executed or not. 
The **sleep()** has been used so as to give a good UI feel, and also to prevent some async conflicts.

**Edit 1**
_Removed the _package-lock.json_ which was added in a mistake in previous commits._

**Edit 2**
_Added Cancel button, Minor UI changes_
cancel

app/renderer/js/utils/dnd-util.js
```diff
@@ -2,30 +2,168 @@
 
 const ConfigUtil = require(__dirname + '/config-util.js');
 
+const {ipcRenderer} = require('electron');
+
+let count = 0;
```
abhigyank: What is the need of this variable? It seems to be increasing. I don't see a purpose of this variable.

app/renderer/js/utils/dnd-util.js
```diff
@@ -2,30 +2,168 @@
...
+let	dnd;
```
abhigyank:
```suggestion
let dnd;
```

app/renderer/js/utils/dnd-util.js
```diff
@@ -2,30 +2,168 @@
...
+let newSettings;
+let time = -1;
+function makeView() {
+	const dndOffTime = {
```
abhigyank: This piece of code shouldn't be here as it gets the time even before the user chooses the time delay. It could setoff a some amount of less dnd time.

app/renderer/js/utils/dnd-util.js
```diff
@@ -2,30 +2,168 @@
...
+	if (opt.id === 'dnd_1') {
+		time = 1;
+	} else if (opt.id === 'dnd_2') {
+		time = 8;
+	} else if (opt.id === 'dnd_3') {
+		time = 12;
+	} else {
+		time = -1;
+	}
+	if (time > 0) {
```
abhigyank, line 71: The `dndOffTime` could be somewhere here.

app/renderer/js/utils/dnd-util.js
```diff
@@ -2,30 +2,168 @@
...
+}
+
 function toggle() {
 	const dnd = !ConfigUtil.getConfigItem('dnd', false);
 	const dndSettingList = ['showNotification', 'silent'];
 	if (process.platform === 'win32') {
 		dndSettingList.push('flashTaskbarOnMessage');
 	}
-
-	let newSettings;
```
abhigyank, line 12: Any reason to make this a global variable?

app/renderer/js/utils/dnd-util.js
```diff
@@ -2,30 +2,168 @@
...
+	};
+	const timeLeft = document.createElement('span');
+	const timeDisableDNDInput = document.createElement('div');
+	const actionContainer = document.getElementById('actions-container');
+	timeLeft.id = 'timeLeft';
+	actionContainer.prepend(timeLeft);
+	timeDisableDNDInput.className = 'dndInput';
+	for (let i = 1; i <= 4; i++) {
+		const opt = document.createElement('span');
+		if (i === 1) {
```
abhigyank: These should be in a switch case. So many `if-else` is not preferred.

abhigyank (commented):
A quick look through the changes suggests the the logic of implementation is correct and would work correctly. 
However, I don't think this is an optimal implementation of the feature as the code doesn't look very clean and could be prone to issues. Also there has to a lot more changes to UI to make it look in coherence to app UI.  Some ordering and occurrences in the function calls in the `main.js` seems confusing and maybe redundant. 
@Harkishen-Singh For other members to review this you should probably add some comments describing the functions and variables use since the implementation seems a bit unusual to me.  Also could you squash all the commits into one?

app/renderer/js/utils/dnd-util.js
harkishen: the count variable was linked to earlier commit in order to prevent the time selection menu on startup of the app, when the DND mode was already enabled. But, in successive commits, this prob was solved in other way, when i removed the copy of the checkDND and showDND from main.js. 
Now, its of no use. Thanks for pointing out. i will make all the recommended changes and push.

app/renderer/js/utils/dnd-util.js
harkishen: will address this in final commit

app/renderer/js/utils/dnd-util.js
harkishen: by default, we assume the dnd time to be "selected by user" for which, -1 is taken. In the if statements, the DND time would be added only when the time is positive. So, it would never set-off a less time than the present. Addressed in the commit.

app/renderer/js/utils/dnd-util.js
harkishen: sure. The some variable declarations have been misplaced with the improve of code from the first check, when the CI tests failed, in order to satisfy the pre-commit conditions. Addressed in recent commit.

app/renderer/js/utils/dnd-util.js
harkishen: sure. I will address in next commit

app/renderer/js/utils/dnd-util.js
harkishen: now changed.

harkishen:
@abhigyank , I have made all the changes suggested. Please have a look, if it meets the requirements.
For the UI changes, please describe for further refinement.

akashnimare:
@Harkishen-Singh Thanks for working on this. I have not looked deep into the code but I would suggest following changes -
* Document the code/functions etc so that others can see what's going on
* Try to use the appropiate class/function's name + clean up the code
* We mostly use ES6 (arrow functions etc) so you should follow the same 
* This needs heavy changes in the design 

I would suggest discussing the code/implementations/designs etc in  https://chat.zulip.org/#narrow/stream/16-desktop and then make changes according to what we decide so that you don't need to make changes every now and then.

priyank-p:
@Harkishen-Singh Thanks for working on this! I have couple of suggestions:
1) I think creating the whole html for this feature in javascript, specially per element and with the lengthy and complex logic you have in the `makeView` function, is wrong way to do it. You should instead either include the html in the html file or if you really think it needs to be created in js side you should look at how we do it at [webview component](https://github.com/zulip/zulip-electron/blob/master/app/renderer/js/components/webview.js#L27-L39) which is clear way to do it and easier to review.

2) Also, for the checking the time to turn off this feature, you currently have very complex logic that is hard to comprehend hence as mentioned very error-prone. I suggest you use something like [node-schedule](https://www.npmjs.com/package/node-schedule) to schedule the turn off of dnd and save the end time in the `JsonDB` and remove the record once the dnd is turned off. And to handle computer shutdown and turn off cases you should check the saved record on app startup and if there are records you can handle the edge cases there.

3) As a sidenote you have css classes that are in camel case, we use hypen case so you should rename you css classes accordingly like, for example, `.dnd-input` instead of `.dndInput`.

harkishen:
@priyank-p , thanks for the suggestions!
There are few things to note:
**First**, the View part of the Time List is in dynamic way so as to give a more emphasis on the DOM control, which could be done very effectively, if the entire is based in js. Further, it would also allow more space for addition of more features in the same view dynamically in a easy manner by the future contributors. Still, if that's useful, the html way, i will surely do that way. 
**Second**, the Time checking Logic is working in a plain simple way, it extracts the time from the database (if existing) and converts that into the string format, and concatenates the hr with the min. This makes it very easy to compares the two time moments, by converting into the Integer format. The same is done with the present current time during the comparison. I had actually used [node-schedule,](https://www.npmjs.com/package/node-schedule) but later felt not necessary, since that would be an addition of the size to the final electron-package after the builds as that could be solved by mere converting to the string type as described above.
**Third**, I would make those css class changes in the next commit, once the UI is confirmed.
Suggestions are most welcomed!

akashnimare:
Closing in favour of https://github.com/zulip/zulip-desktop/pull/896. Thanks for putting the efforts on this though 👍

PR #605, opened 2018-12-03
domain-util: Include certificates in all requests for icon.

### LINKED ISSUE #463: Realm icon is not set in sidebar despite the icon existing on server
The realm icon fallbacks to default icon (Zulip Logo) in the sidebar even when a server icon exists at the required place.
Discussion on developer chat can be found here [here](https://chat.zulip.org/#narrow/stream/31-production-help/subject/electron.20server.20icon).

Two possible causes for this - 
* There is path issue in the icon URL which needs to be handled. Refer [here](https://chat.zulip.org/#narrow/stream/31-production-help/subject/electron.20server.20icon/near/548293)

* Server icon url is not being respected due to self signed certificate. Refer [here](https://chat.zulip.org/#narrow/stream/31-production-help/subject/electron.20server.20icon/near/548336)

abhigyank:
Fixes: #463
This fixes the issue as now the certificate is included in the request, so it is able to retrieve the icon. Tested on abhigyank.zulipdev.org

### DIFF
--- a/app/renderer/js/utils/domain-util.js
+++ b/app/renderer/js/utils/domain-util.js
@@ -65,7 +65,7 @@ class DomainUtil {
 	addDomain(server) {
 		return new Promise(resolve => {
 			if (server.icon) {
-				this.saveServerIcon(server.icon).then(localIconUrl => {
+				this.saveServerIcon(server).then(localIconUrl => {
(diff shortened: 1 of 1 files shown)


akashnimare:
Works as expected. Merged, thanks, @abhigyank.

Issue #599, opened 2018-11-26
Add context menu in left sidebar's server tab


akashnimare:
We should show the menu options when the user right clicks on the added zulip servers in the left sidebar. We could add multiple items there but for now, adding a `Delete` option which basically removes that org from the app.

kanishk98:
@zulipbot claim

kanishk98:
@akashnimare I picked this up right now, and I was able to get the required context menu. However, when I click on 'Delete', there's no immediately visible action, even though I can confirm that the `removeDomain()` function is getting called. If I kill the app and relaunch it, the organisation is deleted. 

Here's what I'm doing in `app/renderer/js/main.js`:

```
                const $serverImgs = document.querySelectorAll(".server-icons");
		$serverImgs.forEach(($serverImg, index) => {
			$serverImg.addEventListener("contextmenu", e => {
				e.preventDefault();
				const template = [
					{
						label: "Delete organization",
						click: () => {
							DomainUtil.removeDomain(index);
						}
					}
				];
				const contextMenu = Menu.buildFromTemplate(template);
				contextMenu.popup(remote.getCurrentWindow());
			});
			if ($serverImg.src.includes("img/icon.png")) {
				this.displayInitalCharLogo($serverImg);
			}
			$serverImg.addEventListener("error", () => {
				this.displayInitalCharLogo($serverImg);
			});
		});
```

I checked the code used in `server-info-form.js` to delete a domain, and there seems to be some `onChange()` method that's causing the immediate reload we want. 
I'm not sure on how to implement that here, and would love some help.

akashnimare:
This looks good. Maybe reuse the following function -
```js

		this.$deleteServerButton.addEventListener('click', () => {
			dialog.showMessageBox({
				type: 'warning',
				buttons: ['YES', 'NO'],
				defaultId: 0,
				message: 'Are you sure you want to disconnect this organization?'
			}, response => {
				if (response === 0) {
					DomainUtil.removeDomain(this.props.index);
					this.props.onChange(this.props.index);
				}
			});
		});
```
You could also try to reload the app using -
```
click: () => {
		DomainUtil.removeDomain(index);
                ipcRenderer.send('reload-full-app');
	}
```

kanishk98:
@akashnimare works now, thanks. :)
[...]

PR #609, opened 2018-12-09
Fix #608

### LINKED ISSUE #608: Context Menu of sidebar doesn't work in letter icon
A followup issue to #604 .
We fallback to initial letter of the realm in case the icon isn't available, in that case the context menu doesn't work. We should probably include an event listener for context in that case too.

kanishk98:
---

**What's this PR do?**

Adds context menu to previously ignored letter server icons.

#608 and #599 should help understand the issue. Just a simple context menu addition. 

  - [x] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

P. S. My Sublime Text plugin changed the formatting of the code a bit. Please let me know if that's a problem for Zulip.

### DIFF
--- a/app/renderer/js/main.js
+++ b/app/renderer/js/main.js
@@ -139,6 +139,10 @@ class ServerManagerView {
 			settingOptions.dockBouncing = true;
 		}
 
+		if (process.platform !== 'darwin') {
+			settingOptions.autoHideMenubar = false;
+		}
+
 		for (const i in settingOptions) {
 			if (ConfigUtil.getConfigItem(i) === null) {
 				ConfigUtil.setConfigItem(i, settingOptions[i]);
--- a/app/renderer/js/main.js
+++ b/app/renderer/js/main.js
@@ -115,6 +115,7 @@ class ServerManagerView {
 			showNotification: true,
 			autoUpdate: true,
 			betaUpdate: false,
+			customCSS: false,
 			silent: false,
 			lastActiveTab: 0,
 			dnd: false,
--- a/app/renderer/js/main.js
+++ b/app/renderer/js/main.js
@@ -14,6 +14,8 @@ const ConfigUtil = require(__dirname + '/js/utils/config-util.js');
 const DNDUtil = require(__dirname + '/js/utils/dnd-util.js');
 const ReconnectUtil = require(__dirname + '/js/utils/reconnect-util.js');
 const Logger = require(__dirname + '/js/utils/logger-util.js');
+const CommonUtil = require(__dirname + '/js/utils/common-util.js');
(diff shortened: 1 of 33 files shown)


akashnimare:
Can you take care of the linting first? Push changes related to this fix only. Also, have a look -
https://zulip.readthedocs.io/en/latest/contributing/version-control.html

kanishk98:
Of course. 

> Push changes related to this fix only. 

Sorry. I'll keep that in mind next time.

app/renderer/js/main.js
```diff
@@ -207,35 +207,12 @@ class ServerManagerView {
...
-						}
-					}
-				];
-				const contextMenu = Menu.buildFromTemplate(template);
-				contextMenu.popup({ window: remote.getCurrentWindow() });
-			});
+			this.addContextMenu($serverImg, index);
 			if ($serverImg.src.includes('img/icon.png')) {
-				this.displayInitialCharLogo($serverImg);
+				this.displayInitialCharLogo($serverImg, index);
```
abhigyank: The `index` variable is redundant in the `displayInitialCharLogo` function. The `addContextMenu` function is anyway being call at the end of that function so it makes sens to call the `addContextMenu` function here after this call to  `displayInitialCharLogo`. That way the flow makes sense and the use of redundant `index` variable is not required.

app/renderer/js/main.js
```diff
@@ -207,35 +207,12 @@ class ServerManagerView {
...
 			}
-
 			$serverImg.addEventListener('error', () => {
-				this.displayInitialCharLogo($serverImg);
+				this.displayInitialCharLogo($serverImg, index);
```
abhigyank: Same here as above.

abhigyank (commented):
The functionality works correctly for me.

app/renderer/js/main.js
kanishk98: I get that the `index` variable is redundant, for which I've changed the function call to 
`this.addContextMenu($altIcon, $webview.getAttribute('data-tab-id'));`

I don't understand what you mean by the point about the flow though. Since the `$altIcon` is rendered instead of `$serverImg`, isn't it required to attach the context menu to `$altIcon`?

abhigyank: Oh yeah, my bad. That feels correct then! :+1:

kanishk98: Awesome. I'll push the new changes.

kanishk98:
I've evidently messed up a lot of stuff in this branch. Sorry about that. Closing this PR in favour of https://github.com/zulip/zulip-electron/pull/619.

PR #610, opened 2018-12-12
Typescript

priyank-p:
This adds typescript support for both main and renderer process. For now the `app/main/index.js` cannot be written in typescript we can make it work with minor change once we get there.


akashnimare:
Would it make sense if we put the `ts-node.js` into `shared` or maybe `common`. directory?

```
/* To make the util runnable in both main and renderer process */
if (process.type === 'renderer') {
	const remote = require('electron').remote;
	app = remote.app;
} else {
	const electron = require('electron');
	app = electron.app;
}
const tsNode = require('ts-node');
tsNode.register({
	cacheDirectory: app.getPath('userData')
});
```

akashnimare:
This looks good to me, will merge this after pushing a release from master.

PR #612, opened 2018-12-17
Update electron to v3.0.10.

priyank-p:
This also takes care of the issue where upgrading to v3 caused electron to crash.


akashnimare:
Works well in the dev but after building the dmg installer -

image

Built using node - `v8.9.4`

akashnimare:
Works as expected on macOS.

akashnimare:
Thanks @priyank-p 🚀

Issue #565, opened 2018-09-24
[Feature request] Minimize to tray on startup


zingmars:
Right now there's a "Always start minimized" option which does just that - it starts the Zulip app in a minimized state, however it still remains in the task bar and can be switched to using alt-tab and has to be closed to tray manually every time it's started (i.e. on system startup). Would it also be possible to get an option that minimizes the app to tray on startup so one doesn't have to do it manually every time?

akashnimare:
I'm bit confused on what you're asking for. 

> On system startup, auto-start the Zulip app and minimize it to tray.

^^ is this what you want? If yes, then you need to use the following settings -

![image](https://user-images.githubusercontent.com/2263909/46066966-511cdf00-c193-11e8-89ff-64abb665760c.png)

zingmars:
Here's the current behaviour - https://youtu.be/8Mvp0woj9X8. As you can see Zulip opens and minimises to the taskbar, not tray. It can also be seen when alt+tabbing (0:05) since it's minimized to taskbar and not tray.

Here's the behaviour I'm asking for - https://youtu.be/YA5J66fBhTU. As you can see here, the thunderbird tray icon is created and it does not appear in the taskbar at all.

fauust:
Hi, I confirm that on i3wm the "Always start minimized" option is not working...

So maybe you could add a CLI option, for instance:
```
$ zulip -startintray
```

ThePiyushGupta:
@zulipbot claim

zulipbot:
Hello @ThePiyushGupta, it looks like you've currently claimed 1 issue in this repository. We encourage new contributors to focus their efforts on at most 1 issue at a time, so please complete your work on your other claimed issues before trying to claim this issue again.

We look forward to your valuable contributions!

ThePiyushGupta:
@zulipbot claim

ThePiyushGupta:
I have added the basic functionalities .However i would like to know if triggering the "always start minimized" should automatically turn of "start in tray" and vice versa.

zingmars:
That's how I would prefer it.

Maybe it should check which option is set on startup and act accordingly, so people can toggle it according to their desires.

akashnimare:
> I have added the basic functionalities .However i would like to know if triggering the "always start minimized" should automatically turn of "start in tray" and vice versa.

Correct, this should be the expected behaviour.

akashnimare:
[...]

Issue #540, opened 2018-08-14
Electron should present Network Settings before connecting first time


stone212:
---
Linux.  Using 2.3.5

Install Zulip, join an organization, enter email and password.  Zulip connects.

Instead you should be presented Network options.  Depending on your situation it could be a security problem to connect *without* setting up your proxy first.

timabbott:
I don't think we want to put a screen in front of "enter organization URL" -- it definitely doesn't make sense to add an extra step for regular users who don't know anything about their network setup, but it could be reasonable to have a small "network settings" option presented on that screen.  @rishig @gnprice what do you think?

akashnimare:
The other thing we could do here is that includes a link for the network setting page in the error dialog message, so that the user can easily change/update the network settings.

rishig:
yeah, I agree we should have some way of entering in network settings before connecting, and also that having a link to network settings on that initial page seems pretty good.

(and yeah, having a link in the error message also seems good, while we're fixing this)

stone212:
What "error message" are you talking about.  I didn't receive an error message.

Having the network settings option in the initial setup page would be good.  Even a "Do you want to configure your network?" link that brings you to an optional second page could be good.

ThePiyushGupta:
We can add a link to the add-server page to configure network when the user has no servers .
![screenshot from 2018-12-21 23-09-14](https://user-images.githubusercontent.com/16987573/50355738-aa6b9300-0575-11e9-8d24-5e719cb4f060.png)
Would this be fine?

akashnimare:
I'm not convinced on putting this link in the front page. Maybe we could add this within the error dialog? @timabbott, @rishig thoughts?

rishig:
I'm fine with putting a "Change network settings" left-justified, in that font size, on this page.

We do need to show the option somewhere before they add their first organization, right?

stone212:
@ThePiyushGupta I like @rishig 's suggestion:

```
I'm fine with putting a "Change network settings" left-justified, in that font size, on this page.

We do need to show the option somewhere before they add their first organization, right?
```

One small change is I would not write "Change network settings" but "Network and Proxy Settings".

ThePiyushGupta:
So should I finalize this design? 
On hover:
![screenshot from 2018-12-22 13-53-16](https://user-images.githubusercontent.com/16987573/50372243-4090cf00-05f1-11e9-8648-56eec1c27144.png)
else:
[...]

PR #613, opened 2018-12-19
Fix issue #565

ThePiyushGupta:
---

**What's this PR do?**
This PR fixes issue #565 

**Screenshots?**

  - [x] Linux/Ubuntu
  - [ ] Mac/Os
  - [ ] Windows

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -88,6 +88,8 @@ function createMainWindow() {
 	win.once('ready-to-show', () => {
 		if (ConfigUtil.getConfigItem('startMinimized')) {
 			win.minimize();
+		} else if (ConfigUtil.getConfigItem('startInTray')) {
+			win.hide();
 		} else {
 			win.show();
 		}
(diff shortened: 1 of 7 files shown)


akashnimare:
@ThePiyushGupta thanks for working on this. Few changes -
* Can you remove the changes related to the network setting? 
* We just need to hide the window when the start to minimized is on so you don't need to introduce that tray setting option.

ThePiyushGupta:
@akashnimare So what you mean is that alwaysstartminimised =alwaysstartintray and i don't need to create another option in settings.
Secondly all the changes related to issue #540 should be undone

akashnimare:
> So what you mean is that alwaysstartminimised =alwaysstartintray and i don't need to create another option in settings.

Correct. Also, checkout - https://zulip.readthedocs.io/en/latest/contributing/index.html for better commit guidelines.

ThePiyushGupta:
It's better i create a new pull request

Issue #548, opened 2018-08-28
Make the org tab list draggable in left sidebar 


akashnimare:
We should make the organization list in the left sidebar draggable so that users can reorder the added organizations.

![image](https://user-images.githubusercontent.com/2263909/44741855-fe142580-ab1b-11e8-9693-d2cd7f1d04a2.png)

Something like this - 

<img src="https://cl.ly/4cfebba97d93/download/Screen%20Recording%202018-08-28%20at%2011.44%20PM.gif"/>

On a technical level we need to make following changes when the user drags the org tabs -

* Update the index of org in the menu item

![image](https://user-images.githubusercontent.com/2263909/44742054-898db680-ab1c-11e8-9575-90dde07522cc.png)

* Update the index of org in the left sidebar 

* Update the tooltip

timabbott:
`sortable.js` is usually the way to do this sort of reordering.

kanishk98:
@zulipbot claim

kanishk98:
@timabbott I'm sorry, I don't understand. What do you mean when you say `sortable.js`? I can't find any file by that name in the repo.

abhigyank:
`sortable.js` is a JavaScript library for reorderable drag-and-drop ( they have a repo on GitHub) . Akash has already done a lot of work on this, you can check out the `org-tab-drop` branch in this repo and then continue from there.

akashnimare:
@kanishk98 yeah the dragging part is done in the [org-tab-drag](https://github.com/zulip/zulip-electron/commits/org-tab-drag). The remaining part is -
* Update the tooltips
* Update the index of the orgs

kanishk98:
I've been trying to solve the tooltips problem, and I'd like some clarity here:
the `initTabs()` method in `app/renderer/js/main.js` calls `initServer()` using the following code:
```
const servers = DomainUtil.getDomains();
if (servers.length > 0) {
	for (let i = 0; i < servers.length; i++) {
		this.initServer(servers[i], i);
		DomainUtil.updateSavedServer(servers[i].url, i);
		this.activateTab(i);
	}
```
If we're looking to update the server icons, should we also be changing the json file containing domain information?
(Since `DomainUtil.getDomains()` uses that file to pass domains to servers)

akashnimare:
I think in order to get everything right, we gotta save the index of the domain otherwise it would be difficult to keep track of all the indexes after dragging/reloading etc.

zulipbot:
[...]

PR #617, opened 2018-12-23
Make org tabs draggable in sidebar

kanishk98:
---

**What's this PR do?**
Updates `domains.json` on dragging server tabs and (intends to) reorder server tooltips along with that. 

Please refer to the discussion in #548.

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

**Current behaviour on Windows 10 (CSS animations by `sortablejs` not recorded by Windows Game Bar)**
zulip1001201902_01_27am

**Problems in current version**
On my computer, the server tab freezes for a bit before I can switch tabs again. 

**Other possible solutions that I considered**
1. We could just change the server names if we're able to also map the hoverable area of a server tab properly. 
2. We could swap the tooltips' nodes in the DOM (they're just sibling nodes, we should be able to `replaceChild()` as long as we maintain clones of the old tooltip node.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -43,6 +43,7 @@
     "node-json-db": "0.9.2",
     "request": "2.85.0",
     "semver": "5.4.1",
+    "sortablejs": "1.10.0-rc3",
     "wurl": "2.5.0"
   },
   "optionalDependencies": {
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -253,6 +253,7 @@ class ServerManagerView {
 			el.setAttribute('data-tab-id', index.toString());
 		});
 		this.servers = newServers;
+		DomainUtil.batchUpdateDomain(this.servers);
 		this.reloadView(false);
 	}
 
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -1,6 +1,7 @@
 'use strict';
 
 import { ipcRenderer, remote, clipboard, shell } from 'electron';
+import Sortable from 'sortablejs';
 import { feedbackHolder } from './feedback';
 
 import path = require('path');
@@ -79,6 +80,7 @@ class ServerManagerView {
 	$settingsButton: HTMLButtonElement;
 	$webviewsContainer: Element;
 	$backButton: HTMLButtonElement;
+	$orgsList: HTMLElement;
 	$dndButton: HTMLButtonElement;
 	$addServerTooltip: HTMLElement;
 	$reloadTooltip: HTMLElement;
@@ -90,8 +92,10 @@ class ServerManagerView {
 	$sidebar: Element;
 	$fullscreenPopup: Element;
 	$fullscreenEscapeKey: string;
+	$sortableList: any;
 	loading: AnyObject;
 	activeTabIndex: number;
+	servers: Domain[];
 	tabs: ServerOrFunctionalTab[];
 	functionalTabs: AnyObject;
 	tabIndex: number;
@@ -123,6 +127,7 @@ class ServerManagerView {
 		this.$dndTooltip = $actionsContainer.querySelector('#dnd-tooltip');
 
 		this.$sidebar = document.querySelector('#sidebar');
+		this.$orgsList = document.querySelector('#tabs-container');
 
 		this.$fullscreenPopup = document.querySelector('#fullscreen-popup');
 		this.$fullscreenEscapeKey = process.platform === 'darwin' ? '^⌘F' : 'F11';
@@ -131,6 +136,7 @@ class ServerManagerView {
 		this.loading = {};
 		this.activeTabIndex = -1;
 		this.tabs = [];
+		this.servers = DomainUtil.getDomains();
 		this.presetOrgs = [];
 		this.functionalTabs = {};
 		this.tabIndex = 0;
@@ -237,9 +243,26 @@ class ServerManagerView {
 		}
 	}
 
+	onEnd(): void {
+		const newServers: Domain[] = [];
+		const tabElements = document.querySelectorAll('#tabs-container .tab');
+		tabElements.forEach((el, index) => {
+			const oldIndex = Number(el.getAttribute('data-tab-id')) % this.servers.length;
+			newServers.push(this.servers[oldIndex]);
+			// TODO: Change this to read data from in-memory store or DomainUtil.
+			el.setAttribute('data-tab-id', index.toString());
+		});
+		this.servers = newServers;
+		this.reloadView(false);
+	}
+
 	initSidebar(): void {
 		const showSidebar = ConfigUtil.getConfigItem('showSidebar', true);
 		this.toggleSidebar(showSidebar);
+		this.$sortableList = Sortable.create(this.$orgsList, {
+			dataIdattr: 'data-sortable-id',
+			onEnd: this.onEnd.bind(this)
+		});
 	}
 
 	async queueDomain(domain: any): Promise<boolean> {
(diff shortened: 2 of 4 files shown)


abhigyank:
Don't know if its because its in WIP but the webviews (incll sidebar) hang for me if I try to drag and change the order,

kanishk98:
> Don't know if its because its in WIP but the webviews (incll sidebar) hang for me if I try to drag and change the order,

I don't know why that's happening. Can you explain any specific steps you took so that I can reproduce this?
I've updated the main PR description with a GIF showing sidebar behaviour on my end.

kanishk98:
I've found another problem that you may see if you clear app data, add a couple (or more) organisations and then drag them around:

image

This is caused because of line number 178 in `app/renderer/js/main.js`, as follows:
`domains.push(DomainUtil.getDomain(oldIndex));`

I'll need to get the actual old index from the `domains.json` file instead of just taking it from the DOM. Will get to that in some time.

abhigyank:
@kanishk98  attached a gif. You can also see the console error for `onHover` which probably is causing this. If you wamt I'd provide the full traceback or debug it for you.
peek 2018-12-24 22-03

kanishk98:
@abhigyank thanks for the gif. Looks like `this.state` has not been properly defined. I'll get to it.

kanishk98:
As for the problem I mentioned above (related to clearing app data, dragging new domains, and getting a crash), that's because the `data-tab-id` attribute only increments from the index of the previously added organisation. Can I get some ideas on where to start with this bug?

abhigyank:
Yes, that is the main issue here. From what I recall, Akash had previously suggested storing the index of the servers in domain.json itself. Then we'd be work all functions according to that index.

kanishk98:
> Yes, that is the main issue here. From what I recall, Akash had previously suggested storing the index of the servers in domain.json itself. Then we'd be work all functions according to that index.

I've used another solution, as follows:
```js
tabElements.forEach((el, index) => {
	const oldIndex = +Number(el.getAttribute('data-tab-id')) % tabElements.length;
```

This should work fine for the reset and drag crash.

kanishk98:
I'm unable to understand why the webviews freeze after dragging the server tabs. I've logged that they load and the index of the registered clicks is correct (and corresponds to the correct server). I'd really like some pointers on why this might happen.

kanishk98:
To keep maintainers updated on the status of this PR, I've attached a gif below. 

ezgif-5-52261d499369

I think this achieves the required functionality. Just working on a couple of things to make sure minimum code changes are needed. :)

akashnimare:
Looks great 👍

app/renderer/js/main.js
```diff
@@ -156,12 +156,12 @@ class ServerManagerView {
 				const domains = [];
 				const tabElements = document.querySelectorAll('#tabs-container .tab');
 				tabElements.forEach((el, index) => {
-					const oldIndex = +el.getAttribute('data-tab-id');
+					const oldIndex = +Number(el.getAttribute('data-tab-id'));
```
priyank-p: The use of `Number` is useless here, that's what the `+` does here.

kanishk98:
Hello, just checking in to ask if any other changes are required here. :)

Edit: I realized a little too late that running `git rebase` was the wrong idea here. The commit history is a little messed up on my end (duplicated commits) so if the maintainers give me the go ahead here, I'll clean up git history on my local machine and send in another PR. 
Sorry, this work is from quite a while ago. :)

app/package.json
```diff
@@ -40,7 +40,8 @@
     "node-json-db": "0.9.1",
     "request": "2.85.0",
     "semver": "5.4.1",
-    "wurl": "2.5.0"
+    "wurl": "2.5.0",
+    "sortablejs": "*"
```
akashnimare: @kanishk98 use the latest working version.

akashnimare (changes requested):

app/package.json
kanishk98: Fixed it and pushed the changes.

kanishk98:
I observed something right now while re-testing this PR: if the server tabs are moved around quickly and then the user chooses to disconnect an organisation from the context menu _before_ the changes made by the reordering of tabs are saved to disk, then the user will have accidentally deleted the wrong organisation. 

In addition, say we have 3 servers. Now the update to the `domain.json` file is made sequentially for all 3 of their new indices in the sidebar. At a certain instant of time, it is possible that the same organisation may be present at 2 indices according to the `domain.json` file (the changes are still getting saved to disk). If the user chooses to perform a hard reload during this time, then the orgs will be duplicated in the sidebar and it'll be pretty confusing behaviour to have that. 

I'm guessing that a solution to this problem would be to cache the server indices in memory instead of reading them from the disk, but I'd like to discuss this properly before I work on a solution.

kanishk98:
Updated PR to TypeScript.

app/renderer/main.html
```diff
@@ -15,7 +15,7 @@
             </div>
             <div id="sidebar" class="toggle-sidebar">
                 <div id="view-controls-container">
-                    <div id="tabs-container"></div>
+                    <div id="tabs-container" class="simple-list"></div>
```
priyank-p: We might want to pick a better name for this class here like `organizations-list` or `sortable-organizations-list` here.

app/renderer/js/main.ts
```diff
@@ -231,14 +233,37 @@ class ServerManagerView {
 		}
 	}
 
+	onEnd(): void {
+		const newServers: any[] | ServerOrFunctionalTab[] = [];
+		const tabMap: any = {};
```
priyank-p: You should create a interface here for this, it's just:
```suggestion
interface TabMap {
    [key: number]: number;
}
const tabMap: TabMap = {};
```

priyank-p: Also we should probably add a detailed comment here what `tabMap` stores and how it works. I used know the logic back almost a year ago and now looking at this I have no idea what this is which is the perfect reason to add a comment here.

akashnimare:
@kanishk98 please update the PR. @priyank-p can you help him with the implementation/logic here?

kanishk98:
> please update the PR. 

Fixed merge conflicts - I think the functionality was mostly alright. Would love a review anytime. :)

app/renderer/js/main.ts
```diff
@@ -123,6 +126,7 @@ class ServerManagerView {
 		this.$dndTooltip = $actionsContainer.querySelector('#dnd-tooltip');
 
 		this.$sidebar = document.querySelector('#sidebar');
+		this.$drag = document.querySelector('#simple-list');
```
priyank-p: I think we could be more verbose here i.e. `this.$sortableDrag`

app/renderer/js/main.ts
```diff
@@ -94,7 +94,7 @@ class ServerManagerView {
 	$fullscreenEscapeKey: string;
 	loading: AnyObject;
 	activeTabIndex: number;
-	servers: ServerOrFunctionalTab[];
+	servers: any[];
```
priyank-p: why is this switched back to `any`?

app/main/index.ts
```diff
@@ -366,7 +366,9 @@ app.on('ready', () => {
 
 app.on('before-quit', () => {
 	mainWindow.webContents.send('save-domains');
-	isQuitting = true;
+	setTimeout(() => {
+		isQuitting = true;
```
priyank-p: Hmm, how did you come to this conclusion here? This seems a workaround than a correct fix.

priyank-p (commented):
Left some comments! Looks like you forgot to add `sortablejs` as a dependency?

kanishk98:
@priyank-p thanks a lot for the review! I think a couple of those mistakes you listed above are because I accidentally force pushed a faulty commit at the time of review, but I'll have a look again and finalize this PR by tonight. :)

kanishk98:
I've updated this PR and used a batch-write function for replacing the old order of domains with the new order. This works well for me, I'd love for everyone to try it out on their end and check for missing servers/duplication (this is why we're reading the order of servers from memory for the most part).

akashnimare:
@kanishk98 can you fix the merge conflict? Also, I have noticed that we are not loading the server which is being dragged and I think we want to load that server instead of the new org on that index. Also, the order is not right -

https://cl.ly/8d7c41fba8dc

manavmehta:
@akashnimare I can't reproduce this
> Also, the order is not right 
>
The order is working well for me
@abhigyank can you have a look please ?

akashnimare:
I would suggest we should first look into the [DB migration](https://github.com/zulip/zulip-desktop/pull/810) and then work on this feature.

manavmehta:
@akashnimare sure. But is that related to this feature or that is a higher priority ?

abhigyank:
Due to no recent activity, closing in favour of #1059 .

Issue #551, opened 2018-08-31
bug: Flashing server icons in left sidebar


akashnimare:
When the app starts, there is flashing in the left sidebar in server tabs- this is because we first try to load the all the servers and then focus on the last activated server -

<img src="https://dzwonsemrish7.cloudfront.net/items/2R1V0i2b0i3P0X1g0U1U/Screen%20Recording%202018-08-31%20at%2005.31%20PM.gif?v=a2868147">

I'm imaging the correct flow for this would be -
* First load the last activated server
* Lazy load all the other servers

kanishk98:
@zulipbot claim

zulipbot:
Hello @kanishk98, you have been unassigned from this issue because you have not updated this issue or any referenced pull requests for over 14 days.

You can reclaim this issue or claim any other issue by commenting `@zulipbot claim` on that issue.

Thanks for your contributions, and hope to see you again soon!

kanishk98:
Have submitted a PR that I think solves the issue. Will improve solution according to suggestions by maintainers.

PR #621, opened 2019-01-02
Fix flashing icons of servers in sidebar

kanishk98:
**What's this PR do?**

Changes order in which servers are loaded and rendered in the sidebar. I'd like to stop doing the latter half by retaining the server tabs

#551 should provide the required context. 
I've changed the order of loading servers, but rendering items in the sidebar according to the old order is turning out to be a problem (my understanding of how `initServer()` works is faulty here. Does it not render `this.tabs` in `app/renderer/js/main.js` according to the order in which the elements are present in the array?)

This problem is similar to the one I'm currently facing in https://github.com/zulip/zulip-electron/pull/617. 

**Screenshots?** (according to https://github.com/zulip/zulip-electron/pull/621/commits/3ec20ca9f96a633d9229e746adf0144d0bb62d94, last updated on Jan 7)

zulip0701201905_05_54pm

The refresh you see has been forced by me. 

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -312,11 +312,24 @@ class ServerManagerView {
 		if (servers.length > 0) {
 			for (let i = 0; i < servers.length; i++) {
 				this.initServer(servers[i], i);
-				DomainUtil.updateSavedServer(servers[i].url, i);
-				this.activateTab(i);
 			}
 			// Open last active tab
-			this.activateTab(ConfigUtil.getConfigItem('lastActiveTab'));
+			let lastActiveTab = ConfigUtil.getConfigItem('lastActiveTab');
+			if (lastActiveTab >= servers.length) {
+				lastActiveTab = 0;
+			}
+			// checkDomain() and webview.load() for lastActiveTab before the others
+			DomainUtil.updateSavedServer(servers[lastActiveTab].url, lastActiveTab);
+			this.activateTab(lastActiveTab);
+			for (let i = 0; i < servers.length; i++) {
+				// after the lastActiveTab is activated, we load the others in the background
+				// without activating them, to prevent flashing of server icons
+				if (i === lastActiveTab) {
+					continue;
+				}
+				DomainUtil.updateSavedServer(servers[i].url, i);
+				this.tabs[i].webview.load();
+			}
 			// Remove focus from the settings icon at sidebar bottom
 			this.$settingsButton.classList.remove('active');
 		} else if (this.presetOrgs.length === 0) {


abhigyank:
> my understanding of how initServer() works is faulty here. Does it not render this.tabs in app/renderer/js/main.js according to the order in which the elements are present in the array?

It should call initServer in the order in which the domains.json saves the file.

I am not too sure if the changes you have done would lead us a good final solution. This part to implement is a bit tricky keeping in mind the user experience. You could try experimenting a bit on this. Understanding the function call orders with indexes should definitely help.

abhigyank:
@kanishk98 I don't think your code logic is correct. For me refreshing, causes me to lose a server. Clicking on 2nd tab opens the 3rd tab, the order of tabs changes on its own on refreshing.

kanishk98:
@abhigyank thanks for the feedback. I'll see what might've caused you to lose a server and change the click behaviour. 

As for changing the order of tabs, I'm loading the last active tab as the first server in the list on every refresh. I assumed that we'd want that to be the first tab. Is that what you mean by the incorrect order?

abhigyank:
> I assumed that we'd want that to be the first tab. Is that what you mean by the incorrect order?

I think we'd want that to the be the first tab to load but not necessarily the first tab in the list on the left sidebar. The order on the sidebar should remain same but only the loading order could vary. @akashnimare Feel free to weigh in here.

kanishk98:
@abhigyank What I understand from the previous comments is that we'd want to initialize the servers in the sidebar as per their original order, but the actual loading of their webviews happens only after the last active tab has been activated. If that's the case, I've changed my `initTabs()` to the following: 

```js
initTabs() {
		const servers = DomainUtil.getDomains();
		if (servers.length > 0) {
			for (let i = 0; i < servers.length; i++) {
				this.initServer(servers[i], i);
				DomainUtil.updateSavedServer(servers[i].url, i);
			}
			// activate last active tab and load its webview first
			const lastActiveTab = ConfigUtil.getConfigItem('lastActiveTab');
			this.activateTab(lastActiveTab);
			for (let i = 0; i < servers.length; i++) {
				if (i === lastActiveTab) {
					continue;
				}
				this.tabs[i].loadWebview();
			}
			// Remove focus from the settings icon at sidebar bottom
			this.$settingsButton.classList.remove('active');
		} else {
			this.openSettings('AddServer');
		}
	}
```
If you think I've got the correct idea now, I'll push this code.

kanishk98:
@akashnimare @abhigyank does this PR look good?

Edit: I'm guessing the work on this PR is done and waiting review. I'll remove the WIP tag; please let me know if there is something else that needs to be done.

abhigyank:
Need some time to review this :)

akashnimare:
@kanishk98 can you fix the merge conflicts? Also, the desired behaviour should be -

* Load the last active server first and focus the same in the left sidebar
* Load all the rest servers in the background + don't focus on the tabs

kanishk98:
Updated this PR's logic and also fixed the merge conflicts.
@akashnimare please let me know what you think whenever you're free.

akashnimare:
@kanishk98 can we make the code cleaner a bit, maybe something like this -
```js

let lastActiveTab = ConfigUtil.getConfigItem('lastActiveTab');

if (lastActiveTab >= servers.length) {
	lastActiveTab = 0;
}

// Following is the logic to handle which webview we need to load first
for (let i = 0; i < servers.length; i++) {
	if (i === lastActiveTab) {
		// load last active tab first
		DomainUtil.updateSavedServer(servers[lastActiveTab].url, lastActiveTab);
		this.activateTab(lastActiveTab);
		continue;
	}
	// load other webviews
	DomainUtil.updateSavedServer(servers[i].url, i);
	this.tabs[i].webview.load();
}

```

kanishk98:
@akashnimare wouldn't the above snippet load them in order anyway? We would activate only the last active server, yes, but don't we also want to load that one first?

akashnimare:
Merged. Thanks, @kanishk98.

PR #625, opened 2019-01-08
notification: Ensure backward compatibilty when using narrow.by_topic.

priyank-p:
We should ensure backward compatibilty when using narrow.by_topic since
the recent rename in the webapp of the function narrow.by_subject to
narrow.by_topic is only in master and not in any stable release yet it
could break the notification function for many people. Furthermore this ensure
general backward-compatibilty.

### DIFF
--- a/app/renderer/js/notification/helpers.js
+++ b/app/renderer/js/notification/helpers.js
@@ -141,7 +141,8 @@ function parseReply(reply) {
 
 function setupReply(id) {
 	const { narrow } = window;
-	narrow.by_topic(id, { trigger: 'notification' });
+	const narrowByTopic = narrow.by_topic || narrow.by_subject;
+	narrowByTopic(id, { trigger: 'notification' });
 }
 
 module.exports = {


akashnimare:
Good catch. Thanks, @priyank-p.

timabbott:
Interesting -- can you submit a PR to the webapp repo adding comments noting any API likes this that you're using?  I'd like us to know when we're breaking compatabiility when making changes there.

(Long term, it seems like we should do all things like this through the electron bridge, so that the API between the two is super explicit).

priyank-p:
The only three thing we use currently is the `narrow.by_topic`, `page_params.default_language` and `lightbox.open`.

I think I will move the `lightbox` call to use `electron_bridge`, and probably add a comment for `page_params.default_language` and the `narrow.by_topic` on webapp.

The thing we use `narrow.by_topic` for is to focus to the stream from which the notification was generated and insert the reply in the compose box and sent it. I think coming up with a better solution on webapp side is tricky or maybe there a way we can use the webapp API to achieve this?

timabbott:
Well, I think if we made it call `electron_bridge.narrow_by_topic` (if available), then we could preserve that API even if we changed the internal data structures in the webapp codebase, using a little shim.  Similarly with the others.

Issue #623, opened 2019-01-07
Allow muting organizations


xeruf:
In Discord, I can simply right click a server and then mute it. This will prevent any badges from showing up from this server, in the server overview and in the tray. This could have three modes, similarly to how Slack handles notifications, or even allow configuring it granularly:
* Show badge for all messages
* Only for DMs, Mentions & keywords
* Never show badge

Zulip discussion: https://chat.zulip.org/#narrow/stream/137-feedback/topic/Mute.20organizations

timabbott:
Thanks for the report; I agree this would be a useful feature to add to our new per-organization context menu.

kanishk98:
I'd love to work on this issue. I saw the discussion on the chat thread, and I think I should first start with the context menu and then add options for muting in the organizations page of preferences as well.

kanishk98:
@zulipbot claim

showell:
There's some discussion here about just doing this for the whole app (i.e. not desktop specific):

https://chat.zulip.org/#narrow/stream/101-design/topic/mute.20organization/near/683125

zulipbot:
Hello @kanishk98, you have been unassigned from this issue because you have not updated this issue or any referenced pull requests for over 14 days.

You can reclaim this issue or claim any other issue by commenting `@zulipbot claim` on that issue.

Thanks for your contributions, and hope to see you again soon!

akashnimare:
@Xerus2000 first two setting options have been added in the webapp. Checkout - https://github.com/zulip/zulip-desktop/pull/626

manavmehta:
Do we still have to work on the third setting of "Never Show Badge'' or has it been done on the webapp ?
@akashnimare @timabbott Seems to have been implemented already ?

abhigyank:
Feature added in Webapp - Settings-> Notifications -> Unread count summary (appears in desktop sidebar and browser tab) . 

Closing.

xeruf:
@abhigyank how about renaming this to "Unread count badge"? Even though I myself opened this issue back then, I could not find the setting for it until looking here again.

timabbott:
@xeruf can you bring this up in #desktop on chat.zulip.org?  Comments on closed issues are easy to miss.

xeruf:
Done: https://chat.zulip.org/#narrow/stream/6-frontend/topic/Rename.20unread.20count.20setting/near/1290697

PR #626, opened 2019-01-14
[WIP] Mute organization badge and server count

kanishk98:
---

**What's this PR do?**
Adds context menu and settings option to mute/unmute badge count of connected organizations. 

https://github.com/zulip/zulip-electron/issues/623 should provide the required context. Please also have a look at the chat thread referenced.

**Screenshots?**
image
image

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/components/webview.js
+++ b/app/renderer/js/components/webview.js
@@ -116,10 +116,15 @@ class WebView extends BaseComponent {
 		});
 	}
 
+	updateBadgeCount(count) {
+		this.badgeCount = count;
+		this.props.onTitleChange();
+	}
+
 	getBadgeCount(title) {
 		const mutedOrganizations = ConfigUtil.getConfigItem('mutedOrganizations');
 		const { url } = this.props;
-		if (!!mutedOrganizations[url]) {
+		if (mutedOrganizations[url]) {
 			return 0;
 		}
 		const messageCountInTitle = (/\(([0-9]+)\)/).exec(title);
--- a/app/renderer/js/components/webview.js
+++ b/app/renderer/js/components/webview.js
@@ -117,6 +117,11 @@ class WebView extends BaseComponent {
 	}
 
 	getBadgeCount(title) {
+		const mutedOrganizations = ConfigUtil.getConfigItem('mutedOrganizations');
+		const { url } = this.props;
+		if (!!mutedOrganizations[url]) {
+			return 0;
+		}
 		const messageCountInTitle = (/\(([0-9]+)\)/).exec(title);
 		return messageCountInTitle ? Number(messageCountInTitle[1]) : 0;
 	}
--- a/app/renderer/js/main.js
+++ b/app/renderer/js/main.js
@@ -513,7 +513,7 @@ class ServerManagerView {
 	}
 
 	addContextMenu($serverImg, index) {
-		let mutedOrganizations = ConfigUtil.getConfigItem('mutedOrganizations');
+		const mutedOrganizations = ConfigUtil.getConfigItem('mutedOrganizations');
 		const url = DomainUtil.getDomain(index).url;
 		$serverImg.addEventListener('contextmenu', e => {
 			e.preventDefault();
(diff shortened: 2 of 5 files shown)


app/renderer/js/components/webview.js
```diff
@@ -117,6 +117,11 @@ class WebView extends BaseComponent {
 	}
 
 	getBadgeCount(title) {
+		const mutedOrganizations = ConfigUtil.getConfigItem('mutedOrganizations');
+		const { url } = this.props;
+		if (!!mutedOrganizations[url]) {
```
abhigyank: Jus out of curiosity, any reason to use `!!mutedOrganizations[url]` instead of simply `mutedOrganizations[url]` ?

abhigyank (commented):
The feature works as expected and required. Good work on that.
A couple of suggestions with implementation though - 
* Instead of storing whether org is muted in configutil, I think i'd be better to store that as a property of the org in domain-util.json.
* I feel Muting/Unmuting a realm causing a refresh could be inconvenience to the user. I feel we should be able to do maybe to this without refreshing.
* Not sure about the design of the additions in the organisations page. Maybe @akashnimare  would weigh in on that.

app/renderer/js/components/webview.js
kanishk98: Not really, just the way I got used to. I'll change this along with the Travis fixes and squash the commits.

kanishk98:
> Instead of storing whether org is muted in configutil, I think i'd be better to store that as a property of the org in domain-util.json.

I attempted that initially by adding a `notify` flag to the `serverConf` object, but this approach didn't work out for `zulipdev.org` servers. The flag would get added and then removed from the file. This was before you helped me with the console logs; I'll let you know how this goes.

> I feel Muting/Unmuting a realm causing a refresh could be inconvenience to the user. I feel we should be able to do maybe to this without refreshing.

I agree. I'll get to work. :)

kanishk98:
@abhigyank the latest commit addresses the refresh problem, but behaviour is undesirable in one case - if the user is on the org preferences page and then mutes an org using the context menu, it does not trigger an update of the DOM element in the pane. 
My initial intuition is that if we can keep track of where the user is in the app, then we can figure out a way to cause the update in the preferences page if the user mutes the organization using the context menu.

abhigyank:
@kanishk98 the unread count now doesn't change for me in any case. Now since we are not reloading the app, you would need explicitly remove the badge count of the div `server-tab-badge` of that icon. 

> if the user is on the org preferences page and then mutes an org using the context menu, it does not trigger an update of the DOM element in the pane.

Not sure how we can tackle this but I think we can make use of IPCRenderers 'forward-message' feature  here.

kanishk98:
Ah, sorry. Should've checked thoroughly.

> Not sure how we can tackle this but I think we can make use of IPCRenderers 'forward-message' feature here.

Thanks! I'll check that out.

showell:
Some relevant discussion is here:

https://chat.zulip.org/#narrow/stream/101-design/topic/mute.20organization/near/683125

akashnimare:
@rishig anything else we should be doing here functionality wise?

rishig:
What does the setting currently do?
Also, I think the conclusion was that this should be implemented in the webapp?

app/renderer/js/main.js
```diff
@@ -530,6 +533,29 @@ class ServerManagerView {
 							}
 						});
 					}
+				},
+				{
+					label: (mutedOrganizations[url] ? 'Unmute' : 'Mute') + ' organization',
```
rishig, line 538: Note that stuff like this is not going to work with translations

kanishk98:
> What does the setting currently do?

The setting hides the number of unread messages in the sidebar and the tray icon for all muted organizations. 

> Also, I think the conclusion was that this should be implemented in the webapp?

If you see [this thread](https://chat.zulip.org/#narrow/stream/137-feedback/topic/Mute.20organizations), you'll see that the last few messages indicate that the feature is required in the desktop app too. One of the suggestions from the community (see https://github.com/zulip/zulip-electron/issues/623#issuecomment-454795980) is to extend this functionality to other environments, but this is certainly needed on the desktop. :)

app/renderer/js/main.js
kanishk98: I'm sorry, I didn't think about that. I'll need to have a look at how translations are currently handled. Will get to it.

rishig:
ah, I guess we should have been more explicit. Given that it needs to be reflected on all clients, the setting should be stored on the UserProfile object on the server, and controlled by something in the webapp's settings page. That single setting will control how we set unread counts on desktop/mobile/terminal/web.

akashnimare:
@kanishk98 any update on this? https://github.com/zulip/zulip/pull/12677 should be helpful I guess. @rishig would it make sense to add an option like the following to mute an org in the desktop app -

image

rishig:
yeah, for sure; maybe called "None".
Though I imagine the current setting solved most of the problem for most people.

kanishk98:
I'm confused about the discussion here - given that people can now control their notifications across all clients from the web app and we read notifications from the favicon updates, I thought the conclusion here was to move such settings to the web app and continue as is with the desktop app.

kanishk98:
If we're planning to also include a similar setting on the desktop app, how does that add more to what's on the web app right now?

akashnimare:
> yeah, for sure; maybe called "None".

@kanishk98 if we add this in the webapp then we can re-use it in the desktop app to add muting org feature.

rishig:
I also agree with Kanishk's intuition; I think we likely don't need a separate desktop app feature for this. 
The benefit of having settings in the webapp is that those settings will translate to mobile as well.

PR #627, opened 2019-01-14
Changed saveServerIcon() to not use request() on defaultIconUrl

Swapnilr1:
---

**What's this PR do?**
Fixes https://github.com/zulip/zulip-electron/issues/533

**Screenshots?**

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -269,6 +269,10 @@ app.on('ready', () => {
 
 	ipcMain.on('update-menu', (event, props) => {
 		appMenu.setMenu(props);
+		const activeTab = props.tabs[props.activeTabIndex];
+		if (activeTab) {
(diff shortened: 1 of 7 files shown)


abhigyank (approved):
@akashnimare This looks good to me and works as expected, could be merged.
Though the issue itself is not much of a problem to us now cause we fallback to the initial character of the realm, but this would now correctly prevent the error from propping up.

Swapnilr1:
I made a mistake while squashing the commits together, so this is 2 commits instead of 1. I hope this is fine. @akashnimare Could you review this?

Swapnilr1:
Please discard this. I will open up a new PR.

PR #631, opened 2019-01-20
[WIP] Do not check for zulip.ogg if server_settings returns error 404

### LINKED ISSUE #573: Stop checking for zulip.ogg if error from the server is a 400 "BAD_REQUEST" error
Right now, there's a really bad user experience if you typo the hostname on zulipchat.com, see https://chat.zulip.org/#narrow/stream/16-desktop/subject/adding.20org.20that.20doesn't.20exist/near/646444.

From https://github.com/zulip/zulip-electron/pull/236#issuecomment-414450918, there's a proposed solution that should work great:

```
```

And that's very different from the 404; we should only check for `zulip.ogg` if we get a `404` error back from `/server_settings`.

Swapnilr1:
---

**What's this PR do?**
Fixes #573 

**Screenshots?**

  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

I am not currently sure how to handle `*.zulipdev.org`. Looking into it.

### DIFF
--- a/app/renderer/js/utils/domain-util.js
+++ b/app/renderer/js/utils/domain-util.js
@@ -277,28 +277,33 @@ class DomainUtil {
 		return new Promise(resolve => {
 			const filePath = this.generateFilePath(url);
 			const file = fs.createWriteStream(filePath);
-			try {
-				request(serverIconOptions).on('response', response => {
-					response.on('error', err => {
+			if (serverIconOptions.url === defaultIconUrl) {
+				logger.log('Could not get server icon.');
+				resolve(defaultIconUrl);
+			} else {
+				try {
+					request(serverIconOptions).on('response', response => {
+						response.on('error', err => {
+							logger.log('Could not get server icon.');
+							logger.log(err);
+							logger.reportSentry(err);
+							resolve(defaultIconUrl);
+						});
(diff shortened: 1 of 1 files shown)


abhigyank:
> I am not currently sure how to handle *.zulipdev.org. Looking into it.

We whitelist `'zulipdev.org'` since that is a dev server only Zulip developers use and can be prone to different issues. So we need not do checks on it.

kanishk98:
@Swapnilr1 I'm guessing you're relatively new to the Zulip repo, so I'd like to give you some suggestions that the maintainers gave me when I first started contributing; please keep the git history clean. 
For example, I see that you're working on 2 PRs, and it would certainly make the maintainers' lives easier if you used different branches to test both of them (none of them being `master`). 

I'm also new here, and I also faced problems with squashing commits properly earlier, and I'd say that having a graphical Git client goes a long way towards helping one with this. [GitKraken](https://www.gitkraken.com/git-client) is my preferred tool for this specific purpose. They've got a [nice tutorial](https://support.gitkraken.com/working-with-commits/squash/) for this too. 

I hope that helps. Good luck! :)

Swapnilr1:
@kanishk98 Thanks a lot for the advice.

PR #633, opened 2019-01-21
certificate: Use path.sep for path separator to support Windows.

abhigyank:
Earlier path separator was hard-code to "/" which failed in Windows. Now, [`path.sep`](https://nodejs.org/api/path.html#path_path_sep) return `/` for POSIX and `\` for Windows.


akashnimare:
Thanks, @abhigyank :)

Issue #600, opened 2018-11-27
Add /find-accounts link on new server page


akashnimare:
More discussion at - https://chat.zulip.org/#narrow/stream/16-desktop/topic/Find.20accounts

kanishk98:
@akashnimare how would you suggest going about this - a `<webview>` that loads in another window or a core Electron solution that just calls the relevant APIs and displays results accordingly? The second is a little more work (I'm not very familiar with the web side of things on Zulip yet), but I'd personally prefer using that.

kanishk98:
@zulipbot claim

zulipbot:
Hello @kanishk98, you have been unassigned from this issue because you have not updated this issue or any referenced pull requests for over 14 days.

You can reclaim this issue or claim any other issue by commenting `@zulipbot claim` on that issue.

Thanks for your contributions, and hope to see you again soon!

kanishk98:
Still working on this issue (waiting for suggestions on implementation, see [this text](https://chat.zulip.org/#narrow/stream/16-desktop/topic/Find.20accounts/near/691499).

PR #638, opened 2019-01-25
Add option to find accounts by email

kanishk98:
**What's this PR do?**

Adds option to find Zulip accounts by their associated email. 

The functionality implemented uses the same API as [the current /find-accounts page](https://zulipchat.com/accounts/find). Relevant discussion may be found at https://github.com/zulip/zulip-electron/issues/600. 

**Screenshots?**

image

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/pages/preference/find-accounts.js
+++ b/app/renderer/js/pages/preference/find-accounts.js
@@ -58,6 +58,14 @@ class FindAccounts extends BaseComponent {
 				this.findAccounts(this.$serverUrlField.value);
 			}
 		});
+
+		this.$serverUrlField.addEventListener('input', () => {
+			if (this.$serverUrlField.value) {
+				this.$serverUrlField.classList.remove('invalid-input-value');
+			} else {
+				this.$serverUrlField.classList.add('invalid-input-value');
+			}
+		});
 	}
 }
 
--- a/app/renderer/js/pages/preference/find-accounts.js
+++ b/app/renderer/js/pages/preference/find-accounts.js
@@ -1,6 +1,6 @@
 'use-strict';
 
-const request = require('request');
+const { ipcRenderer } = require('electron');
 
 const BaseComponent = require(__dirname + '/../../components/base.js');
 
@@ -12,70 +12,27 @@ class FindAccounts extends BaseComponent {
 
 	template() {
 		return `
-			<div class="settings-card find-accounts-card">
-				<div class="email-input">
-					<div>Email address(es)</div>
-					<input class="setting-input-value" autofocus placeholder="Add up to 10 comma-separated email addresses"/>
-				</div>
-				<div class="email-input">
-					<button class="green sea w-150" id="find-accounts-button">Find accounts</button>
-				</div>
-			</div>
-		`;
-	}
-
-	confirmation() {
-		return `
-			<div class="sub-title">
-				<div>Emails sent! You will only receive emails at addresses associated with Zulip organizations.</div>
-			</div>
-		`;
-	}
-
-	error() {
-		return `
-			<div class="sub-title">
-				<div>Sorry, looks like something went wrong. Try again?</div>
-			</div>
+		<div>
+			<button class="green sea w-150" id="find-accounts-button">Find accounts</button>
+		</div>
 		`;
 	}
 
 	init() {
 		this.$findAccounts = this.generateNodeFromTemplate(this.template());
-		this.$confirmation = this.generateNodeFromTemplate(this.confirmation());
-		this.$error = this.generateNodeFromTemplate(this.error());
+		console.log(this.$findAccounts);
 		this.props.$root.appendChild(this.$findAccounts);
 		this.findAccountsButton = this.$findAccounts.querySelector('#find-accounts-button');
-		this.emailList = this.$findAccounts.querySelector('input.setting-input-value');
-		this.initListeners();
+		this.initListener();
 	}
 
 	findAccounts() {
-		const emails = this.emailList.value;
-		if (emails && emails !== '') {
-			const url = 'https://zulipchat.com/accounts/find?emails=' + emails;
-			request({ url }, error => {
-				if (error) {
-					this.props.$root.appendChild(this.$error);
-				} else {
-					this.props.$root.appendChild(this.$confirmation);
-					this.emailList.value = '';
-				}
-			});
-		}
+		ipcRenderer.send('find-accounts', 'https://zulipchat.com/accounts/find');
 	}
 
-	initListeners() {
+	initListener() {
 		this.findAccountsButton.addEventListener('click', () => {
 			this.findAccounts();
-		});
-
-		this.emailList.addEventListener('keypress', event => {
-			const EnterkeyCode = event.keyCode;
-
-			if (EnterkeyCode === 13) {
-				this.findAccounts();
-			}
 		});
 	}
 }
--- a/app/renderer/js/pages/preference/find-accounts.js
+++ b/app/renderer/js/pages/preference/find-accounts.js
@@ -1,7 +1,6 @@
 'use-strict';
 
-const { dialog } = require('electron').remote;
-const path = require('path');
+const request = require('request');
 
 const BaseComponent = require(__dirname + '/../../components/base.js');
 
@@ -33,17 +32,37 @@ class FindAccounts extends BaseComponent {
 		`;
 	}
 
+	error() {
+		return `
+			<div class="sub-title">
+				<div>Sorry, looks like something went wrong. Try again?</div>
+			</div>
+		`;
+	}
+
 	init() {
 		this.$findAccounts = this.generateNodeFromTemplate(this.template());
 		this.$confirmation = this.generateNodeFromTemplate(this.confirmation());
+		this.$error = this.generateNodeFromTemplate(this.error());
 		this.props.$root.appendChild(this.$findAccounts);
-		this.findAccountsButton = this.$findAccounts.querySelector('#add-accounts-button');
-		this.emailList = this.$findAccounts.querySelectorAll('input.setting-input-value')[0];
+		this.findAccountsButton = this.$findAccounts.querySelector('#find-accounts-button');
+		this.emailList = this.$findAccounts.querySelector('input.setting-input-value');
 		this.initListeners();
 	}
 
 	findAccounts() {
-		// TODO: Write business logic to find accounts
+		const emails = this.emailList.value;
+		if (emails && emails !== '') {
+			const url = 'https://zulipchat.com/accounts/find?emails=' + emails;
+			request({ url }, error => {
+				if (error) {
+					this.props.$root.appendChild(this.$error);
+				} else {
+					this.props.$root.appendChild(this.$confirmation);
+					this.emailList.value = '';
+				}
+			});
+		}
 	}
 
 	initListeners() {
--- a/app/renderer/js/pages/preference/find-accounts.js
+++ b/app/renderer/js/pages/preference/find-accounts.js
@@ -1,6 +1,6 @@
 'use-strict';
 
-const { ipcRenderer } = require('electron');
+const { shell } = require('electron');
 
 const BaseComponent = require(__dirname + '/../../components/base.js');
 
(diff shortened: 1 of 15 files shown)


rishig:
This should probably have a second field ("Server") that has zulipchat.com pre-filled so that it's clear it's only finding accounts from that one server.

app/renderer/js/pages/preference/find-accounts.js
```diff
@@ -0,0 +1,83 @@
...
+		this.props.$root.appendChild(this.$findAccounts);
+		this.findAccountsButton = this.$findAccounts.querySelector('#find-accounts-button');
+		this.emailList = this.$findAccounts.querySelector('input.setting-input-value');
+		this.initListeners();
+	}
+
+	findAccounts() {
+		const emails = this.emailList.value;
+		if (emails && emails !== '') {
+			const url = 'https://zulipchat.com/accounts/find?emails=' + emails;
```
priyank-p: @kanishk98 the domain in the url here should not be hardcoded to `zulipchat.com` here, as @rishig said you should add a field that takes the server url and use that instead.

In case you didn't know, this work for any self-hosted zulip org, but the main benefit of this is mostly on `.zulipchat.com` so we should have that pre-filled. You should also use js to put the cursor in front of `.zulipchat.com`.

app/renderer/js/pages/preference/find-accounts.js
```diff
@@ -0,0 +1,83 @@
...
+					<input class="setting-input-value" autofocus placeholder="Add up to 10 comma-separated email addresses"/>
+				</div>
+				<div class="email-input">
+					<button class="green sea w-150" id="find-accounts-button">Find accounts</button>
+				</div>
+			</div>
+		`;
+	}
+
+	confirmation() {
```
priyank-p: How about de-duplicating confirmation and error function here, since they used same html code but diffrent messages into a new function that return the html with message like:

```javascript
	getMessageElement(msg) {
		return `
			<div class="sub-title">
				<div>${msg}</div>
			</div>
		`;
	}

	confirmation() {
		const confirmationMsg = 'Emails sent! You will only receive ' +
		'emails at addresses associated with Zulip organizations.'; 
	  	return this.getMessageElement(confirmationMsg);
	}
	
	error() {
		const errorMsg = 'Sorry, looks like something went wrong. ' +
		'Try again?';
		return this.getMessageElement(errorMsg);
	}
```

app/renderer/js/pages/preference/find-accounts.js
```diff
@@ -0,0 +1,83 @@
...
+			<div class="sub-title">
+				<div>Sorry, looks like something went wrong. Try again?</div>
+			</div>
+		`;
+	}
+
+	init() {
+		this.$findAccounts = this.generateNodeFromTemplate(this.template());
+		this.$confirmation = this.generateNodeFromTemplate(this.confirmation());
+		this.$error = this.generateNodeFromTemplate(this.error());
```
priyank-p: I think you should only create the elements when needed, so you should be generating them when the needed below in the code path. Most of the time a user is not going to be using this and when they don't it's very unlikely both of the elements would be needed.

app/renderer/js/pages/preference/find-accounts.js
```diff
@@ -0,0 +1,83 @@
...
+	error() {
+		return `
+			<div class="sub-title">
+				<div>Sorry, looks like something went wrong. Try again?</div>
+			</div>
+		`;
+	}
+
+	init() {
+		this.$findAccounts = this.generateNodeFromTemplate(this.template());
```
priyank-p, line 28: Can't we just move this to html file and query it here? I think it would be cleaner.

priyank-p:
@kanishk98 It looks like this doesn't work, you should be doing a `POST` request with emails not a `GET`. I did that but the problem now is that the request get `403` error code because we don't include `crsftoken`. I don't know what do to do about this.

kanishk98:
> @kanishk98 It looks like this doesn't work, you should be doing a `POST` request with emails not a `GET`. I did that but the problem now is that the request get `403` error code because we don't include `crsftoken`. I don't know what do to do about this.

@priyank-p same for me here, and this is after I tested this functionality by entering the URL into the browser and also by entering emails into the app itself. I'm not sure why it worked the first few times around. 

I'll dig around the main zulip repo and ask my questions there. Sorry for the trouble. :(

kanishk98:
@priyank-p the simplest solution here would be to fire up a new `BrowserWindow` that loads the zulipchat.com URL, like so:

image

As you can see, the `csrftoken` is present here. If you think this method shouldn't be used and we should stick to the original implementation, we _could_ consider downloading the contents of the page (I checked the Zulip repo and the `csrftoken` seems to be coming from an AJAX request, so not sure exactly how I would solve that, but I'd be happy to dig around if, like I said above, you're not in favour of a separate window.

priyank-p:
@kanishk98 Let's start a thread on [#desktop stream](https://chat.zulip.org/#narrow/stream/16-desktop) to discuss this further; I don't know the better model going forward with this.

kanishk98:
> @kanishk98 Let's start a thread on [#desktop stream](https://chat.zulip.org/#narrow/stream/16-desktop) to discuss this further; I don't know the better model going forward with this.

There was already a thread on the stream, I've sent in a [text requesting help](https://chat.zulip.org/#narrow/stream/16-desktop/topic/Find.20accounts/near/691499).

kanishk98:
I've followed [the discussion](https://chat.zulip.org/#narrow/stream/16-desktop/topic/Find.20accounts/near/696270) at czo and modified the PR to simply open the /find/account link in the user's default browser. Hope that solves this issue. 

P.S. Apologies for the repeated forced updates. `$ npm run travis` doesn't work on Windows and I typically rely on `$ npm run test` alone to give me results. Didn't work this time as expected.

priyank-p:
> $ npm run travis doesn't work on Windows 

@kanishk98 `npm run travis` is not meant to be run locally it's for Travis CI so that `npm run test-e2e` can work. Since we have  e2e tests disabled in CI temporarily you should just only need to rely on `npm test`, the linting, to pass.

kanishk98:
@priyank-p Yes, I understand that, but I did not catch some newline errors with `npm run test` that turned up in Travis. That's the reason for my confusion. Is that a known problem as well?
If so, I can try looking into it.

priyank-p:
Hmm, that's weird I use windows, along with macos, to develop and I have never ran into such issue so you should defiently try to debug this. This most proabaly is some cross-platform issue that you will need to report to upstream `xo` or `eslint` project to get this solved.

app/renderer/js/pages/preference/find-accounts.js
```diff
@@ -0,0 +1,40 @@
...
+		return `
+		<div>
+			<button class="green sea w-150" id="find-accounts-button">Find accounts</button>
+		</div>
+		`;
+	}
+
+	init() {
+		this.$findAccounts = this.generateNodeFromTemplate(this.template());
+		console.log(this.$findAccounts);
```
akashnimare: @kanishk98 can you remove this?

app/renderer/js/pages/preference/find-accounts.js
akashnimare: @kanishk98 ^^

akashnimare (changes requested):

kanishk98:
@priyank-p @akashnimare I've made the necessary changes. Also saw that the margin of the find-accounts button was a little off the first time around, so I added some to the top. 
(Mentioning this because I squashed unnecessary commits and force-pushed, so the entire conversation wasn't visible)

akashnimare (approved):

akashnimare:
This looks good to me. @priyank-p can we improve it (or clean up the solution a bit)?

abhigyank:
Should we include the information somewhere that works only for zulipchat.com realms?

akashnimare:
I doubt if it's necessary since the find-account page is already pretty clear and has some info related to this.

rishig:
I thought the idea was the user enters their server, and it takes them to that server's find-account page?

akashnimare:
> I thought the idea was the user enters their server, and it takes them to that server's find-account page

Yeah, I missed this part. @kanishk98 can you add this also?

kanishk98:
Please let me know if the new changes work. I know Priyank had suggested that I move the `template()` code in `find-accounts.js` to the HTML in `connected-org-section.js`, but I felt that my way of doing it would maintain consistency (since `add-certificate` is also handled in the same way). Please let me know if the code structure is alright too.

app/renderer/js/pages/preference/find-accounts.js
```diff
@@ -0,0 +1,71 @@
...
+		this.$findAccounts = this.generateNodeFromTemplate(this.template());
+		this.props.$root.appendChild(this.$findAccounts);
+		this.$findAccountsButton = this.$findAccounts.querySelector('#find-accounts-button');
+		this.$serverUrlField = this.$findAccounts.querySelectorAll('input.setting-input-value')[0];
+		this.initListeners();
+	}
+
+	findAccounts(url) {
+		if (!url) {
+			return;
```
priyank-p, line 37: We should probably do something here like make the field border red or something?

app/renderer/js/pages/preference/find-accounts.js
```diff
@@ -0,0 +1,71 @@
...
+		super();
+		this.props = props;
+	}
+
+	template() {
+		return `
+			<div class="settings-card certificate-card">
+				<div class="certificate-input">
+					<div>Organization URL</div>
+					<input class="setting-input-value" value=".zulipchat.com"/>
```
priyank-p: Don't we just want the server input to have `zulipchat.com` instead of `.zulipchat.com`? cc @akashnimare @rishig.

priyank-p (commented):
@kanishk98 This looks much better than the first version, Thank for working on this. I left two comment on this.

app/renderer/js/pages/preference/find-accounts.js
rishig: yeah

kanishk98:
Pushed the suggested changes.

kanishk98:
> We should probably do something here like make the field border red or something?

So I improved a little on Priyank's suggestion and just check if the field is empty before the user hits enter and make the field border red then. It works as below:

4682c358c6ddf0b14c082f7557204a79 (1)

I think we've handled the issue now. Let me know if there's anything else that should be done.

app/renderer/js/pages/preference/find-accounts.js
```diff
@@ -0,0 +1,79 @@
...
+
+	findAccounts(url) {
+		if (!url) {
+			return;
+		}
+		if (!url.includes(':')) {
+			url = 'https://' + url;
+		}
+		ipcRenderer.send('find-accounts', url + '/accounts/find');
+		ipcRenderer.on('find-accounts-reply', (event, result) => {
```
akashnimare: What's the use of this?

akashnimare (changes requested):

app/renderer/js/pages/preference/find-accounts.js
kanishk98: In `index.js`, we check if the app was able to open the URL in the browser. This happens in the following way:

```javascript
ipcMain.on('find-accounts', (event, url) => {
	const isBrowserOpened = electron.shell.openExternal(url);
	event.sender.send('find-accounts-reply', isBrowserOpened);
});
```
This code block is to notify the user if the URL entered was opened by the browser or not.

app/renderer/js/pages/preference/find-accounts.js
```diff
@@ -0,0 +1,79 @@
...
+		this.$findAccountsButton = this.$findAccounts.querySelector('#find-accounts-button');
+		this.$serverUrlField = this.$findAccounts.querySelectorAll('input.setting-input-value')[0];
+		this.initListeners();
+	}
+
+	findAccounts(url) {
+		if (!url) {
+			return;
+		}
+		if (!url.includes(':')) {
```
abhigyank: Checking for `:` is not a good logic to determine the presence of prefix. Check the url starts with `http`. `:` can be included while specifying ports which is usually a case in self-hosted servers so that is not a good check.

app/renderer/js/pages/preference/find-accounts.js
abhigyank: `'Check the organization URL and try again.'` is not a relevant error message here. 
`isBrowserOpened` would be false only if there wasn't an application was available to open the URL. I think we should avoid this altogether.

app/renderer/js/pages/preference/find-accounts.js
```diff
@@ -0,0 +1,79 @@
...
+		super();
+		this.props = props;
+	}
+
+	template() {
+		return `
+			<div class="settings-card certificate-card">
+				<div class="certificate-input">
+					<div>Organization URL</div>
+					<input class="setting-input-value" value="zulipchat.com"/>
```
abhigyank, line 18: Do we want this to be a value or placeholder? 
cc @rishig

abhigyank (changes requested):

app/renderer/js/pages/preference/find-accounts.js
kanishk98: From Priyank's message [here](https://github.com/zulip/zulip-electron/pull/638#discussion_r251202738), we wanted this to be pre-filled with .zulipchat.com as a value earlier, but we've since then decided to change this to zulipchat.com [here](https://github.com/zulip/zulip-electron/pull/638#discussion_r272813020). Either way, we want this to be a pre-filled value.

app/renderer/js/pages/preference/find-accounts.js
kanishk98: Hmm, I didn't think of that. Thanks, I'll change it.

kanishk98:
@abhigyank thanks for the review - I've made the changes you suggested. Please let me know if anything else needs to be done.

akashnimare (approved):

app/renderer/js/pages/preference/find-accounts.js
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	findAccounts(url) {
+		if (!url) {
+			return;
+		}
+		if (!url.startsWith('http')) {
+			url = 'https://' + url;
+		}
+		ipcRenderer.send('find-accounts', url + '/accounts/find');
```
abhigyank: Any need to do this in the main process? We can `openExternal` at this line too.

abhigyank:
Other than https://github.com/zulip/zulip-electron/pull/638#discussion_r279800312,  LGTM.

app/renderer/js/pages/preference/find-accounts.js
kanishk98: Not really, I don't remember why I didn't do that. 
No problem, I'll make the changes right now.

kanishk98:
Included [this change](https://github.com/zulip/zulip-electron/pull/638#discussion_r279800312) as a fixup to [this commit](https://github.com/zulip/zulip-electron/pull/638/commits/832df6fff19c35c2a971a2843e0850a9f8e4fd53).

kanishk98:
This PR has shifted direction significantly from its original idea. I think it's far simpler to send in a cleaner PR rather than continue working with this. 
I'll close this for now and send in another PR tonight with only the approved changes. 
@akashnimare let me know if you have any problems.

PR #639, opened 2019-01-25
webview: Work around buggy focus switching in Electron 3.0.10

### LINKED ISSUE #634: Incorrect Focus after Switching Realms
Running on the latest master. 
When I switch between Realms, I have to click to change focus to the new realm.  To reproduce, 

- add two (or more) realms to Zulip electron
- open compose box in the first realms
- switch to the second realm using the short cut key (Ctrl + 2)
- try using navigation shortcuts - n, j, k, etc. Notice that things don't work! 
- Click on the "page" to bring it into focus. Try the navigation shortcuts now. 
- switch back to the first realm (Ctrl + 1)
- Notice all the previously tried shortcuts typed out in the compose box.

punchagan:
This commit works around a couple of bugs, which seem to be upstream bugs in the
current version of Electron.

First, `webContents.isFocused()` seems to be true even if we just switched to
a tab, and are trying to set it as the focused element. To workaround this, we
check if the `webview` element is the same as `document.activeElement`.

Also, as per https://github.com/electron/electron/issues/15718, it looks like
`blur` needs to be called on the currently active element, before switching
focus to another element i.e., calling `focus` on it.

Closes #634

### DIFF
--- a/app/renderer/js/components/webview.js
+++ b/app/renderer/js/components/webview.js
@@ -170,7 +170,13 @@ class WebView extends BaseComponent {
 	focus() {
 		// focus Webview and it's contents when Window regain focus.
 		const webContents = this.$el.getWebContents();
-		if (webContents && !webContents.isFocused()) {
+		// HACK: webContents.isFocused() seems to be true even without the element
+		// being in focus. So, we check against `document.activeElement`.
+		if (webContents && this.$el !== document.activeElement) {
+			// HACK: Looks like blur needs to be called on the previously focused
+			// element to transfer focus correctly, in Electron v3.0.10
+			// See https://github.com/electron/electron/issues/15718
+			document.activeElement.blur();
 			this.$el.focus();
 			webContents.focus();
 		}


app/renderer/js/components/webview.js
```diff
@@ -165,7 +165,13 @@ class WebView extends BaseComponent {
 	focus() {
 		// focus Webview and it's contents when Window regain focus.
 		const webContents = this.$el.getWebContents();
-		if (webContents && !webContents.isFocused()) {
+		// HACK: webContents.isFocused() seems to be true even without the element
+		// being in focus. So, we check against `document.activeElement`.
+		if (!(this.$el === document.activeElement)) {
```
vsvipul: Why not use simple !== condition ? Also, we need an if condition for webContents here, because we are focusing on it later on. Rest of it seems fine. Solves the issue. LGTM ! :)

vsvipul (approved):

app/renderer/js/components/webview.js
punchagan: Thanks @vsvipul  for the review. I've pushed changes addressing your comments

akashnimare:
Merged. Thanks, @punchagan. Excellent commit message 🙇

punchagan:
Thanks, @akashnimare! :grinning:

PR #598, opened 2018-11-19
[WIP] Fix online check if using proxy

superbiche:
**What's this PR do?**
This PR tries to mitigate (re)connection issues experienced by users behind a proxy as raised in #591 and discussed [here](https://chat.zulip.org/#narrow/stream/16-desktop/topic/no.20internet).  

If `is-online` returns `true`, we're online
If `is-online` returns `false`, no proxy enabled, we're offline
If `is-online` returns `false` **and we use a proxy**, we:
- get the list of domains  
- prepare a properly setup `request` with `ca` and `proxy` options for each domain to fetch the `favicon.ico`
- run a `Promise.race()` against all the requests, ending whenever one domain's favicon has been reached (**means we're kinda online regarding our Zulip instances**)

**Any background context you want to provide?**
The package used to check every 1.5s if we're online/offline, `is-online`, relies on the `got` package which is not compatible with proxies. 
Given that `got` is not compatible with proxy connections, and that there's so many ways a test could work in a given setup and not on another, the best way seems to test for Zulip instances as this is the most relevant.

**Screenshots?**

* Behavior with online/offline/online : https://d.pr/i/6I7nCb  
* Behavior with offline/online : https://d.pr/i/NVWVSG

**You have tested this PR on:**
macOS Mojave 10.14.1

**Important notes:**

I've deliberately commented some code to help test this implementation without a proxy and reach the favicon test code. Would be really amazing if someone could try it with a proxy!

**This is very far from being either a perfect solution but, for me, it's much more valuable being able to reach a Zulip domain than Google or OpenDNS, as what we want to give the user is an insight about whether he can use a Zulip domain and nothing else.**  
Maybe cleaner solutions exist that would work in proxy setups, but when even reaching Google or OpenDNS fails in some very specific setups, and if you can't reach anything else than your Zulip domains, you should still be able to use Zulip.

Feel free to share your thoughts :)

PR #643, opened 2019-01-30
add request options

### LINKED ISSUE #603: Write requestOptions function to replace request
Currently we proxy, certificate and ecdhCurve options in a request. This is leading to lot of redundancy in the code and could break parts where any option is missed. As mentioned in https://github.com/zulip/zulip-electron/pull/598#issuecomment-443425383 it is better to write a requestOptions function that takes the domain as input and return the appropriate object according to domain and settings to be used `request` function,

vsvipul:
**What's this PR do?**
This PR creates a new file request-util.js which replaces every request option in domain-util.js with requestOptions() as mentioned in #603 .

 As mentioned in #598  it is better to write a requestOptions function that takes the domain as input and return the appropriate object according to domain and settings to be used request function,

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

 closes #603

### DIFF
--- a/app/renderer/js/utils/domain-util.js
+++ b/app/renderer/js/utils/domain-util.js
@@ -9,10 +9,7 @@ const escape = require('escape-html');
 
 const Logger = require('./logger-util');
 
-const CertificateUtil = require(__dirname + '/certificate-util.js');
-const ProxyUtil = require(__dirname + '/proxy-util.js');
-const ConfigUtil = require(__dirname + '/config-util.js');
-const SystemUtil = require(__dirname + '/../utils/system-util.js');
+const RequestUtil = require(__dirname + '/../utils/request-util.js');
 
 const logger = new Logger({
 	file: `domain-util.log`,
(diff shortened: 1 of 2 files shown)


vsvipul:
@abhigyank @akashnimare please review

app/renderer/js/utils/domain-util.js
```diff
@@ -109,30 +106,9 @@ class DomainUtil {
 			return Promise.reject('This server has been added.');
 		}
 
-		domain = this.formatUrl(domain);
```
abhigyank: We can't remove this line as then in line 110 `domain` used would be without formatting. So this line is required. (otherwise servers added without preceeding http/s would not be added).

abhigyank (changes requested):

app/renderer/js/utils/domain-util.js
vsvipul: @abhigyank  i have added the line again.. Thanks

abhigyank (approved):
LGTM.
Can be merged after fixing the commit message. @akashnimare

vsvipul:
I have changed the commit message according to the convention

vsvipul:
@akashnimare this PR had conflicts. I just fixed those. Any updates if this can be merged?

abhigyank (approved):
Looks good. :)

akashnimare:
Nice work. Merged. Thanks, @vsvipul, @abhigyank.

PR #644, opened 2019-01-31
pdf-viewer: disable pdf-viewer window

### LINKED ISSUE #640: Disable PDF viewer
Can we disable the PDF viewer until it is fixed upstream?  Unless we want to do our own pdf viewer using pdf.js. 

Currently, it opens up the dialog to save, and also opens up a new window which is supposed to be the actual viewer, but it doesn't display the PDF.  Having to close the extra window is annoying.

vsvipul:
**What's this PR do?**
Disables pdf viewer as required in #640 ..
closes #640 

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -198,31 +198,31 @@ app.on('ready', () => {
 		app.quit();
 	});
 
-	// Show pdf in a new BrowserWindow
-	ipcMain.on('pdf-view', (event, url) => {
-		// Paddings for pdfWindow so that it fits into the main browserWindow
-		const paddingWidth = 55;
-		const paddingHeight = 22;
+	// Code to show pdf in a new BrowserWindow (currently commented out due to bug-upstream)
+	// ipcMain.on('pdf-view', (event, url) => {
+	// 	// Paddings for pdfWindow so that it fits into the main browserWindow
+	// 	const paddingWidth = 55;
+	// 	const paddingHeight = 22;
 
-		// Get the config of main browserWindow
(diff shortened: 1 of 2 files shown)


akashnimare:
@vsvipul thanks for working on this. Instead of removing the code, maybe just comment it out since we'll use it at some point in future. @abhigyank what do you think?

abhigyank:
Yeah, sounds right. Also add a comment as to the reason for commenting that part due to being a bug upstream.

vsvipul:
@akashnimare @abhigyank i have commented out the code mentioning the bug-upstream. Please tell me if some changes are required.

app/renderer/js/utils/link-util.js
```diff
@@ -35,11 +35,12 @@ class LinkUtil {
 		return isImageUrl.test(url);
 	}
 
-	isPDF(url) {
-		// test for pdf extension
-		const isPDFUrl = /\.(pdf)\?*.*$/i;
-		return isPDFUrl.test(url);
-	}
+	// Code to show pdf in a new BrowserWindow (currently commented out due to bug-upstream)
```
abhigyank: Commenting this part of the code doesn't seem to be needed. Others are fine for me.

vsvipul: done

abhigyank (approved):
Looks good to me @akashnimare .

akashnimare:
Merged, thanks @abhigyank, @vsvipul 🚀

PR #653, opened 2019-02-13
Changed saveServerIcon() to not use request() on defaultIconUrl

Swapnilr1:
---

**What's this PR do?**
Fixes issue #533 

**Screenshots?**

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

This is corrected pull request (earlier it was #627). I had made changes in master branch, so it was difficult to correct that PR. @akashnimare @abhigyank Please review this and apologies for the delay.

### DIFF
--- a/app/renderer/js/utils/domain-util.js
+++ b/app/renderer/js/utils/domain-util.js
@@ -277,28 +277,33 @@ class DomainUtil {
 		return new Promise(resolve => {
 			const filePath = this.generateFilePath(url);
 			const file = fs.createWriteStream(filePath);
-			try {
-				request(serverIconOptions).on('response', response => {
-					response.on('error', err => {
+			if (serverIconOptions.url === defaultIconUrl) {
+				logger.log('Could not get server icon.');
+				resolve(defaultIconUrl);
+			} else {
+				try {
+					request(serverIconOptions).on('response', response => {
+						response.on('error', err => {
+							logger.log('Could not get server icon.');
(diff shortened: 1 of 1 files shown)


Swapnilr1:
@akashnimare Please review.

abhigyank (approved):

Swapnilr1:
@akashnimare ping.

Swapnilr1:
@akashnimare Please review this.

akashnimare:
I'm holding this PR for a couple of reasons -
a) We are still not sure whether we should use the default Zulip icon or not (this is when the org haven't set the realm icon)
b) There are some discussions on CZO which suggest we should be using the default realm character icon
We first need to make the decision on what exactly we want here and then go for the solution. 
cc @rishig, @abhigyank, @priyank-p.

rishig:
I kind of think we should just have something that looks clearly like an error state, since this should be a pretty rare condition.

abhigyank:
Before the realm character icon implementation, we had this Zulip icon as a fallback. Since moving to realm character icon implementation this use case has become almost minimal and should be a very rare occasion that the Zulip fallback icon is being used. So we can entire remove that part or keep it as is and merge this PR to avoid the occurrence of a harmless error.

rishig:
oh, I would say to actually remove the realm character icon fallback as well, but we can do that after we fix the underlying bug that is causing it to appear at all.

PR #654, opened 2019-02-13
Check for zulip.ogg only if server error is 404, for old Zulip servers.

Swapnilr1:
---

**What's this PR do?**
Fixes issue #573 

**Screenshots?**

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

@akashnimare @abhigyank Please review this (corrected version of https://github.com/zulip/zulip-electron/pull/631 ).

### DIFF
--- a/app/renderer/js/utils/domain-util.js
+++ b/app/renderer/js/utils/domain-util.js
@@ -127,7 +127,7 @@ class DomainUtil {
 
 		// If certificate for the domain exists add it as a ca key in the request's parameter else consider only domain as the parameter for request
 		// Add proxy as a parameter if it sbeing used.
-		const checkDomain = {
+		const checkDomainFallback = {
 			url: domain + '/static/audio/zulip.ogg',
 			ca: (certificateLocation) ? certificateLocation : '',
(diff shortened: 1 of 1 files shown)


timabbott:
Can we include unit tests for this logic?  It's getting complicated.

Swapnilr1:
@timabbott Sure, I will add a unit test for this after I read some documentation regarding tests. Regarding the logic, I am not sure how to simplify it, so I would really appreciate some help if you think I have done it in a more complicated manner than is possible.

timabbott:
That wasn't a comment on the logic changes here -- it's just that this part of the codebase is the sort of thing with several conditionals that should really have unit tests.  And that changing it is a good opportunity to write that test suite (saves a lot of manual testing to understand what happens in each case).

Swapnilr1:
@timabbott As per https://chat.zulip.org/#narrow/stream/16-desktop/topic/Unit.20Tests/near/705350 I think I need to wait for #526 to get merged first, and then I can write a unit test.

akashnimare:
@kanishk98 we can close this, right?

kanishk98:
Yeah, the new validation logic should take care of this instead.

Issue #578, opened 2018-10-03
Add URL copy/paster in left sidebar.


rishig:
Discussion at https://chat.zulip.org/#narrow/stream/16-desktop/subject/access.20to.20urls/near/649032, but the summary is to add something like

![image](https://user-images.githubusercontent.com/890911/46441532-9d2fcc80-c71b-11e8-8b4b-acf2ed358818.png)

To the left sidebar to allow copying and pasting Zulip URLs. (Will need an appropriate icon for it; the circle is just a placeholder).

sunitharavi9:
@zulipbot claim

zulipbot:
Welcome to Zulip, @isunitha98selvan! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-electron/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* Sign the [Dropbox Contributor License Agreement](https://opensource.dropbox.com/cla/), so that Zulip can use your code.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

zulipbot:
Hello @isunitha98selvan, you have been unassigned from this issue because you have not updated this issue or any referenced pull requests for over 14 days.

You can reclaim this issue or claim any other issue by commenting `@zulipbot claim` on that issue.

Thanks for your contributions, and hope to see you again soon!

sunitharavi9:
Working on it.

zulipbot:
**ERROR:** This active issue has no assignee.

zulipbot:
**ERROR:** This active issue has no assignee.

zulipbot:
**ERROR:** This active issue has no assignee.

zulipbot:
**ERROR:** This active issue has no assignee.

zulipbot:
**ERROR:** This active issue has no assignee.

zulipbot:
**ERROR:** This active issue has no assignee.

zulipbot:
[...]

Issue #470, opened 2018-04-03
zulip:// URL Scheme for electron


stone212:
When I receive a missed message email from my Zulip, I see a link to read the thread.  When I click on it, it opens my browser.

But I view zulip with electron.

Can electron register a new URI scheme (zulip://) on the system it is installed on?

To be really useful you would couple this with a change in the back-end email messages so that there is an option to:

> Click here to log in to Zulip and view your new messages.
> 
> OR
> 
> Click here to open your new messages in your dedicated Zulip client.

### References
* Mobile issue (has the main technical discussion, and links to chat threads):
  https://github.com/zulip/zulip-flutter/issues/76
* Server/web issue: https://github.com/zulip/zulip/issues/11574

akashnimare:
Interesting request. Quite possible since Slack have this deep linking feature.

abhigyank:
This would an interesting feature to add. I think we can use `app.setAsDefaultProtocolClient())` to set `zulip://`. However it has different implementation ways for different OSes that needs to be looked at.
[Documentation](https://electronjs.org/docs/api/app#appsetasdefaultprotocolclientprotocol-path-args)

akashnimare:
Few things we need to add to make it work -
* Need to add `open message in the desktop app` button in the missed message email
* Pass the protocol (just need to add the protocol this in the package.json)
* Validate the passed parameter which is basically the server address. 

<img src="https://cl.ly/2P0k3B2o211S/download/Screen%20Recording%202018-04-03%20at%2004.02%20PM.gif">

kanishk98:
@zulipbot claim

kanishk98:
This seems to be fairly simple for a macOS implementation. Since the `app.on('open-url')` event getting fired sends the entire URL to the app, things get easy for a proper implementation there. 
I can't find any info in the [docs](https://electronjs.org/docs/api/app#appsetasdefaultprotocolclientprotocol-path-args) about Windows or Linux about the same level of functionality though, apart from maybe using the `makeSingleInstance` method, which is now deprecated. Not sure how to proceed here?

@akashnimare can we get some ideas from Slack's implementation of the same functionality here?

zulipbot:
Hello @kanishk98, you have been unassigned from this issue because you have not updated this issue or any referenced pull requests for over 14 days.

You can reclaim this issue or claim any other issue by commenting `@zulipbot claim` on that issue.

Thanks for your contributions, and hope to see you again soon!

[...]

PR #655, opened 2019-02-14
Add URL copy/paster to left sidebar

kanishk98:
**What's this PR do?**

Adds a URL container to the left sidebar that displays the URL of the active organization and copies it to the clipboard automatically.
Entering a URL into the container makes the app check if the domain is already saved and then switch to the relevant tab if that's the case. 

Check out #578 for the same. 

**Screenshots?**

image

**Possible improvements**
#578 shows a suggestion for a copy button. I couldn't design it as intended (inside the input field), and I think it could make for a nice improvement. 

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/css/main.css
+++ b/app/renderer/css/main.css
@@ -420,20 +420,18 @@ webview.focus {
 .url-container {
     position: absolute;
     margin-left: 53px;
+    margin-top: 5px;
 }
 
 .url-input-value {
     flex-grow: 1;
+    margin-left: 2px;
     font-size: 14px;
     border-radius: 4px;
     padding: 6px 8px;
     border: #ededed 2px solid;
     background: #fff;
     max-width: 450px;
-}
-
-.url-input-value:focus {
-    border: #4EBFAC 2px solid;
 }
 
 
--- a/app/renderer/css/main.css
+++ b/app/renderer/css/main.css
@@ -417,6 +417,25 @@ webview.focus {
     display: none !important;
 }
 
+.url-container {
+    position: absolute;
+    margin-left: 53px;
+}
+
+.url-input-value {
+    flex-grow: 1;
+    font-size: 14px;
+    border-radius: 4px;
+    padding: 6px 8px;
+    border: #ededed 2px solid;
+    background: #fff;
+    max-width: 450px;
+}
+
+.url-input-value:focus {
+    border: #4EBFAC 2px solid;
+}
+
 
 /* Full screen Popup container  */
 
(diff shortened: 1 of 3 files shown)


kanishk98:
Hello, just checking in to ask if any changes are required here. :)

abhigyank:
@kanishk98 Is the copying limited to just the server name or does it also show the topic or stream part? 
While pasting does it go to the the narrow view if we paste a link of a narrow or just focuses on the correct org?

kanishk98:
>  Is the copying limited to just the server name or does it also show the topic or stream part?

It's limited to the server URL only. 

> While pasting does it go to the the narrow view if we paste a link of a narrow or just focuses on the correct org?

There's no web-based stuff involved here. So it'll simply change tabs to another server if the URL of an already saved URL is pasted into the field. 

Currently, if a URL belonging to a server that has not been saved is pasted into the field, the app doesn't do anything. I didn't want to add an indication without a discussion. @abhigyank what do you think we should do in that case? Simply clear the field?

abhigyank:
> It's limited to the server URL only.

Okay, understood. I am not sure if that servers us much purpose here, I mean the chances of someone pasting chat.zulip.org in a toolbar to open the issue would be a very rare case I suppose, unless someone has like 20-30 added organizations (again a very very rare case, max I have seen is about 12).
Typically the use case would be as mentioned in the issue would be to narrow into a view after pasting an url, since it would help people directly opening links from browser or emails (this needs to be done very very carefully as can lead to attacks). 

>  if a URL belonging to a server that has not been saved is pasted into the field, the app doesn't do anything.

This can simple like showing a red exclamation icon or a red boundary - something that websites do to show incorrect passwords dynamically.

kanishk98:
> Typically the use case would be as mentioned in the issue would be to narrow into a view after pasting an url, since it would help people directly opening links from browser or emails (this needs to be done very very carefully as can lead to attacks).

Alright, I understand. I'll be addressing both #470 and this together since they're somewhat similar. Will get to it after my proposal submission, if that's okay. 

> This can simple like showing a red exclamation icon or a red boundary - something that websites do to show incorrect passwords dynamically.

Easy enough. I'll do this as soon as I can.
Thanks for the review.

kanishk98:
> This can simple like showing a red exclamation icon or a red boundary - something that websites do to show incorrect passwords dynamically.

Done and pushed. Works as shown below. 

link

I've already implemented the desired redirection in #716. I think I should wait for a proper review on that before using the same logic here. We can just merge both of them then.

Issue #659, opened 2019-02-20
Shortcut for cycling through organizations


abhigyank:
We can add two shortcuts to cycle through organizations. Ctrl+Tab could switch to the next organization in sidebar ( and to the first if currently on the last in list) and Ctrl+Shift+Tab could switch to the previous organization in sidebar ( and to the last if currently on the first in list).
These two options could also be added on the Windows menu.

This would make switching more "simple" on the desktop app.

abhigyank:
@zulipbot add "Type: Enhancement"

zulipbot:
**ERROR:** Label "Type: Enhancement, help wanted" does not exist and was thus not added to this issue.

abhigyank:
@zulipbot add "Type: Enhancement"

kanishk98:
@zulipbot claim

kanishk98:
@abhigyank I'd like some input on the preferred option to implement shortcuts here. I could write an event listener without using any external library, or I could do what [Electron docs](https://electronjs.org/docs/tutorial/keyboard-shortcuts) suggest and use the Mousetrap library. The latter option would definitely be more readable and require less work going ahead (assuming we're going to continue adding BrowserWindow - level shortcuts in the app).

kanishk98:
I tried both approaches mentioned above, but I realized that maintaining window-level control was somewhat tricky and cumbersome, especially when the user was typing into input fields. Electron's `MenuItem` implementation solves those problems, at the cost of adding an extra item in the Window menu up top. I don't think that's a bad idea. 

That's the approach I've used in the referenced PR.

zulipbot:
Hello @kanishk98, you have been unassigned from this issue because you have not updated this issue or any referenced pull requests for over 14 days.

You can reclaim this issue or claim any other issue by commenting `@zulipbot claim` on that issue.

Thanks for your contributions, and hope to see you again soon!

kanishk98:
Submitted PR and waiting for review.

PR #663, opened 2019-02-23
Switch to next server on Ctrl+Tab

kanishk98:
**What's this PR do?**

Adds a new option to the Window menu that allows users to change to the
next organization cyclically. Activated by Ctrl/Cmd + Tab.

Please have a look at #659. 

**Screenshots?**

ezgif com-crop

  - [x] Windows
  - [x] Linux/Ubuntu
  - [x] macOS

### DIFF
--- a/app/renderer/js/pages/preference/shortcuts-section.js
+++ b/app/renderer/js/pages/preference/shortcuts-section.js
@@ -131,11 +131,11 @@ class ShortcutsSection extends BaseSection {
                     <td>Toggle DevTools for Active Tab</td>
                   </tr>
                   <tr>
-                    <td><kbd>${userOSKey}</kbd> + <kbd>Tab</kbd></td>
+                    <td><kbd>Ctrl</kbd> + <kbd>Tab</kbd></td>
                     <td>Switch to next organization</td>
                   </tr>
                   <tr>
-                    <td><kbd>${userOSKey}</kbd> + <kbd>Shift</kbd> + <kbd>Tab</kbd></td>
+                    <td><kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>Tab</kbd></td>
                     <td>Switch to previous organization</td>
                   </tr>
                 </table>
--- a/app/renderer/js/pages/preference/shortcuts-section.js
+++ b/app/renderer/js/pages/preference/shortcuts-section.js
@@ -130,6 +130,14 @@ class ShortcutsSection extends BaseSection {
                     <td><kbd>Option</kbd><kbd>${userOSKey}</kbd><kbd>U</kbd></td>
                     <td>Toggle DevTools for Active Tab</td>
                   </tr>
+                  <tr>
+                    <td><kbd>${userOSKey}</kbd> + <kbd>Tab</kbd></td>
+                    <td>Switch to next organization</td>
+                  </tr>
+                  <tr>
+                    <td><kbd>${userOSKey}</kbd> + <kbd>Shift</kbd> + <kbd>Tab</kbd></td>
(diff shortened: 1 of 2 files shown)


akashnimare:
This looks good. Will merge it after the new release.

abhigyank:
See if you can also include Ctrl+Shift+Tab to go to previous server.

kanishk98:
@abhigyank sure, shouldn’t be a problem. I’ve got an exam this morning, will finish this up once I get back. :)

Also, just a suggestion, we use Ctrl + Shift + T to restore a tab we accidentally closed in the browser. Would it be a good idea to include this as a shortcut too, in case someone accidentally deletes an organisation (now that we’re adding more options to the context menu in the sidebar, that seems possible)

abhigyank:
I don't suppose deleting an organization accidently is much of a common use case and could be confusing for users.

kanishk98:
> I don't suppose deleting an organization accidently is much of a common use case and could be confusing for users.

Fair enough. We can always work on it later depending on feedback.

akashnimare:
Okay, this doesn't work on macOS. Maybe because the `CMD + Tab` shortcut is globally registered to switching between apps. @priyank-p any idea?

kanishk98:
> Okay, this doesn't work on macOS. Maybe because the `CMD + Tab` shortcut is globally registered to switching between apps. 

Right, makes sense. I checked [Chrome keyboard shortcuts](https://support.google.com/chrome/answer/157179?hl=en) and they seem to use ⌘ + Option + Right arrow for switching to the next tab. Would it be a good idea to do the same here?

akashnimare:
> Right, makes sense. I checked Chrome keyboard shortcuts and they seem to use ⌘ + Option + Right arrow for switching to the next tab. Would it be a good idea to do the same here?

CTRL + Tab also works for switching the tabs. Let's use the same in our app also. i.e. CTRL + Tab on all the OSes.

app/renderer/js/pages/preference/shortcuts-section.js
```diff
@@ -130,6 +130,14 @@ class ShortcutsSection extends BaseSection {
                     <td><kbd>Option</kbd><kbd>${userOSKey}</kbd><kbd>U</kbd></td>
                     <td>Toggle DevTools for Active Tab</td>
                   </tr>
+                  <tr>
+                    <td><kbd>${userOSKey}</kbd> + <kbd>Tab</kbd></td>
```
akashnimare: This should be CTRL.

app/renderer/js/pages/preference/shortcuts-section.js
```diff
@@ -130,6 +130,14 @@ class ShortcutsSection extends BaseSection {
...
+                    <td>Switch to next organization</td>
+                  </tr>
+                  <tr>
+                    <td><kbd>${userOSKey}</kbd> + <kbd>Shift</kbd> + <kbd>Tab</kbd></td>
```
akashnimare: Same.

akashnimare (changes requested):

app/renderer/js/pages/preference/shortcuts-section.js
kanishk98: I've pushed the changes needed. I don't have access to a Mac right now, so I can't check for myself if the section looks as expected. Please let me know if it doesn't.

akashnimare:
Merged. Thanks, @kanishk98.

abhigyank:
I think we forgot to change userOSKey to Ctrl in settings.

akashnimare:
We did change it for the macOS, for Windows/Linux userOSkey is fine.

PR #666, opened 2019-02-27
Add option to hide menu bar to View menu

kanishk98:
---

**What's this PR do?**

Adds option to auto-hide the menu bar (previously found only in Preferences > General) to the View menu.

I felt that the View menu should have this setting because it's more intuitive to find something like this there. This pattern is also present in multiple other apps, and as someone who hates menu bars, I'd love to have it easily accessible. 

  - [x] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/menu.js
+++ b/app/main/menu.js
@@ -165,6 +165,7 @@ class AppMenu {
 		}, {
 			label: 'Auto hide Menu bar',
 			checked: ConfigUtil.getConfigItem('autoHideMenubar'),
+			visible: process.platform !== 'darwin',
 			click(item, focusedWindow) {
 				if (focusedWindow) {
 					const newValue = !ConfigUtil.getConfigItem('autoHideMenubar');
--- a/app/main/menu.js
+++ b/app/main/menu.js
@@ -171,6 +171,7 @@ class AppMenu {
 					const newValue = !ConfigUtil.getConfigItem('autoHideMenubar');
 					focusedWindow.setAutoHideMenuBar(newValue);
 					focusedWindow.setMenuBarVisibility(!newValue);
+					focusedWindow.webContents.send('toggle-autohide-menubar', newValue);
 					ConfigUtil.setConfigItem('autoHideMenubar', newValue);
 				}
 			},
--- a/app/main/menu.js
+++ b/app/main/menu.js
@@ -162,6 +162,17 @@ class AppMenu {
 					ConfigUtil.setConfigItem('showSidebar', newValue);
 				}
 			}
+		}, {
+			label: 'Auto hide Menu bar',
+			accelerator: 'CommandOrControl+B',
+			click(item, focusedWindow) {
+				if (focusedWindow) {
+					const newValue = !ConfigUtil.getConfigItem('autoHideMenubar');
+					focusedWindow.setAutoHideMenuBar(newValue);
+					focusedWindow.setMenuBarVisibility(!newValue);
+					ConfigUtil.setConfigItem('autoHideMenubar', newValue);
+				}
+			}
 		}];
 	}
 
--- a/app/main/menu.js
+++ b/app/main/menu.js
@@ -164,7 +164,7 @@ class AppMenu {
 			}
 		}, {
 			label: 'Auto hide Menu bar',
-			checked: ConfigUtil.getConfigItem('autoHideMenubar'),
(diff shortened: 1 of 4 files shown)


abhigyank:
@kanishk98  Has there been any discussion regarding the `Ctrl+B` shortcut for this? I think that is conflicting with the bold shortcut of the webapp and most people would up getting confused.

kanishk98:
> @kanishk98 Has there been any discussion regarding the `Ctrl+B` shortcut for this? I think that is conflicting with the bold shortcut of the webapp and most people would up getting confused.

Yes, that’s true. I haven’t really thought about the shortcut; I figured I’d add the functionality and we could easily change the shortcut later.

akashnimare:
We don't need to show this on macOS

image

add the following to hide it -
```
visible: process.platform !== 'darwin',
```

app/main/menu.js
```diff
@@ -162,6 +162,18 @@ class AppMenu {
 					ConfigUtil.setConfigItem('showSidebar', newValue);
 				}
 			}
+		}, {
+			label: 'Auto hide Menu bar',
```
akashnimare, line 166: Should be hidden on macOS.

akashnimare (changes requested):

akashnimare (approved):

akashnimare:
We also need to update the value of the setting in the setting > General so you need to use IPC in order to do it (check show sidebar setting in menu for an example).
```js
			label: 'Toggle Sidebar',
			accelerator: 'CommandOrControl+Shift+S',
			click(item, focusedWindow) {
				if (focusedWindow) {
					const newValue = !ConfigUtil.getConfigItem('showSidebar');
					focusedWindow.webContents.send('toggle-sidebar', newValue);
					ConfigUtil.setConfigItem('showSidebar', newValue);
				}
			}

```

kanishk98:
@akashnimare please have another look at this PR. I've reviewed it once more on my end, and this is how it behaves on my computer:

menu

app/main/menu.js
```diff
@@ -162,6 +162,20 @@ class AppMenu {
...
+			checked: ConfigUtil.getConfigItem('autoHideMenubar'),
```
akashnimare: @priyank-p, @abhigyank I have a doubt here. What would happen here if the config files fail on startup? Maybe we should use the following to avoid that situation -

```
checked: ConfigUtil.getConfigItem('autoHideMenubar') || false ,
```

kanishk98: @akashnimare I believe I can answer that.

> What would happen here if the config files fail on startup?

This could mean two things, I think - either the `settings.json` file is completely malformed, in which case there'll be an error saying "There was an error in saving the settings" and the default settings will be set, or the field is absent and the `checked` field is therefore `null`. 
That would be the same as setting it to `false`, but I agree that my implementation could be better. 

I'll set it to:
```javascript
checked: ConfigUtil.getConfigItem('autoHideMenubar', false)
```

kanishk98: I'll also add a default setting of `autoHideMenubar: false` in `main.js`.

app/renderer/js/main.js
```diff
@@ -625,6 +626,22 @@ class ServerManagerView {
...
 
+		ipcRenderer.on('toggle-autohide-menubar', (event, autoHideMenubar, updateMenu) => {
+			if (updateMenu) {
+				ipcRenderer.send('update-menu', {
+					tabs: this.tabsForIpc,
+					activeTabIndex: this.activeTabIndex
+				});
+				return;
+			}
+			const selector = 'webview:not([class*=disabled])';
```
akashnimare: @kanishk98 can you make a function for this since we are using this in `toggle-sidebar` also?

akashnimare (changes requested):

akashnimare:
@kanishk98 can you make the changes so that I can merge it?

kanishk98:
Sorry, must've missed this. Making the changes right now.

kanishk98:
@akashnimare I've made the changes. The method is used while toggling the sidebar and automatically hiding the menu. 

Let me know if there's anything else to be done.

vsvipul (approved):
Works fine on Ubuntu. LGTM !

akashnimare:
Merged. Thanks, @kanishk98.

PR #667, opened 2019-02-27
pdf-view: add pdfjs implementation for opening pdfs

### LINKED ISSUE #664: Use pdf.js to open pdf files
PDFs files currently download onto the system when clicked on .We can used pdf.js to open them instead. A pdf reader was first used but had a bug upstream due to which it had to be disabled.

vsvipul:
fixes #664.

**What's this PR do?**
Adds the module electron-pdf-window which is an implementation of pdf.js for electron apps. 

**Screenshots?**
screenshot from 2019-02-28 01-48-13

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -5,6 +5,7 @@ const fs = require('fs');
 const electron = require('electron');
 const windowStateKeeper = require('electron-window-state');
 const isDev = require('electron-is-dev');
+const PDFWindow = require('electron-pdf-window');
 const appMenu = require('./menu');
 const { appUpdater } = require('./autoupdater');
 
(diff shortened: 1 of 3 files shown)


abhigyank:
The functionality works fine for me including download and print. 
However the module is not pdf.js but rather [electron-pdf-window](https://github.com/gerhardberger/electron-pdf-window) . It seems to be unmaintained and probably could have dependency vulnerability.

vsvipul:
@abhigyank the [pdfjs-dist](https://github.com/mozilla/pdfjs-dist) available for npm does not contain an inbuilt viewer while electron-pdf-window is an implementation of official pdfjs with the viewer inbuilt. If we want to use the official pdfjs-dist module, we'll have to whip up our own viewer files, which will require adding a lot of files. I'll look more into it.

vsvipul:
@abhigyank can't pdf-electron-window be used as a temporary fix until [this issue is fixed upstream](https://github.com/electron/electron/issues/12337), which will be very soon, as people are working on it.

abhigyank:
@akashnimare  pdf-electron-window is uses pdfjs itself so we could probably use this as well. There is an upstream PR for fixing this, it should probably be merged till our next release, this should be good till then.

abhigyank (approved):

vsvipul:
@akashnimare are we not merging this due to an extra dependency ?

akashnimare:
@vsvipul we should try to merge this if Electron is not going to support pdf in the near future.

kanishk98:
Are we still considering this as a possible improvement? If so, I can fix the merge conflicts and pitch in too.

akashnimare:
@kanishk98 what's the status of the PDF feature in the recent release of the Electron? If there is no plan from the Electron's side we can go ahead with this PR.

andreas-bulling:
This seems like a really useful feature - any news on this?

andersk:
The new plan is to open uploaded files (including PDFs) in the system web browser, using the server-side feature introduced in zulip/zulip#14500 so you don’t have to log in again. This will give a better UX than we would likely have the resources to write ourselves, without increasing the potential attack surface by adding new interactions between the app and untrusted user content.

PR #673, opened 2019-03-04
index.html: fix div indentation

vsvipul:
Fix a div indentation in index.html file

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/main.html
+++ b/app/renderer/main.html
@@ -12,39 +12,39 @@
   <div id="content">
     <div class="popup">
       <span class="popuptext hidden" id="fullscreen-popup"></span>
-  </div>
-  <div id="sidebar" class="toggle-sidebar">
-    <div id="view-controls-container">
-      <div id="tabs-container"></div>
-      <div id="add-tab" class="tab functional-tab">
-        <div class="server-tab" id="add-action">
-          <i class="material-icons">add</i>
+    </div>
+    <div id="sidebar" class="toggle-sidebar">
+      <div id="view-controls-container">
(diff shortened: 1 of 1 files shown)


abhigyank:
Surprised this existed! @akashnimare  shouldn't `xo` be taking care of such things?

akashnimare:
> Surprised this existed! @akashnimare shouldn't xo be taking care of such things?

I don't think that xo covers that file 🤔

abhigyank:
Ideally it should, shouldn't it?
I think we can merge this PR but have a look at the `xo` config for not covering this @akashnimare.
@vsvipul Can you look v into that, and create a PR of you find a fix?

vsvipul:
Sure, I'll look into it

priyank-p:
xo is basically provides rules pre-configured on top of eslint and is a JavaScript linter so that's why it can't lint html files. Though there might be eslint plugin you can add in xo config that provides html linting but that very unlikely. The correct way to lint html files would be using a separate html linter here.

akashnimare:
@vsvipul can you look into this, I think we could use `eslint-html-plugin` with `xo`.

vsvipul:
@akashnimare i looked into and found that `eslint-html-plugin` is used to lint JS contained in `<script>` tags in html files. It is not for linting HTML files. The appropriate approach here would be use a linter that can lint all html, css and js files , and not only JS. We need to find something like git-lint (pip) for npm and replace it with xo.

vsvipul:
Fixed in #726 . So, closing this.

PR #674, opened 2019-03-04
Add loading indicator to sidebar

### LINKED ISSUE #430: Loading indicator while navigating in webview
Add some kind of indicator of loading when we move from one page to another in a webview.
Example: When one logs in (especially through third party auth) or logs out, add an indication that the request for logging in or logging out has taken place and that the next webpage is loading.

This can be added just like a browser like Chrome shows loading indicator by changing the reload button when a page is loading.

Discussion can be found [here](https://chat.zulip.org/#narrow/stream/16-electron/topic/Loading.20indicator).

kanishk98:
Browser-like loading indicator added to the sidebar for better UX.

---

**What's this PR do?**

Follows up on the discussion [here](https://chat.zulip.org/#narrow/stream/16-desktop/topic/Loading.20indicator/near/519181) to add a loading indicator to the sidebar. 

The icon I've used is a simple loop picked up from the Material Design library. We can certainly include the more conventional spinner too. 

Fixes #430. Relevant discussion there too. 

**Screenshots**

image

(Please observe that the case I've shown here is the login issue we've been talking about, and that the reload icon has been replaced by the loop icon).

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/main.js
+++ b/app/renderer/js/main.js
@@ -31,6 +31,7 @@ class ServerManagerView {
 
 		const $actionsContainer = document.getElementById('actions-container');
 		this.$reloadButton = $actionsContainer.querySelector('#reload-action');
+		this.$loadingIndicator = $actionsContainer.querySelector('#loading-action');
 		this.$settingsButton = $actionsContainer.querySelector('#settings-action');
 		this.$webviewsContainer = document.getElementById('webviews-container');
 		this.$backButton = $actionsContainer.querySelector('#back-action');
@@ -38,6 +39,7 @@ class ServerManagerView {
 
 		this.$addServerTooltip = document.getElementById('add-server-tooltip');
 		this.$reloadTooltip = $actionsContainer.querySelector('#reload-tooltip');
+		this.$loadingTooltip = $actionsContainer.querySelector('#loading-tooltip');
 		this.$settingsTooltip = $actionsContainer.querySelector('#setting-tooltip');
 		this.$serverIconTooltip = document.getElementsByClassName('server-tooltip');
 		this.$backTooltip = $actionsContainer.querySelector('#back-tooltip');
@@ -49,6 +51,7 @@ class ServerManagerView {
 		this.$fullscreenEscapeKey = process.platform === 'darwin' ? '^⌘F' : 'F11';
 		this.$fullscreenPopup.innerHTML = `Press ${this.$fullscreenEscapeKey} to exit full screen`;
 
+		this.loading = {};
 		this.activeTabIndex = -1;
 		this.tabs = [];
 		this.functionalTabs = {};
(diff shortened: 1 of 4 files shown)


kanishk98:
Whatever you say, boss. @zulipbot

abhigyank:
Will review this over the weekend.

abhigyank (commented):
* The reload button changes to a loop icon when loading. I don't think that looks good from a UI point of view. Like to hear what others think about this.
* The feature is broken. Once you start navigating inside the webapp, it stays as the loop icon and doesn't change back to the reload icon. Same is the case with Add Organization tab and consequently the settings tab.
* If we continue with current icon, a tooltip on hover would be good.
* I like how the loop icon takes into consideration different webviews and is something that should be present.

kanishk98:
> The reload button changes to a loop icon when loading. I don't think that looks good from a UI point of view. Like to hear what others think about this.

I agree. The loop icon is just a placeholder, we can certainly replace that with something else. 

> The feature is broken. Once you start navigating inside the webapp, it stays as the loop icon and doesn't change back to the reload icon. Same is the case with Add Organization tab and consequently the settings tab.

I checked this behaviour. Webapp's a problem I didn't notice earlier, I'll send a fix today. I think I should just disable the loading icon from showing up in Settings. 
I don't understand where Add Organization fails for you. I'm attaching a gif for reference. 

loading_long

> If we continue with current icon, a tooltip on hover would be good.

I'll add that right away. No need to wait for selecting a new icon before adding the tooltip. 

Thanks for the review! I'll send in the changes asap.

kanishk98:
@akashnimare @abhigyank can you please review the most recent commits and let me know if there's anything else to be done?

kanishk98:
Alright, I've rebased this branch onto master and resolved all merge conflicts. Please let me know if this feature works as expected now.

akashnimare:
Looks good. @kanishk98 can you change the loader to this -

https://codepen.io/anon/pen/KjVqqg

kanishk98:
@akashnimare made the required changes. 
Also squashed all commits and updated the commit message.

akashnimare:
Thanks, @kanishk98. FYI, @rishig.

app/renderer/js/main.js
```diff
@@ -343,6 +356,14 @@ class ServerManagerView {
 				isActive: () => {
 					return this.functionalTabs[tabProps.name] === this.activeTabIndex;
 				},
+				switchLoading: (loading, url) => {
```
abhigyank, line 359: Do we require this in functional tabs? I think we are not showing loading in settings part.

abhigyank:
I love that now we can show loading in logout and third party auth. But there are certain issues with reloading - 
* I am confused as to what the behaviour will be when the browser is loading and icon is rotating. We can probably disable clicking on it then.
* When the a server has loaded and spinning has stopped then click on reload - server reloads but the icon doesnt spins. Not sure why the issue here is though. GIF (15 sec) - 
reload

kanishk98:
> I am confused as to what the behaviour will be when the browser is loading and icon is rotating. We can probably disable clicking on it then.

Yeah, good catch. I'll disable this.

> When the a server has loaded and spinning has stopped then click on reload - server reloads but the icon doesnt spins. Not sure why the issue here is though. 

Yeah, that's because once the user has navigated in the page, we turn off the loading indicator in the sidebar. I'll change this to sync with the green loading gif we have in the main screen. Should work out well enough then.

PR #681, opened 2019-03-12
Add custom configurations for enterprise

kanishk98:
---

**What's this PR do?**

Considers a new custom config file for managing user-level settings of the app. We expect that the admin deploying Zulip at enterprise will `echo` the filepath for this config file into `settings.json` and then we can read this as replacements for settings in the app. 

A major request by enterprise users is that they should be able to regulate user-level settings by maintaining a `custom_settings.json` file at a system-level settings. Following the discussion [here](https://chat.zulip.org/#narrow/stream/16-desktop/topic/high.20availability/near/701040), I have implemented a simple method to read enterprise settings using `fs`.
This assumes that the admin will `echo` the filepath of the custom config file as per their choice in the script they write and the app will just read that.

I also realised that some settings (say auto updates) should be disabled by admins but some settings (like showing the sidebar) should be accessible to the user. So I've added an `admin` field to every config item. By default, it is false, which means that admin access to change that field is not required. Handled that in ConfigUtil's `setConfigItem` directly.

**Slated improvements**

I've sent in this draft PR to make sure everyone invested in this functionality (quite a few people) can provide feedback as we go along. That being said, I plan to add the following features over the next couple of days:

- [x] 1. Extend the switches' permissions regulation to other settings fields and dialog boxes
- [x] 2. Extend similar permissions to preset organizations
- [x] 3. Investigate options we have in terms of granularity and add support for the relevant ones
- [x] 4. Document the new files made and the process of editing custom settings

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/css/preference.css
+++ b/app/renderer/css/preference.css
@@ -453,6 +453,11 @@ i.open-tab-button {
     user-select: none;
 }
 
+.disallowed:hover {
+    background-color: rgba(241, 241, 241, 1.000);
+    cursor: not-allowed;
+}
+
 input.toggle-round + label {
     padding: 2px;
     width: 50px;
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -228,7 +228,7 @@ class ServerManagerView {
 			const setting = i as keyof SettingsOptions;
 			// give preference to defaults defined in global_config.json
 			if (EnterpriseUtil.configItemExists(setting)) {
-				ConfigUtil.setConfigItem(setting, EnterpriseUtil.getConfigItem(setting));
+				ConfigUtil.setConfigItem(setting, EnterpriseUtil.getConfigItem(setting), true);
 			} else if (ConfigUtil.getConfigItem(setting) === null) {
 				ConfigUtil.setConfigItem(setting, settingOptions[setting]);
 			}
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -138,12 +138,12 @@ class ServerManagerView {
 
 	init(): void {
 		this.loadProxy().then(() => {
+			this.initDefaultSettings();
 			this.initSidebar();
 			this.initPresetOrgs();
 			this.initTabs();
 			this.initActions();
 			this.registerIpcs();
-			this.initDefaultSettings();
 		});
 	}
 
@@ -226,8 +226,11 @@ class ServerManagerView {
 
 		for (const i in settingOptions) {
 			const setting = i as keyof SettingsOptions;
-			if (ConfigUtil.getConfigItem(i) === null) {
-				ConfigUtil.setConfigItem(i, settingOptions[setting]);
+			// give preference to defaults defined in global_config.json
+			if (EnterpriseUtil.configItemExists(setting)) {
+				ConfigUtil.setConfigItem(setting, EnterpriseUtil.getConfigItem(setting));
+			} else if (ConfigUtil.getConfigItem(setting) === null) {
+				ConfigUtil.setConfigItem(setting, settingOptions[setting]);
 			}
 		}
 	}
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -140,7 +140,9 @@ class ServerManagerView {
 		this.loadProxy().then(() => {
 			this.initDefaultSettings();
 			this.initSidebar();
-			this.initPresetOrgs();
+			if (EnterpriseUtil.configFile) {
+				this.initPresetOrgs();
+			}
 			this.initTabs();
 			this.initActions();
 			this.registerIpcs();
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -285,8 +285,15 @@ class ServerManagerView {
 				ipcRenderer.send('reload-full-app');
 			}
 		} else if (domainsAdded.length > 0) {
-			// no enterprise domains added
-			const { title, content } = Messages.enterpriseOrgError(domainsAdded.length);
+			// find all orgs that failed
+			const failedDomains: string[] = [];
+			for (const org of this.presetOrgs) {
+				if (DomainUtil.duplicateDomain(org)) {
+					continue;
+				}
+				failedDomains.push(org);
+			}
+			const { title, content } = Messages.enterpriseOrgError(domainsAdded.length, failedDomains);
 			dialog.showErrorBox(title, content);
 			if (DomainUtil.getDomains().length === 0) {
 				// no orgs present, stop showing loading gif
@@ -721,8 +728,12 @@ class ServerManagerView {
 							message: 'Are you sure you want to disconnect this organization?'
 						}, response => {
 							if (response === 0) {
-								DomainUtil.removeDomain(index);
-								ipcRenderer.send('reload-full-app');
+								if (DomainUtil.removeDomain(index)) {
+									ipcRenderer.send('reload-full-app');
+								} else {
+									const { title, content } = Messages.orgRemovalError(DomainUtil.getDomain(index).url);
+									dialog.showErrorBox(title, content);
+								}
 							}
 						});
 					}
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -21,6 +21,7 @@ import ReconnectUtil = require('./utils/reconnect-util');
 import Logger = require('./utils/logger-util');
 import CommonUtil = require('./utils/common-util');
 import EnterpriseUtil = require('./utils/enterprise-util');
+import Messages = require('./../../resources/messages');
 
 const logger = new Logger({
 	file: 'errors.log',
@@ -283,6 +284,10 @@ class ServerManagerView {
 			} else {
 				ipcRenderer.send('reload-full-app');
 			}
+		} else if (domainsAdded.length > 0) {
+			// no enterprise domains added
+			const { title, content } = Messages.enterpriseOrgError(domainsAdded.length);
+			dialog.showErrorBox(title, content);
 		}
 	}
 
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -20,6 +20,7 @@ import DNDUtil = require('./utils/dnd-util');
 import ReconnectUtil = require('./utils/reconnect-util');
 import Logger = require('./utils/logger-util');
 import CommonUtil = require('./utils/common-util');
+import EnterpriseUtil = require('./utils/enterprise-util');
 
 const logger = new Logger({
 	file: 'errors.log',
@@ -135,6 +136,7 @@ class ServerManagerView {
 	init(): void {
 		this.loadProxy().then(() => {
 			this.initSidebar();
+			this.initPresetOrgs();
 			this.initTabs();
 			this.initActions();
 			this.registerIpcs();
@@ -230,6 +232,40 @@ class ServerManagerView {
 	initSidebar(): void {
 		const showSidebar = ConfigUtil.getConfigItem('showSidebar', true);
 		this.toggleSidebar(showSidebar);
+	}
+
+	async queueDomain(domain: any): Promise<boolean> {
+		// allows us to start adding multiple domains to the app simultaneously
+		// promise of addition resolves in both cases, but we consider it rejected
+		// if the resolved value is false
+		try {
+			const serverConf = await DomainUtil.checkDomain(domain);
+			await DomainUtil.addDomain(serverConf);
+			return true;
+		} catch (err) {
+			logger.error(err);
+			logger.error('Could not add ' + domain + '. Please contact your system administrator.');
+			return false;
+		}
+	}
+
+	async initPresetOrgs(): Promise<void> {
+		// read preset organizations from global_config.json and queues them
+		// for addition to the app's domains
+		const presetOrgs = EnterpriseUtil.getConfigItem('presetOrganizations', []);
+		// set to true if at least one new domain is added
+		const domainPromises = [];
+		for (const url in presetOrgs) {
+			if (DomainUtil.duplicateDomain(presetOrgs[url])) {
+				continue;
+			}
+			domainPromises.push(this.queueDomain(presetOrgs[url]));
+		}
+		const domainsAdded = await Promise.all(domainPromises);
+		if (domainsAdded.includes(true)) {
+			// at least one domain was resolved
+			ipcRenderer.send('reload-full-app');
+		}
 	}
 
 	initTabs(): void {
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -94,6 +94,7 @@ class ServerManagerView {
 	tabs: ServerOrFunctionalTab[];
 	functionalTabs: AnyObject;
 	tabIndex: number;
+	presetOrgs: string[];
 	constructor() {
 		this.$addServerButton = document.querySelector('#add-tab');
 		this.$tabsContainer = document.querySelector('#tabs-container');
@@ -129,6 +130,7 @@ class ServerManagerView {
 		this.loading = {};
 		this.activeTabIndex = -1;
 		this.tabs = [];
+		this.presetOrgs = [];
 		this.functionalTabs = {};
 		this.tabIndex = 0;
 	}
@@ -253,14 +255,14 @@ class ServerManagerView {
 		// read preset organizations from global_config.json and queues them
 		// for addition to the app's domains
 		const preAddedDomains = DomainUtil.getDomains();
-		const presetOrgs = EnterpriseUtil.getConfigItem('presetOrganizations', []);
+		this.presetOrgs = EnterpriseUtil.getConfigItem('presetOrganizations', []);
 		// set to true if at least one new domain is added
 		const domainPromises = [];
-		for (const url in presetOrgs) {
-			if (DomainUtil.duplicateDomain(presetOrgs[url])) {
+		for (const url of this.presetOrgs) {
+			if (DomainUtil.duplicateDomain(url)) {
 				continue;
 			}
-			domainPromises.push(this.queueDomain(presetOrgs[url]));
+			domainPromises.push(this.queueDomain(url));
 		}
 		const domainsAdded = await Promise.all(domainPromises);
 		if (domainsAdded.includes(true)) {
@@ -296,8 +298,11 @@ class ServerManagerView {
 			this.activateTab(ConfigUtil.getConfigItem('lastActiveTab'));
 			// Remove focus from the settings icon at sidebar bottom
 			this.$settingsButton.classList.remove('active');
-		} else {
+		} else if (this.presetOrgs.length === 0) {
+			// not attempting to add organisations in the background
 			this.openSettings('AddServer');
+		} else {
+			this.showLoading(true);
 		}
 	}
 
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -252,6 +252,7 @@ class ServerManagerView {
 	async initPresetOrgs(): Promise<void> {
 		// read preset organizations from global_config.json and queues them
 		// for addition to the app's domains
+		const preAddedDomains = DomainUtil.getDomains();
 		const presetOrgs = EnterpriseUtil.getConfigItem('presetOrganizations', []);
 		// set to true if at least one new domain is added
 		const domainPromises = [];
@@ -264,7 +265,22 @@ class ServerManagerView {
 		const domainsAdded = await Promise.all(domainPromises);
 		if (domainsAdded.includes(true)) {
 			// at least one domain was resolved
-			ipcRenderer.send('reload-full-app');
+			if (preAddedDomains.length > 0) {
+				// user already has servers added
+				// ask them before reloading the app
+				dialog.showMessageBox({
+					type: 'question',
+					buttons: ['Yes', 'Later'],
+					defaultId: 0,
+					message: 'New server' + (domainsAdded.length > 1 ? 's' : '') + ' added. Reload app now?'
+				}, response => {
+					if (response === 0) {
+						ipcRenderer.send('reload-full-app');
+					}
+				});
+			} else {
+				ipcRenderer.send('reload-full-app');
+			}
 		}
 	}
 
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -288,6 +288,10 @@ class ServerManagerView {
 			// no enterprise domains added
 			const { title, content } = Messages.enterpriseOrgError(domainsAdded.length);
 			dialog.showErrorBox(title, content);
+			if (DomainUtil.getDomains().length === 0) {
+				// no orgs present, stop showing loading gif
+				this.openSettings('AddServer');
+			}
 		}
 	}
 
--- a/app/renderer/js/utils/enterprise-util.ts
+++ b/app/renderer/js/utils/enterprise-util.ts
@@ -52,6 +52,19 @@ class EnterpriseUtil {
 		this.reloadDB();
 		return (this.enterpriseSettings[key] !== undefined);
 	}
+
+	isPresetOrg(url: string): boolean {
+		if (!this.configItemExists('presetOrganizations')) {
+			return false;
+		}
+		const presetOrgs = this.enterpriseSettings.presetOrganizations;
+		for (const org of presetOrgs) {
+			if (url.includes(org)) {
+				return true;
+			}
+		}
+		return false;
+	}
 }
 
 export = new EnterpriseUtil();
--- a/app/renderer/js/utils/enterprise-util.ts
+++ b/app/renderer/js/utils/enterprise-util.ts
@@ -13,6 +13,7 @@ let instance: null | EnterpriseUtil = null;
 class EnterpriseUtil {
 	// todo: replace enterpriseSettings type with an interface once settings are final
 	enterpriseSettings: any;
+	configFile: boolean;
 	constructor() {
 		if (instance) {
 			return instance;
@@ -30,6 +31,7 @@ class EnterpriseUtil {
 
 		enterpriseFile = path.resolve(enterpriseFile);
 		if (fs.existsSync(enterpriseFile)) {
+			this.configFile = true;
 			try {
 				const file = fs.readFileSync(enterpriseFile, 'utf8');
 				this.enterpriseSettings = JSON.parse(file);
@@ -37,11 +39,16 @@ class EnterpriseUtil {
 				logger.log('Error while JSON parsing global_config.json: ');
 				logger.log(err);
 			}
+		} else {
+			this.configFile = false;
 		}
 	}
 
 	getConfigItem(key: string, defaultValue?: any): any {
 		this.reloadDB();
+		if (!this.configFile) {
+			return defaultValue;
+		}
 		if (defaultValue === undefined) {
 			defaultValue = null;
 		}
@@ -50,11 +57,14 @@ class EnterpriseUtil {
 
 	configItemExists(key: string): boolean {
 		this.reloadDB();
+		if (!this.configFile) {
+			return false;
+		}
 		return (this.enterpriseSettings[key] !== undefined);
 	}
 
 	isPresetOrg(url: string): boolean {
-		if (!this.configItemExists('presetOrganizations')) {
+		if (!this.configFile || !this.configItemExists('presetOrganizations')) {
 			return false;
 		}
 		const presetOrgs = this.enterpriseSettings.presetOrganizations;
--- a/app/renderer/js/utils/enterprise-util.ts
+++ b/app/renderer/js/utils/enterprise-util.ts
@@ -1,0 +1,57 @@
+import fs = require('fs');
+import path = require('path');
+
+import Logger = require('./logger-util');
+
+const logger = new Logger({
+	file: 'enterprise-util.log',
+	timestamp: true
+});
+
+let instance: null | EnterpriseUtil = null;
+
+class EnterpriseUtil {
+	// todo: replace enterpriseSettings type with an interface once settings are final
+	enterpriseSettings: any;
+	constructor() {
+		if (instance) {
+			return instance;
+		}
+		instance = this;
+
+		this.reloadDB();
+	}
+
+	reloadDB(): void {
+		let enterpriseFile = '/etc/zulip-desktop-config/global_config.json';
+		if (process.platform === 'win32') {
+			enterpriseFile = 'C:\\Program Files\\Zulip-Desktop-Config\\global_config.json';
+		}
+
+		enterpriseFile = path.resolve(enterpriseFile);
+		if (fs.existsSync(enterpriseFile)) {
+			try {
+				const file = fs.readFileSync(enterpriseFile, 'utf8');
+				this.enterpriseSettings = JSON.parse(file);
+			} catch (err) {
+				logger.log('Error while JSON parsing global_config.json: ');
+				logger.log(err);
+			}
+		}
+	}
+
+	getConfigItem(key: string, defaultValue?: any): any {
+		this.reloadDB();
+		if (defaultValue === undefined) {
+			defaultValue = null;
+		}
+		return this.configItemExists(key) ? this.enterpriseSettings[key] : defaultValue;
+	}
+
+	configItemExists(key: string): boolean {
+		this.reloadDB();
+		return (this.enterpriseSettings[key] !== undefined);
+	}
+}
+
+export = new EnterpriseUtil();
--- a/app/resources/messages.ts
+++ b/app/resources/messages.ts
@@ -22,6 +22,13 @@ class Messages {
 		\nUnless you have a good reason to believe otherwise, you should not proceed.
 		\nYou can click here if you'd like to proceed with the connection.`;
 	}
+
+	enterpriseOrgError(length: number): any {
+		return {
+			title: `Could not add ${length} ${length === 1 ? `organization` : `organizations`}`,
+			content: `Please contact your system administrator.`
+		};
+	}
 }
 
 export = new Messages();
--- a/app/resources/messages.ts
+++ b/app/resources/messages.ts
@@ -23,9 +23,20 @@ class Messages {
 		\nYou can click here if you'd like to proceed with the connection.`;
 	}
 
-	enterpriseOrgError(length: number): any {
+	enterpriseOrgError(length: number, domains: string[]): any {
+		let domainList = '';
+		for (const domain of domains) {
+			domainList += `• ${domain}\n`;
+		}
 		return {
-			title: `Could not add ${length} ${length === 1 ? `organization` : `organizations`}`,
+			title: `Could not add the following ${length === 1 ? `organization` : `organizations`}`,
+			content: `${domainList}\nPlease contact your system administrator.`
+		};
+	}
+
+	orgRemovalError(url: string): any {
+		return {
+			title: `Removing ${url} is a restricted operation.`,
 			content: `Please contact your system administrator.`
 		};
 	}
--- a/app/resources/messages.ts
+++ b/app/resources/messages.ts
@@ -1,3 +1,8 @@
+interface DialogBoxError {
+	title: string;
+	content: string;
+}
+
 class Messages {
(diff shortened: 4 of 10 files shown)


app/renderer/js/utils/custom-config-util.js
```diff
@@ -0,0 +1,93 @@
...
+			this.db.reload();
+		} catch (err) {
+			logger.error('Error while reloading custom_settings.json: ');
+			logger.error(err);
+		}
+		const value = this.db.getData('/')[key];
+		return (value !== undefined);
+	}
+
+	removeConfigItem(key) {
```
akashnimare: @kanishk98 I don't think we will be able to use this function since we won't have rights to remove/delete the data from that particular directory. cc, @priyank-p.

akashnimare (changes requested):

app/renderer/js/utils/custom-config-util.js
kanishk98: Okay, makes sense. I'll remove this method.

priyank-p: Also, I don't think we will need the `node-json-db` here. We just need to attempt to read the json which you can do with with reading the file with `fs` and `JSON.parse`.

app/renderer/js/main.js
```diff
@@ -147,6 +149,20 @@ class ServerManagerView {
 		}
 	}
 
+	initEnterpriseSettings() {
+		const filepath = ConfigUtil.getConfigItem('enterpriseFilepath', null);
```
priyank-p: The whole chunk here is wrong; we want something like
```javascript
const possibleFilePaths = [
  '/etc/zulip-desktop/enterprise.json', // mac and linux
  'C://Program Files/Zulip-Desktop-Config' // windows
].map(filepath => path.resolve(filepath);
```
and then check each file path by looping through the each of them. Also import the node.js path module here to make sure it parse the file path so they are cross-platform safe.

app/renderer/js/main.js
```diff
@@ -147,6 +149,20 @@ class ServerManagerView {
...
+		if (filepath) {
+			fs.readFile(filepath, (err, data) => {
+				if (err) {
+					logger.reportSentry(err);
```
priyank-p: This is wrong. If the config file doesn't exist it will report that to sentry and spam it with these reports. It is ok for the config file to not exist.

app/renderer/js/main.js
```diff
@@ -147,6 +149,20 @@ class ServerManagerView {
...
+					return;
+				}
+				const enterpriseSettings = JSON.parse(data);
+				console.log(enterpriseSettings);
```
priyank-p: Remove this debug `console.log`

app/renderer/js/utils/config-util.js
```diff
@@ -50,7 +50,7 @@ class ConfigUtil {
 			this.setConfigItem(key, defaultValue);
 			return defaultValue;
 		} else {
-			return value;
+			return value.value;
```
priyank-p: What's the story behind this change? Can you add a comment here or in the commit message about this.

app/renderer/js/utils/config-util.js
```diff
@@ -66,7 +66,24 @@ class ConfigUtil {
 	}
 
 	setConfigItem(key, value) {
-		this.db.push(`/${key}`, value, true);
+		const item = {};
+		if (this.isConfigItemExists(key) && this.getConfigItem(key).admin) {
+			// if admin is set to true, user is not allowed to change that setting
+			return;
+		}
+		if (this.isConfigItemExists(key) && (value === null || value.admin === undefined)) {
```
priyank-p: All of these conditionals look fragile can you explain what these are if statement are doing for all the cases. I definitely feel there is cleaner and easier way to write these here. Also, I recommend checking `value.admin !== true` here.

app/renderer/js/utils/enterprise-util.js
```diff
@@ -0,0 +1,53 @@
+const fs = require('fs');
```
priyank-p: This change here should have been done in the first commit; I ended up reviewing a change that you end up removing in the end.

app/renderer/js/utils/enterprise-util.js
```diff
@@ -0,0 +1,53 @@
...
+		} else {
+			enterpriseJsonPath = '/etc/zulip/enterprise_config.json';
+		}
+		try {
+			const file = fs.readFileSync(enterpriseJsonPath, 'utf8');
+			this.enterpriseSettings = JSON.parse(file);
+		} catch (err) {
+			logger.error('Error while JSON parsing enterprise_config.json: ');
+			logger.error(err);
+			logger.reportSentry(err);
```
priyank-p: See the comment I added earlier for reporting this.

app/renderer/js/utils/enterprise-util.js
```diff
@@ -0,0 +1,53 @@
...
+		instance = this;
+
+		this.reloadDB();
+		return instance;
+	}
+
+	reloadDB() {
+		let enterpriseJsonPath = null;
+		if (process.platform === 'win32') {
+			enterpriseJsonPath = 'C:\\Program Files\\Zulip\\enterprise_config.json';
```
priyank-p: I would avoid using `Zulip` directory since if the admin installs a desktop app it ends up here. Also see the comment I made earlier (possible would be marked as outdated) about using path module and the file paths.

app/renderer/js/utils/config-util.js
```diff
@@ -65,26 +66,11 @@ class ConfigUtil {
 		return (value !== undefined);
 	}
 
-	setConfigItem(key, value) {
-		const item = {};
-		if (this.isConfigItemExists(key) && this.getConfigItem(key).admin) {
-			// if admin is set to true, user is not allowed to change that setting
-			return;
+	setConfigItem(key, value, override = false) {
```
priyank-p: What the point of having the optional override parameter, for future use?

priyank-p (commented):
@kanishk98 sorry for delayed review! I went through half of the changes didn't get to review the preset orgs commit and later.

First, I recommend you clean up you git history, you made a lot of changes in first couple of commit that you changed in later commit. I just waste of time to review those changes.

app/renderer/js/utils/enterprise-util.js
akashnimare: Would be great if we can use the electron's userData api here to get the correct path of the installed app.

app/renderer/js/utils/config-util.js
kanishk98: While initializing default settings for the first time, the app may read a few admin-only fields from the custom config file that will need to reflect in the user's `settings.json`. To keep code simple in `setConfigItem()`, I just set the optional override parameter to true when initializing the app. Requires less/no changes elsewhere in the app that this function is used.

kanishk98: Here is the code in `initDefaultSettings()`:
```javascript
if (EnterpriseUtil.isConfigItemExists(i)) {
	ConfigUtil.setConfigItem(i, EnterpriseUtil.getConfigItem(i), true);
} else {
	ConfigUtil.setConfigItem(i, settingOptions[i]);
}
```

priyank-p: OK, make sense.

kanishk98: Thanks for the review. I'll clean up git history and make the changes required.

kanishk98:
@akashnimare @priyank-p it'll be great if either of you can find some time to review the recent changes. 

I'll be making the required UI changes that should correspond to disallowed, admin-only actions and setup stuff (`initPresetOrgs()`) in a couple of hours.

app/renderer/js/main.js
```diff
@@ -152,6 +158,34 @@ class ServerManagerView {
 		this.toggleSidebar(showSidebar);
 	}
 
+	initPresetOrgs() {
+		const presetOrgs = EnterpriseUtil.getConfigItem('presetOrganizations', []);
+		const addDomainPromises = [];
+		for (const url in presetOrgs) {
+			if (!DomainUtil.duplicateDomain(presetOrgs[url])) {
```
priyank-p: Let's do:
```javascript
if (DomainUtil.duplicateDomain(presetOrgs[url])) {
  continue;
}
```

here and the rest the stuff at bottom since the chunk inside the `if` statement is huge.

app/renderer/js/main.js
```diff
@@ -152,6 +158,34 @@ class ServerManagerView {
 		this.toggleSidebar(showSidebar);
 	}
 
+	initPresetOrgs() {
```
priyank-p: Let make this a `async` function so we can use `await` and avoid the complex `then` chaining!

app/renderer/js/utils/enterprise-util.js
```diff
@@ -0,0 +1,69 @@
...
+		this.reloadDB();
+		const value = this.enterpriseSettings[key];
+		if (value === undefined) {
+			return defaultValue;
+		} else {
+			return value.value;
+		}
+	}
+
+	isConfigItemExists(key) {
```
priyank-p: For the function name here `isConfigItemExists` sounds odd; let's make it `configItemExists`.

app/renderer/js/main.js
```diff
@@ -142,7 +144,11 @@ class ServerManagerView {
 
 		for (const i in settingOptions) {
 			if (ConfigUtil.getConfigItem(i) === null) {
-				ConfigUtil.setConfigItem(i, settingOptions[i]);
+				if (EnterpriseUtil.isConfigItemExists(i)) {
```
priyank-p: The `i` here is a setting name here right? I think it makes sense to change the `i` variable `setting` in a separate commit.

app/renderer/js/utils/enterprise-util.js
```diff
@@ -0,0 +1,68 @@
...
+		this.reloadDB();
+		const value = this.enterpriseSettings[key];
+		return (value !== undefined);
+	}
+}
+
+module.exports = new EnterpriseUtil();
```
priyank-p: Missing ending newline here.

app/renderer/js/utils/enterprise-util.js
```diff
@@ -0,0 +1,68 @@
...
+		if (this.isConfigItemExists(key)) {
+			return this.enterpriseSettings[key].isAdminOnly;
+		}
+		// if item does not exist, user can change setting
+		return false;
+	}
+
+	getConfigItem(key, defaultValue = null) {
+		this.reloadDB();
+		const value = this.enterpriseSettings[key];
```
priyank-p:
```javascript
const { value } = this.enterpriseSettings[key];
```

here, so we don't need to do `value.value` below it would be just `value`.

app/renderer/js/utils/enterprise-util.js
```diff
@@ -0,0 +1,68 @@
...
+		if (value === undefined) {
+			return defaultValue;
+		} else {
+			return value.value;
+		}
```
priyank-p: This `if else` block could be just `return (value || defaultValue)`.

app/renderer/js/main.js
priyank-p: Add a comment here stating what this does. I think this initializes default based on enterprise config. If so mention that in a comment.

app/renderer/js/utils/enterprise-util.js
```diff
@@ -25,17 +26,17 @@ class EnterpriseUtil {
 	reloadDB() {
 		let enterpriseJsonPath = null;
 		if (process.platform === 'win32') {
-			enterpriseJsonPath = 'C:\\Program Files\\Zulip\\enterprise_config.json';
+			enterpriseJsonPath = 'C:\\Program Files\\Zulip-Desktop-Config\\enterprise_config.json';
 		} else {
-			enterpriseJsonPath = '/etc/zulip/enterprise_config.json';
+			enterpriseJsonPath = '/etc/zulip-desktop-config/enterprise_config.json';
```
priyank-p: These changes should have been part of the previous change.

@kanishk98 [these docs](https://zulip.readthedocs.io/en/latest/git/fixing-commits.html#fixing-older-commits) should help you out with this stuff if you are having trouble or just not confident in doing git rebase.

app/renderer/js/utils/enterprise-util.js
```diff
@@ -25,17 +26,17 @@ class EnterpriseUtil {
...
 		}
 		try {
+			enterpriseJsonPath = path.resolve(enterpriseJsonPath);
 			const file = fs.readFileSync(enterpriseJsonPath, 'utf8');
 			this.enterpriseSettings = JSON.parse(file);
 		} catch (err) {
-			logger.error('Error while JSON parsing enterprise_config.json: ');
-			logger.error(err);
-			logger.reportSentry(err);
+			logger.log('Error while JSON parsing enterprise_config.json: ');
```
priyank-p: ditto :)

app/renderer/js/utils/domain-util.js
```diff
@@ -103,11 +112,14 @@ class DomainUtil {
...
 	// ignoreCerts parameter helps in fetching server icon and
 	// other server details when user chooses to ignore certificate warnings
-	checkDomain(domain, ignoreCerts = false, silent = false) {
+	checkDomain(domain, {ignoreCerts = false, silent = false, override = false} = {}) {
 		if (!silent && this.duplicateDomain(domain)) {
 			// Do not check duplicate in silent mode
 			return Promise.reject('This server has been added.');
 		}
+		if (EnterpriseUtil.isAdminOnly('presetOrganizations') && override === false) {
+			return Promise.reject('Adding organizations is a restricted operation. Please contact your system administrator.');
```
priyank-p: I don't think we want to restrict users from adding a new organization. The most we should restrict them to do is removing a preset org (and even that I feel like should be a new setting).

cc @akashnimare

priyank-p (commented):
@kanishk98 Left a lot of cleanup comment for code, some are simple whereas others are more involved changes.

Also left a comment about the preset org commit. Can you briefly explain what everything does? For example, adding what config affect what parts of the app, etc.

app/renderer/js/utils/enterprise-util.js
kanishk98: Sure sure. I forgot to rebase these changes.

app/renderer/js/utils/domain-util.js
kanishk98: Okay, I'll wait for more discussion on this.

app/renderer/js/main.js
kanishk98: Ah okay. I'd not seen `async` used anywhere so far, I just assumed we didn't like it in the codebase for some reason. I'll change it.

priyank-p: It not visible in the codebase because when most of the stuff was written `async` wasn't supported. Try using whenever possible it will simplify the code significantly. The `then` changing gets hard to follow.

kanishk98:
> Left a lot of cleanup comment for code, some are simple whereas others are more involved changes.

Sure, I'll clean this up. 

> Can you briefly explain what everything does? For example, adding what config affect what parts of the app, etc.

I'll be documenting this, but here's the sample .json file I've been using to test this PR:
```json
{
	"autoHideMenubar": {
		"value": false,
		"isAdminOnly": true
	}, 
	"useSystemProxy": {
		"value": false, 
		"isAdminOnly": false
	},
	"presetOrganizations": {
		"value": ["https://kanishk.com"],
		"isAdminOnly": true
	}
}
```

We have support for the settings handled by `ConfigUtil` and `presetOrganizations` so far. If a setting's `isAdminOnly` field is set, the user can't change that setting, and if it's the `presetOrganizations` field that's admin-only, then the user can neither add nor delete organizations. (Will be changing this based on discussion above).

kanishk98:
Fixed up relevant commits directly and force-pushed. Continuing further improvements now. :)

kanishk98:
I realized I hadn't handled the case when an admin might want to change the preset organizations they had decided on initially. Now when that happens, I think the app should silently add the new organizations in the background and if the addition is successful, then it should prompt the user to reload the app (similar to how we notify the user that an update has been downloaded). 

If an organization has been deleted and the app loads after this config change, then I think we should immediately reload the app without waiting for the user to confirm. 

I think I've cleaned up my commits as required by @priyank-p, so I'll go ahead and start writing these changes. Please let me know what you think.

kanishk98:
I've acted on the comment above and also fixed something we overlooked wrt admin-only settings earlier: the user would've been able to go into `settings.json` and change stuff there. So I altered [this commit](https://github.com/zulip/zulip-desktop/pull/681/commits/577b282ad9b510774fb38ef829e4400ea16d2f90) and rebased to accomodate this change. Here's the new line of code when considering default settings:

```javascript
for (const setting in settingOptions) {
			if (ConfigUtil.getConfigItem(setting) === null || EnterpriseUtil.isAdminOnly(setting)) {
```

This forces the app to reset any admin-only settings each time it is loaded.

kanishk98:
I have a few UI-specific concerns:

1. When adding preset organizations, the app's UI doesn't indicate anything. This could be solved via 2 ways, I think - one is by showing a rotating icon overlaying the app or (my preference) a rotating icon in the sidebar similar to the one implemented in #674. 
2. When trying to change an admin-only setting, we could either point out to the user every time that the setting is fixed (not a good idea), show the switch snapping back to its config-oriented setting by using a timeout, or we could shake the app angrily (I believe macOS apps do something similar when they're experiencing an error, not sure). 
3. What kind of UI should we present to the admins for modifying these settings, if any? I believe we discussed a web UI earlier on the czo private group, but I'd like to be a little more specific. 

@rishig please let us know what you think.

priyank-p:
@kanishk98 I started a thread on [#desktop Enterprise Configuration UI](https://chat.zulip.org/#narrow/stream/16-desktop/topic/Enterprise.20Configuration.20UI) for this.

priyank-p:
I pushed a version that:
- Add a commit on top that just adds `enterprise-util.js`.
- Removes `require("process")` since its unnecessary and weird.
- In the `reloadDB` function you were reading the file and parsing it in the `try` block and in the `catch` block you are logging that there is error parsing json which could be wrong  since the `fs.readFile` can throw error if the file doesn't exist. I handled the case by moving the try block inside the if condition that check `fs.

kanishk98:
A list of changes I've made since the last review:

1. Modified [ee937f](https://github.com/zulip/zulip-desktop/pull/681/commits/ee937f9250a31ea3e3dd586f6f182a091ee9a9af) to use `async-await` instead of `then` chains. 
2. Detect changes in the config file so that the app automatically removes any organisations not in the config if preset organisations are set as admin-only and also adds apps in the background if new orgs are added later.
3. Show `cursor: not-allowed` on admin-only fields.
4. Disable required configurable settings in the menu if setting is admin-only. 
5. Show the standard loading gif when the app has no domains added and orgs are being added in the background.

app/renderer/js/main.js
```diff
@@ -159,6 +160,31 @@ class ServerManagerView {
...
+		const presetOrgs = EnterpriseUtil.getConfigItem('presetOrganizations', []);
+		// set to true if at least one new domain is added
+		let domainAdded = false;
+		for (const url in presetOrgs) {
+			if (DomainUtil.duplicateDomain(presetOrgs[url])) {
+				continue;
+			}
+			try {
+				// eslint-disable-next-line no-await-in-loop
+				const serverConf = await DomainUtil.checkDomain(presetOrgs[url], {override: true});
```
priyank-p: Using `await` in for loop, as the linting rule enforces, is bad idea. Essentially you are waiting for one operation to finish while other could have been in progress. You should split the code inside of the try block into a async function, then on each loop call the function which will return a promise and store it in a array. Then after the loop do `await Promise.all(<array-of-promise>);`.

kanishk98: You're right, I should've taken care of that. 
I'll resolve with `false` when there's an error and with `true` when the domain gets added successfully, since we want to reload if at least one org gets added.

kanishk98:
Let me know if any more changes are required.

app/renderer/js/utils/domain-util.js
akashnimare: Agreed.

kanishk98:
I've changed the PR to allow adding orgs at all times, and show a prompt disabling removal of the orgs specified by the admin, if they're set to be admin-only. 
I also added a doc explaining the enterprise config file. 

@priyank-p do you think we should consider using an npm module to check if the user is root, and then provide an auth prompt or override access depending on that? (Currently, we'll need the user to change the config file and then reload the app). 

@akashnimare I'd mentioned writing a script for generating the config file earlier - I think accepting input on the URLs required and the admin only setting is the best idea for that. (Similar to what we get when we run `$ npm init`)

kanishk98:
> you think we should consider using an npm module to check if the user is root, and then provide an auth prompt or override access depending on that?

I think this is a good idea. I added [sudo-prompt](https://www.npmjs.com/package/sudo-prompt) and this is the flow for removing an org now if preset orgs are admin-only:

sudo

One complication that this brings is that we'll need to remember the orgs which were deleted by admins so we don't keep removing them. 

(would just editing the config file directly from here be a good idea?)

rishig:
I think we should just show an error message saying that this organization is configured by /path/to/config/file, and say to contact their Zulip administrator.

kanishk98:
> I think we should just show an error message saying that this organization is configured by /path/to/config/file, and say to contact their Zulip administrator.

Yeah, that's what Akash and I had considered earlier, but I figured it might be easier for the admins (and certainly the devs) using the app to just get access temporarily to the setting. 
Let me know if you think I should still revert this commit, though.

rishig:
No comment on the dev experience part, but I think the production version of this feature should not have the ability to change the config settings without directly changing the file, since adding a UI for it adds some implicit expectations that will be a big can of worms.

kanishk98:
> since adding a UI for it adds some implicit expectations that will be a big can of worms.

Fair enough, I'll drop the commit. Thanks!

app/renderer/js/main.js
```diff
@@ -159,6 +160,35 @@ class ServerManagerView {
...
 	}
 
+	async queueDomain(domain) {
+		try {
+			const serverConf = await DomainUtil.checkDomain(domain, {override: true});
+			await DomainUtil.addDomain(serverConf);
+			return true;
+		} catch (err) {
+			logger.error(err);
+			logger.error('Could not add ' + domain + '. Please contact your system administrator.');
```
priyank-p: Hmm, do you mean to put this in dialog? Just checking since this is going to log file and this seems more like user-facing error message.

kanishk98: No, I figured that would be annoying, because any domains not added yet and present in the config file will be checked and added, so any incorrect domains will keep opening up a dialog every time the app loads. 
We could probably show a dialog the first time that this fails and then only log the messages from then on. What do you think?

kanishk98: @priyank-p I've added a commit that shows an error dialog when none of the organizations in `presetOrgs` can be added. If you think we need to be more detailed, then let's use this dialog to list all domains that failed to load. 
(If we just replace the `logger` statement with the `dialog`, then every failing org will trigger an error dialog).

app/resources/messages.js
```diff
@@ -22,6 +22,13 @@ class Messages {
 		\nUnless you have a good reason to believe otherwise, you should not proceed.
 		\nYou can click here if you'd like to proceed with the connection.`;
 	}
+
+	failEnterpriseOrgError(length) {
```
priyank-p: `enterpriseOrgError` along sound good to me as the function name here.

kanishk98: Alright, no problem. I'll change that.

app/renderer/css/preference.css
```diff
@@ -453,6 +453,10 @@ i.open-tab-button {
     user-select: none;
 }
 
+input.toggle-disabled + label {
+    cursor: not-allowed;
```
priyank-p, line 458: A good followup for this commit might be showing a tooltip or text showing why it's not allowed.

app/renderer/js/main.js
```diff
@@ -608,7 +608,7 @@ class ServerManagerView {
 							message: 'Are you sure you want to disconnect this organization?'
 						}, response => {
 							if (response === 0) {
-								if (EnterpriseUtil.isAdminOnly('presetOrganizations')) {
+								if (EnterpriseUtil.isAdminOnly('presetOrganizations') && this.presetOrgs.includes(DomainUtil.getDomain(index).url)) {
```
priyank-p: This commit probably needs to be squash with the commit that introduced the feature.

docs/Enterprise.md
```diff
@@ -0,0 +1,26 @@
+# Configuring Zulip Desktop for multiple users
+
+If you're a system admin and want to add certain organizations to the Zulip app for
+all users of your system, you can do so by creating an enterprise config file.
+The file should be placed at `/etc/zulip-desktop-config` for Linux and macOS computers
+and inside `C:\Program Files\Zulip-Desktop-Config` on Windows.
+It must be named `enterprise_config.json` in both cases. 
+
+To specify the preset organization you want to add for other users, you will need to
+add below to the `enterprise_config.json`, replace `https://chat.zulip.org` to the
```
priyank-p:
```suggestion
add the json shown below to the `enterprise_config.json`, replace `https://chat.zulip.org` to the
```

app/renderer/js/main.js
```diff
@@ -153,6 +162,76 @@ class ServerManagerView {
...
+					if (response === 0) {
+						ipcRenderer.send('reload-full-app');
+					}
+				});
+			} else {
+				ipcRenderer.send('reload-full-app');
+			}
+		} else {
+			// no enterprise domain added
+			const { title, content } = Messages.failEnterpriseOrgError(this.presetOrgs.length);
```
priyank-p:
```suggestion
			const { title, content } = Messages.enterpriseOrgError(this.presetOrgs.length);
```

This should be `enterpriseOrgError` since that was added to `Message` in this commit?

app/renderer/js/utils/enterprise-util.js
```diff
@@ -0,0 +1,63 @@
...
+class EnterpriseUtil {
+
+	constructor() {
+		if (instance) {
+			return instance;
+		}
+		instance = this;
+
+		this.reloadDB();
+		return instance;
```
priyank-p: Remove this return statement.
```suggestion
		
```

priyank-p (commented):
@kanishk98 this looks great but posted bunch of comments. Also not that you are missing `.` at the end of commit messages can you also fix that.

kanishk98:
Included changes suggested by @priyank-p. Also removed `isAdminOnly()` requirements as suggested by Tim and Rishi on czo earlier. 

Here's a screenshot of the tooltip suggested: 

Screenshot from 2019-06-29 05-57-42

app/renderer/js/pages/preference/base-section.js
```diff
@@ -18,7 +18,7 @@ class BaseSection extends BaseComponent {
 	}
 
 	generateOptionTemplate(settingOption, disabled) {
-		const label = disabled ? `<label class="disallowed" />` : `<label/>`;
+		const label = disabled ? `<label class="disallowed" title="Setting locked by admin."/>` : `<label/>`;
```
rishig: @timabbott, thoughts on what this string should be? I like the "Setting locked by X" format, but not sure what to call the person doing the locking here.

timabbott: I think "Setting locked by your system administrator." is probably correct, since "system administrator" pretty unambiguously refers to folks with root access on your computer.

kanishk98:
> I think "Setting locked by your system administrator." is probably correct

Great, changed the message. :)

app/renderer/js/pages/preference/base-section.js
rishig: yeah, I like that.

kanishk98:
Converted the branch to TS. Please have a look and tell me if this works well.

akashnimare:
@vsvipul, @priyank-p can you guys review?

app/resources/messages.ts
```diff
@@ -22,6 +22,20 @@ class Messages {
 		\nUnless you have a good reason to believe otherwise, you should not proceed.
 		\nYou can click here if you'd like to proceed with the connection.`;
 	}
+
+	enterpriseOrgError(length: number): any {
```
vsvipul: @kanishk98 You can create an interface here instead of any as you already know that it has title and content only. 
```
interface ErrorType {
    title: string;
    content: string;
}
```
and then use it here, instead of any. Same for orgRemovalError.

docs/Enterprise.md
```diff
@@ -0,0 +1,26 @@
+# Configuring Zulip Desktop for multiple users
+
+If you're a system admin and want to add certain organizations to the Zulip app for
+all users of your system, you can do so by creating an enterprise config file.
+The file should be placed at `/etc/zulip-desktop-config` for Linux and macOS computers
+and inside `C:\Program Files\Zulip-Desktop-Config` on Windows.
+It must be named `enterprise_config.json` in both cases. 
+
+To specify the preset organization you want to add for other users, you will need to
```
vsvipul, line 9: Change to "organizations".

app/renderer/js/utils/enterprise-util.ts
```diff
@@ -0,0 +1,69 @@
...
+
+const logger = new Logger({
+	file: 'enterpise-util.log',
+	timestamp: true
+});
+
+let instance: null | EnterpriseUtil = null;
+
+class EnterpriseUtil {
+	enterpriseSettings: any;
```
vsvipul, line 15: Once the settings we'll allow are finalized, we'll need to remove this any.

app/renderer/js/main.ts
```diff
@@ -93,6 +95,7 @@ class ServerManagerView {
 	tabs: ServerOrFunctionalTab[];
 	functionalTabs: AnyObject;
 	tabIndex: number;
+	presetOrgs: any[];
```
vsvipul: We can create an interface for this too, if we know what it is going to contain.

akashnimare:
@priyank-p, @vsvipul, @kanishk98 let's take care of this PR and merge it asap so that we can include this in the next stable release.

docs/Enterprise.md
priyank-p: Considering the flow of the docs and the example below this should be left as `organization` because we give a example of specifying a single organization.

app/renderer/js/utils/enterprise-util.ts
kanishk98: Yeah, I'll leave a todo comment here for future work.

kanishk98:
@akashnimare made the changes as suggested.

akashnimare (approved):

akashnimare:
> Document the new files made and the process of editing custom settings

@kanishk98 is this done?

kanishk98:
> @kanishk98 is this done?

Yeah, it's at `docs/Enterprise.md`, isn't it?

akashnimare:
@kanishk98 should we show the name of the org here?

image

Like -
```
Could not add the following organizations:
* Name of org
Please contact your system administrator
```

akashnimare:
The PR works as expected on macOS. @priyank-p, @vsvipul, @abhigyank can you folks also give it a try so that we can merge it?

ssvenn:
I think it would be better to look for global_config.json in the directory where the Zulip app is installed instead of hardcoding a path. It would probably be more flexible and it should survive updates as long as the file is only included in the archive as a template like ./examples/global_config.json

priyank-p:
@ssvenn so the reason we decided to choose the path we currently have is because it is easier for administrator to add the config in one place instead of having to do so for every single user. But I suppose it wouldn't be hard to implement support for placing the file in multiple location; @kanishk98 can you implement this it should be relatively easy to implement.

I think it better to look for it in user data directory instead of app directory as @ssvenn suggest because that what other apps do.

akashnimare:
> suppose it wouldn't be hard to implement support for placing the file in multiple locations; 

This might complicate things, IMO should keep only one location. We wanted to keep this file separate with app's file so chose that location. Let's keep it as it is for now and we can always change it later based on more feedbacks etc.

kanishk98:
I agree with @akashnimare here, but I'd love to discuss the multiple locations thing for the next iteration of this feature. 

Are we allowing the user to specify the location for the files themselves or do we add another location on our own in the user directory and give preference to the admin's settings whenever there's a conflict?

ssvenn:
I figured looking for the configuration file in the program directory was just a more common and tidier implementation with one less directory in C:\Program Files\, the ability to work on computers where C:\ isn't the system drive, multiple instances and so on. It's a pretty minor thing and maybe only relevant for Windows. If you end up reading the config from many locations the order of precedence should probably be system directory, program directory, user directory.

docs/Enterprise.md
```diff
@@ -0,0 +1,26 @@
...
+add the `json` shown below to the `enterprise_config.json`, replace `https://chat.zulip.org` to the
```
akashnimare: You need to update the name of this file to `global_config.json`.

docs/Enterprise.md
```diff
@@ -0,0 +1,26 @@
+# Configuring Zulip Desktop for multiple users
+
+If you're a system admin and want to add certain organizations to the Zulip app for
+all users of your system, you can do so by creating an enterprise config file.
+The file should be placed at `/etc/zulip-desktop-config` for Linux and macOS computers
+and inside `C:\Program Files\Zulip-Desktop-Config` on Windows.
+It must be named `enterprise_config.json` in both cases.
```
akashnimare: ^^ `global_config.json`.

akashnimare (changes requested):

app/renderer/js/main.ts
```diff
@@ -232,6 +240,59 @@ class ServerManagerView {
...
+			await DomainUtil.addDomain(serverConf);
+			return true;
+		} catch (err) {
+			logger.error(err);
+			logger.error('Could not add ' + domain + '. Please contact your system administrator.');
+			return false;
+		}
+	}
+
+	async initPresetOrgs(): Promise<void> {
```
akashnimare, line 260: Can you document this function? Add a comment or two.

app/renderer/js/main.ts
```diff
@@ -232,6 +240,59 @@ class ServerManagerView {
 		this.toggleSidebar(showSidebar);
 	}
 
+	async queueDomain(domain: any): Promise<boolean> {
```
akashnimare, line 245: Please document this.

akashnimare (changes requested):

app/renderer/js/utils/enterprise-util.ts
```diff
@@ -0,0 +1,70 @@
+import fs = require('fs');
+import path = require('path');
+
+import Logger = require('./logger-util');
+
+const logger = new Logger({
+	file: 'enterpise-util.log',
```
akashnimare: Typo - `enterpise`?

app/resources/messages.ts
```diff
@@ -22,6 +27,20 @@ class Messages {
 		\nUnless you have a good reason to believe otherwise, you should not proceed.
 		\nYou can click here if you'd like to proceed with the connection.`;
 	}
+
+	enterpriseOrgError(length: number): DialogBoxError {
+		return {
+			title: `Could not add ${length} ${length === 1 ? `organization` : `organizations`}`,
```
akashnimare: @kanishk98 can you update this as we discussed?

kanishk98: This is what it looks like now:

image

Just checking a few things to make sure there are no problems with this. Will push in a bit.

kanishk98: I'll remove the length, that's unnecessary.

akashnimare:
As of now, the error message looks like this -

image

We also need to add the org url here.

kanishk98:
Works fine for me: 

image

Try running `npm run clean-ts-files` first? Not doing so often doesn't reflect changes quickly for me.

akashnimare (approved):

vsvipul (approved):
Works fine for me. The app fails to even load if a wrong format is used for the config file. Leaving that, everything works fine.

akashnimare:
@kanishk98 great work 👌  This was the most demanded feature, glad we finished it. Thanks, everyone for the testing, inputs and reviews. @rishig would be great if you can look at the docs when you get time. @timabbott FYI.

PR #685, opened 2019-03-14
Allow insecure requests on user approval.

### LINKED ISSUE #684: Allow insecure requests when user chooses to ignore cert warnings
In certain requests like for fetching realm icons, `request` tries to verifies server certificates and fails when the user has chosen to ignore cert warnings and has not uploaded the cert.
In those cases we should mark that this server verification is ignored and ask `request` to ignore the cert verification by passing that parameter.

vsvipul:
**What's this PR do?**
Fixes #684 

**To test**
1. Comment out "zulipdev.org" in domain-util.js
2. Try adding abhigyank.zulipdev.org to realms.
3. The realm icon and name loads perfectly.

**Screenshots?**
Screenshot from 2019-03-16 15-09-22

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/utils/domain-util.js
+++ b/app/renderer/js/utils/domain-util.js
@@ -64,9 +64,10 @@ class DomainUtil {
 	}
 
 	addDomain(server) {
+		const ignoreCerts = server.ignoreCerts;
 		return new Promise(resolve => {
 			if (server.icon) {
-				this.saveServerIcon(server).then(localIconUrl => {
+				this.saveServerIcon(server, ignoreCerts).then(localIconUrl => {
 					server.icon = localIconUrl;
 					this.db.push('/domains[]', server, true);
 					this.reloadDB();
@@ -103,7 +104,9 @@ class DomainUtil {
 		return false;
 	}
 
-	checkDomain(domain, silent = false) {
+	// ignoreCerts parameter helps in fetching server icon and
+	// other server details when user chooses to ignore certificate warnings
+	checkDomain(domain, ignoreCerts = false, silent = false) {
 		if (!silent && this.duplicateDomain(domain)) {
 			// Do not check duplicate in silent mode
 			return Promise.reject('This server has been added.');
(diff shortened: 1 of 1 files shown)


app/renderer/js/utils/domain-util.js
```diff
@@ -270,7 +272,8 @@ class DomainUtil {
 			ca: (certificateLocation) ? certificateLocation : '',
 			proxy: proxyEnabled ? ProxyUtil.getProxy(url) : '',
 			ecdhCurve: 'auto',
-			headers: { 'User-Agent': SystemUtil.getUserAgent() }
+			headers: { 'User-Agent': SystemUtil.getUserAgent() },
+			rejectUnauthorized: (server.ignoreCerts) ? !server.ignoreCerts : true
```
kanishk98: Wouldn't it be more readable to just write `rejectUnauthorized: !server.ignoreCerts`? Spent a couple of minutes more than I should have on this. :P

vsvipul: Thanks for your review @kanishk98 . That is done because previous organizations already added into domains.json file would not have the attribute ignoreCerts, and assigning rejectUnauthorized an unknown value can cause problems, as the function checkDomain() is called every time the application restarts. Thanks.

vsvipul: The PR is still WIP btw.

abhigyank:
@vsvipul  Thanks for working on this. I tested it, it works fine. 
I think it could be good a couple of comments explaining the need of the `ignoreCerts` parameters in some function since the some calls can be a bit complicated to someone not involved to understand.

vsvipul:
@abhigyank i have added the comments. Thanks.

akashnimare:
@vsvipul I'm guessing this won't work on already added servers. Is there anything we could do for folks who have already added this kind of servers?

vsvipul:
@akashnimare Unless they remove and add the realm and accept the ignore certificates warning, we can't change it. Or we can make the default value of ignoreCerts to be 'true', but i don't think that will be viable as it can compromise security.

akashnimare:
Merged. Thanks, @vsvipul and @abhigyank.

abhigyank:
> Is there anything we could do for folks who have already added this kind of servers? 

I'd really not want to do that as in those cases if a server turns bad cert after user adds it during cert it'd continue to be treated as of the user accepted using bad cert.

PR #690, opened 2019-03-28
system-presence: Add ability to detect system presence.

vsvipul:
**What's this PR do?**
Created two electron-bridge variables.
idle_on_system tells whether the user is idle or not.
last_active_on_system gives the time the user was last
active. Both variables will be used by the webapp to set
user status.

Fixes #352.

Corresponding webapp PR [here](https://github.com/zulip/zulip/pull/12003).

**How to test?**
You can test by console logging the two variables at line 97 of preload.js. They will update based on user activity.

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -344,6 +344,18 @@ app.on('ready', () => {
 	ipcMain.on('realm-icon-changed', (event, serverURL, iconURL) => {
 		page.send('update-realm-icon', serverURL, iconURL);
 	});
+
+	// Update user idle status for each realm
+	// Set user idle if no activity for 60s
+	ipcMain.on('detect-presence', event => {
+		electron.powerMonitor.querySystemIdleState(60, idleState => {
+			if (idleState === 'active') {
+				event.sender.send('set-active');
+			} else {
+				event.sender.send('set-idle');
+			}
+		});
+	});
 });
 
 app.on('before-quit', () => {
--- a/app/renderer/js/preload.js
+++ b/app/renderer/js/preload.js
@@ -7,14 +7,24 @@ const ConfigUtil = require(__dirname + '/utils/config-util.js');
 const LinkUtil = require(__dirname + '/utils/link-util.js');
 const params = require(__dirname + '/utils/params-util.js');
 
+// we have and will have some non camelcase stuff
+// while working with zulip and electron bridge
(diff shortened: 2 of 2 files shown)


app/renderer/js/electron-bridge.js
```diff
@@ -32,6 +32,10 @@ electron_bridge.on('realm_icon_url', iconURL => {
 	ipcRenderer.send('realm-icon-changed', serverURL, iconURL);
 });
```
akashnimare: Maybe add a few comments here about what's going on so that others can also understand?

app/renderer/js/electron-bridge.js
```diff
@@ -32,6 +32,10 @@ electron_bridge.on('realm_icon_url', iconURL => {
...
 
+electron_bridge.on('desktop_presence', timeOut => {
```
akashnimare: Maybe add a few comments here about what's going on so that others can also understand?

akashnimare (changes requested):

abhigyank:
Lets wait on this to first figure out the webapp par in https://github.com/zulip/zulip/pull/12003 . 
For now, in this PR `console.log` part is working correctly for me as expected.

app/main/index.js
```diff
@@ -344,6 +344,18 @@ app.on('ready', () => {
 	ipcMain.on('realm-icon-changed', (event, serverURL, iconURL) => {
 		page.send('update-realm-icon', serverURL, iconURL);
 	});
+
+	// Update user idle status for each realm
+	// Set user idle if no activity for 60s
+	ipcMain.on('detect-presence', event => {
+		electron.powerMonitor.querySystemIdleState(60, idleState => {
```
abhigyank, line 351: We should put the value `60` in a variable like `idleThreshhold`.

abhigyank:
@**Vipul Sharma**  the desktop PR implementation looks appropriate to me. It is working as expected.
LGTM.

app/main/index.js
vsvipul: sure, will do that.

vsvipul:
Opened in a new PR [here](https://github.com/zulip/zulip-desktop/pull/748).

Issue #675, opened 2019-03-04
The initial character we display has wrong first character


priyank-p:
Right now it seems like we are displaying `h` for chat.zulip.org instead of `Z`. The bug here seems to be that we are using the url instead of organization name. The place we get the organization name is the `name` attribute in the `<webview>` element which is passed in from the [`WebView` component](https://github.com/zulip/zulip-electron/blob/master/app/renderer/js/components/webview.js#L36).

abhigyank:
We do get the name itself and not the url. I think the issue here is that, when we don't the realm_image, we do not get the realm name as well and we then default the organization name to the url which is prefixed with https and hence is shown as h.

rhythm-sharma:
@zulipbot claim

zulipbot:
Hello @rrhythmsharma!

Thanks for your interest in Zulip! You have attempted to claim an issue without the labels "help wanted", "good first issue". Since you're a new contributor, you can only claim and submit pull requests for issues with the [help wanted](https://github.com/zulip/zulip-electron/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22help+wanted%22) or [good first issue](https://github.com/zulip/zulip-electron/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22good+first+issue%22) labels.

If this is your first time here, we recommend reading our [guide for new contributors](https://zulip.readthedocs.io/en/latest/overview/contributing.html) before getting started.

vsvipul:
@zulipbot claim

vsvipul:
@zulipbot abandon

abhigyank:
@zulipbot claim

zulipbot:
Hello @abhigyank, you have been unassigned from this issue because you have not updated this issue or any referenced pull requests for over 14 days.

You can reclaim this issue or claim any other issue by commenting `@zulipbot claim` on that issue.

Thanks for your contributions, and hope to see you again soon!

kanishk98:
@zulipbot claim

kanishk98:
Edit: I'd earlier asked for help on reproducing the issue, but I think I got it: the app is not able to get the realm_name without access to the internet and in some cases, the `domain.json` file is not updated with the correct `realm_name`. 

To fix this, I've simply taken in the first word following the `https://` part of the server URL and used that as a "trimmed domain" for the app to take the first character from. You may see an example below. 

[...]

PR #693, opened 2019-03-31
Trim domain to first word in server URL

kanishk98:
If the app is unable to fetch the realm name, the server alias will now
default to first word after the protocol scheme in the server URL.

---

**What's this PR do?**

Changes `serverConf` object so that if the app is unable to save the correct realm name, the sidebar receives the first character of the URL after `://`. 

Discussion at #675 

**Screenshots?**

54884441-20dea080-4e97-11e9-9c7d-ccbf4e32d54f

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/utils/domain-util.js
+++ b/app/renderer/js/utils/domain-util.js
@@ -104,6 +104,12 @@ class DomainUtil {
 		return false;
 	}
 
+	trimDomain(domain) {
+		let subDomain = domain.substring(domain.lastIndexOf('://') + 3);
+		subDomain = subDomain.substring(0, subDomain.indexOf('.'));
+		return subDomain;
+	}
+
 	// ignoreCerts parameter helps in fetching server icon and
 	// other server details when user chooses to ignore certificate warnings
 	checkDomain(domain, ignoreCerts = false, silent = false) {
@@ -142,7 +148,7 @@ class DomainUtil {
 		const serverConf = {
 			icon: defaultIconUrl,
 			url: domain,
-			alias: domain,
+			alias: this.trimDomain(domain),
 			ignoreCerts
 		};
 
--- a/app/renderer/js/utils/domain-util.js
+++ b/app/renderer/js/utils/domain-util.js
@@ -105,6 +105,11 @@ class DomainUtil {
 	}
 
 	trimDomain(domain) {
+		/*
+			sometimes, the app is unable to fetch realm_name
+			we then show the first letter character after the protocol scheme
+			instead of the first character of the realm_name
+		*/
 		let subDomain = domain.substring(domain.lastIndexOf('://') + 3);
 		subDomain = subDomain.substring(0, subDomain.indexOf('.'));
 		return subDomain;


akashnimare:
Would it make sense to put the `trimDomain` function in the new `request-util.js` ?

vsvipul (commented):
The function trimDomain() works as expected. It extracts abc from https://abc.zulipchat.com or zulip from http://zulip.orgname.com. We still don't get the org name in the second case, where user accesses a self-hosted server. For rest of the cases, works fine :)

vsvipul:
> Would it make sense to put the `trimDomain` function in the new `request-util.js` ?

No, it wouldn't i guess. The request-util returns header options for all requests in domain-util, and does not return alias name.

app/renderer/js/utils/domain-util.js
```diff
@@ -104,6 +104,12 @@ class DomainUtil {
 		return false;
 	}
 
+	trimDomain(domain) {
```
akashnimare, line 107: Can you add a comment here about what we are doing here?

akashnimare (changes requested):

abhigyank:
There are three cases of url here :-
*  abc.domain.com - Show a
* zulip.domain.com / server1.zulip.domain.com - Show d / show s
* domain.com - show d (very rare case)

It is easy for us to implement for the first case as done in this case. The problem arises when we want to distinguish between first and second case which I feel is impossible for us to do. I am not sure if showing z in case of 2nd case an improvement from the the current implementation of showing the defaultIcon or is infact an increment of poor UX. Really unsure as to what the way forward should be here in this case.

kanishk98:
> zulip.domain.com / server1.zulip.domain.com - Show d / show s

@abhigyank correct me if I'm wrong, but the problem here is just the word `zulip` in the server URL, right?
If that's the case, we can check if the first word of the URL after the protocol scheme starts with `zulip`. If so, then we assume that the URL is of the form zulip.domain.com and show d. 
The current case should show s in case of server1.zulip.domain.com, and I'm guessing that's acceptable. (If not, we can simply check for `zulip` and find the next immediate word that exists. Of course, we'll have to make sure we don't overrun case 1 when doing this).

kanishk98:
Please let me know if any further changes are required here.

abhigyank:
>  but the problem here is just the word zulip in the server URL, right?
If that's the case, we can check if the first word of the URL after the protocol scheme starts with zulip. If so, then we assume that the URL is of the form zulip.domain.com and show d.

I don't want us to hard-code here (I am never in favour of hard-coding anything! :sweat_smile: ) The reason being we don't want to assume that its always zulip in such a subdomain. I am not sure what is the right way to go here. :confused:

kanishk98:
I agree, hard coding is a pretty hacky solution. I’m also unsure what we
should do. :(

kanishk98:
@rishig would you mind having a look at this whenever you're free and telling us what you think?

rishig:
I think we should just show some kind of error state if the realm doesn't have an avatar set (e.g. a box with a `?`).

kanishk98:
> I think we should just show some kind of error state if the realm doesn't have an avatar set (e.g. a box with a ?).

I think that'll be a good solution when we don't have this kind of behaviour come up frequently (which is the case right now). When @akashnimare is able to find time to review #714 and we cut down on the number of times this fallback icon needs to be used, we can use the `?`. Sound good?

Issue #676, opened 2019-03-06
Implement HTML and CSS Linting 


vsvipul:
I recently found out that HTML and CSS files are not linted by xo (linter). 

I think it would be better if we could use a plugin to lint html files. 
If can't find any html plugin with xo, maybe we can use separate linters for HTML and CSS.

- [x] Implement HTML linting
- [x] Implement CSS linting

vsvipul:
@zulipbot claim

zulipbot:
Hello @vsvipul, you have been unassigned from this issue because you have not updated this issue or any referenced pull requests for over 14 days.

You can reclaim this issue or claim any other issue by commenting `@zulipbot claim` on that issue.

Thanks for your contributions, and hope to see you again soon!

vsvipul:
sorry, i have been busy due to my midsems. i'll continue working from tomorrow. @zulipbot

vsvipul:
@zulipbot add "enhancement"

zulipbot:
**ERROR:** Label "enhancement" does not exist and was thus not added to this issue.

vsvipul:
@zulipbot add "Type: Enhancement"

vsvipul:
@zulipbot claim

zulipbot:
Hello @vsvipul, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

vsvipul:
@zulipbot Done with CSS, working with HTML :)

PR #699, opened 2019-04-03
lint: Implement CSS linting with stylelint

vsvipul:
Adds linting to CSS files using stylelint.
Closes the CSS part of #676.

**What's this PR do?**
Implements CSS linting using [stylelint](https://github.com/stylelint/stylelint).
Format CSS file according to zulip webapp's [lint config](https://github.com/zulip/zulip/blob/master/.stylelintrc).

Some options have been excluded from lint config which are-
1. "color-hex-case": "lower"
2. "selector-pseudo-element-colon-notation": "double"
3. "color-no-hex": true
4. "color-named": "never"

These rules required changing more than just spaces and tabs, so i excluded them. I intend to add them separately later after this PR is merged.

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "start": "electron app --disable-http-cache --no-electron-connect",
     "reinstall": "node ./tools/reinstall-node-modules.js",
     "postinstall": "electron-builder install-app-deps",
-    "test": "xo",
+    "test": "stylelint app/renderer/css/*.css && xo",
     "test-e2e": "gulp test-e2e",
     "dev": "gulp dev & nodemon --watch app/main --watch app/renderer --exec 'npm test' -e html,css,js",
     "pack": "electron-builder --dir",
@@ -132,6 +132,7 @@
     "nodemon": "^1.14.11",
     "pre-commit": "1.2.2",
     "spectron": "^5.0.0",
+    "stylelint": "^9.10.1",
     "tap-colorize": "^1.2.0",
     "tape": "^4.8.0",
     "xo": "0.18.2"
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
     "start": "electron app --disable-http-cache --no-electron-connect",
     "reinstall": "node ./tools/reinstall-node-modules.js",
     "postinstall": "electron-builder install-app-deps",
-    "test": "xo",
+    "test": "xo && stylelint app/renderer/css/*.css",
```
kanishk98: I think it's important to Windows users that `xo` is run at the end. That test takes care of line endings, which can mess up quite often and I think they're taken care of by post-commit git hooks anyway. So if `xo` fails because of line endings, `stylelint` won't run and we don't want that.

package.json
```diff
@@ -179,5 +180,14 @@
...
     ]
+  },
+  "stylelint": {
+    "rules": {
+      "block-no-empty": null,
+      "color-no-invalid-hex": true,
+      "declaration-colon-space-after": "always",
+      "indentation": "tab",
+      "max-empty-lines": 2
+    }
```
kanishk98: These rules look good to me, just asking to confirm if there's been a discussion with the maintainers regarding them.

kanishk98 (approved):
Looks good to me. Fails and passes appropriately based on my testing. (Please do fix the Windows issue though).

package.json
vsvipul: I based the indentation rules on zulip webapp. I also intend to use the rules from the webapp once, these basic rules get merged.

package.json
vsvipul: hmm... If stylelint fails, would xo run ?

vsvipul:
> Looks good to me. Fails and passes appropriately based on my testing. (Please do fix the Windows issue though).

@kanishk98 Thank u so much for your review. Which windows issue are you referring to ?

package.json
kanishk98: No, but then the code has actually failed the test. In the present case, even correctly styled code (which xo says is wrong but passes Travis tests) will not go through stylelint. So I think we should change the order.

kanishk98:
> Which windows issue are you referring to ?

The one you've already addressed in the comments. :) [This one](https://github.com/zulip/zulip-electron/pull/699#discussion_r271947224).

package.json
vsvipul: ok got it.. will do.. thanks :)

akashnimare:
Okay, boys - one suggestion. Since we have two linters now we could create a new script like -
```json
"lint": "stylelint app/renderer/css/*.css && xo",
"test: "lint"
```

abhigyank:
Should we put the config in a `.stylelintrc` file or is it fine in the package.json?

vsvipul:
@akashnimare i feel something like this will be better-
```
"lint-css": "stylelint app/renderer/css/*.css",
"lint-js": "xo",
"test: "lint-css && lint-js"
```
We can add `lint-html` later on .

vsvipul:
@abhigyank we can use a separate file but there was no `.eslintrc` for xo, so i did not create one for stylelint as well.

timabbott:
We should use a `.stylelintrc`.  Might be worth grabbing the config from the Zulip webapp so we use a consistent style.

akashnimare:
@vsvipul can you create a `.eslintrc` for xo and .`stylelintrc` for stylelint?

vsvipul:
> ```
> "lint-css": "stylelint app/renderer/css/*.css",
> "lint-js": "xo",
> "test: "lint-css && lint-js"
> ```
> We can add `lint-html` later on .

@akashnimare yeah. i'll work on adding those.. what do you think about the above package.json commands?

akashnimare:
Looks good for now.

vsvipul:
@akashnimare xo doesn't support a separate config file for now. It will add support for it later. Issue [here](https://github.com/xojs/xo/issues/330). I have created a separate .stylelintrc with similar config as zulip/zulip (remaining commands in top of this PR, will add them later, as they require some changes in code) and also changed the npm commands. Please review.

vsvipul:
@akashnimare a reminder for this.

akashnimare:
Merged. Thanks, everyone!

PR #714, opened 2019-04-07
Revert to fallback icon only when needed

kanishk98:
We currently check if the server icon has updated everytime the app
loads. If a network error happens and the app can't fetch this icon,
then the character icon is loaded next time even though the original
server icon was saved in user's app data. This commit forces the app to
consider the original icon in such cases.

---

Discussion [here](https://chat.zulip.org/#narrow/stream/16-desktop/topic/realm.20icon.20in.20dock/near/716234) and [here](https://chat.zulip.org/#narrow/stream/16-desktop/topic/realm-icon). 

**Screenshots?**

image

**Testing procedure**

Load the app once. Make sure the icons render properly in the sidebar. Then go to `domain-util` and change the `saveServerIcon()` method to something similar:

```javascript
saveServerIcon(server, ignoreCerts = false) {
	const url = 'https://fsjfbsk.com';
	const domain = server.url;
```

Reload the app a couple of times. The server icon should stay the same even though the URL is incorrect.

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/utils/domain-util.ts
+++ b/app/renderer/js/utils/domain-util.ts
@@ -247,12 +247,15 @@ class DomainUtil {
 
 	updateSavedServer(url: string, index: number): void {
 		// Does not promise successful update
+		const oldIcon = this.getDomain(index).icon;
 		const { ignoreCerts } = this.getDomain(index);
 		this.checkDomain(url, ignoreCerts, true).then(newServerConf => {
 			this.saveServerIcon(newServerConf, ignoreCerts).then(localIconUrl => {
-				newServerConf.icon = localIconUrl;
-				this.updateDomain(index, newServerConf);
-				this.reloadDB();
+				if (!oldIcon || localIconUrl !== '../renderer/img/icon.png') {
+					newServerConf.icon = localIconUrl;
+					this.updateDomain(index, newServerConf);
+					this.reloadDB();
+				}
 			});
 		});
 	}


abhigyank:
@akashnimare  do you feel the solution approach to the issue is correct? I don't feel this is what we should be doing here to solve this issue.

timabbott:
I'm a bit confused as to what this code is doing.  I think the correct behavior is that if we get an error trying to fetch the icon provided by the server, we should keep the old icon, but if we don't get an error, we should replace the icon.

kanishk98:
>  I think the correct behavior is that if we get an error trying to fetch the icon provided by the server, we should keep the old icon, but if we don't get an error, we should replace the icon.

That's what we're doing here. I can explain:

Every time the app loads the renderer process, `main.js` calls `updateSavedServer(servers[i].url, i)` from the `initTabs()` method. 

The `updateSavedServer()` method calls the `checkDomain()` method and then uses the resolved server config from there to call `saveServerIcon()`, which is responsible for fetching the new server icon and saving that in the user's app data folder. 
If the config was incorrect or the network fails when `saveServerIcon()` is running, it encounters an error and resolves with the default `localIconUrl`, which is our character icon. 

We want the server icon to be updated in 2 cases: when the fetched icon is not the character icon and when the server is loaded the first time. My code takes care of that.

kanishk98:
@akashnimare would you mind having a look at this whenever you find time?

akashnimare:
> If a network error happens and the app can't fetch this icon,
then the character icon is loaded next time even though the original
server icon was saved in user's app data. 

We show the character icon whenever an error (network/proxy etc) occurs -
```

if ($serverImg.src.includes('img/icon.png')) {   				 
  this.displayInitialCharLogo($serverImg, index);
}			
$serverImg.addEventListener('error', () => {
    this.displayInitialCharLogo($serverImg, index);
});
```

kanishk98:
> We show the character icon whenever an error (network/proxy etc) occurs -

I'm unsure of whether you're saying we should not have this fix in the first place or if you're questioning if the fix really does work, so I'll just answer both:

> Having this fix:

We've discussed this in multiple places, most notably [here](https://chat.zulip.org/#narrow/stream/16-desktop/topic/realm.20icon.20in.20dock/near/716234) - people tend to get confused when there's a character icon, and it does come up when they reload the app quickly. 

> Whether the character icon shows up on network error:

If the user's app data folder has the correct image location, then it won't raise an error because it'll just return the locally saved icon, right? Attaching a gif below with no network, app shut down:

fix

kanishk98:
Fixed merge conflicts, ready to merge.

akashnimare:
Merged. Thanks, @kanishk98.

Issue #689, opened 2019-03-25
Menubar text colour illegible -- should respect system theme


skewty:
---

- **Operating System**:
  - [ ] Windows
  - [x] Linux/Ubutnu
  - [ ] macOS
- **Clear steps to reproduce the issue**:
1. Install Ubuntu 19.04
2. Leave the theme as default `Yaru`
    * The menubar font color seems to be legible if `Yaru-dark` theme is selected.
3. Install Zulip 2.4.0 from PPA
- **Relevant error messages and/or screenshots**:
![Screenshot from 2019-03-25 15-40-56](https://user-images.githubusercontent.com/9087223/54958902-a7859300-4f14-11e9-9366-c8cec459ea66.png)

akashnimare:
@vsvipul, @kanishk98 can you folks reproduce this?

kanishk98:
Don't have immediate access to a Ubuntu system right now, but will check it out as soon as I can.

vsvipul:
Has been illegible for me as well, since the beginning.. I am on Ubuntu 18.04 . Though @skewty 's menu bar seems to be more illegible.
![Screenshot from 2019-03-26 20-34-36](https://user-images.githubusercontent.com/21038781/55008415-fc093b00-5006-11e9-8d84-d937ac80f050.png)

kanishk98:
Seems like a consistent problem. 
If we can figure out a way to change the colouring of the menu bar, I remember checking out [this module](https://www.npmjs.com/package/@jakejarrett/gtk-theme) for use in Zulip on Linux to determine the background colour of the theme and changing colours accordingly. Could this method be useful?

(I don't think Electron offers a native API to change the appearance of the title bar though, so that would have to be designed by us if we're choosing the CSS route)

akashnimare:
As far as I know, there is an API available for macOS to detect the theme but sadly no APIs for Linux/Windows. @skewty can you reproduce this issue in other apps which are based on electron?

vsvipul:
@akashnimare upgrading the electron version to 3.1.6 fixed this for me as shown. Can we bump up the version, should i send in a PR? 
![Screenshot from 2019-03-28 02-26-11](https://user-images.githubusercontent.com/21038781/55111775-20e1d900-5101-11e9-90c8-5cdc82123829.png)

P.S. this was fixed [here](https://github.com/electron/electron/issues/15823), and thus requires updating the electron version, as this might not be backported to current version 3.0.1, as far as i could find.

[...]

PR #715, opened 2019-04-07
menu: Reorder file menu and add option to Add Organization.

vsvipul:
**What's this PR do?**
Reorder the File Menu. Also, adds a new option to
add a new organization to the file menu.

Changes made according to the discussion [here](https://chat.zulip.org/#narrow/stream/16-desktop/topic/Zulip.20URL).

Someone needs to test this on macOS.

**Screenshots?**
Screenshot from 2019-04-07 21-30-07

Separators aren't visible on my system because of issue [here](https://github.com/zulip/zulip-electron/issues/689).

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/menu.js
+++ b/app/main/menu.js
@@ -259,6 +259,21 @@ class AppMenu {
 		return [{
 			label: `${app.getName()}`,
 			submenu: [{
+				label: 'Add Organization',
+				accelerator: 'Cmd+Shift+N',
+				click(item, focusedWindow) {
+					if (focusedWindow) {
+						AppMenu.sendAction('new-server');
+					}
+				}
+			}, {
+				label: 'Toggle Do Not Disturb',
+				accelerator: 'Cmd+Shift+M',
+				click() {
+					const dndUtil = DNDUtil.toggle();
+					AppMenu.sendAction('toggle-dnd', dndUtil.dnd, dndUtil.newSettings);
+				}
+			}, {
 				label: 'Desktop Settings',
 				accelerator: 'Cmd+,',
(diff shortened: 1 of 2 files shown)


abhigyank:
cc @rishig

rishig:
looks great! We should use `Ctrl+,` rather than `Ctrl+comma`, I think? 
It's what other products seem to do e.g. for `Ctrl++`, and also means we don't have to worry about i18n.

vsvipul:
@rishig i have indeed used 'Ctrl+,' everywhere in the PR, electron is showing it as 'Ctrl+comma' (as shown in the screenshot).

rishig:
Interesting! Probably worth spending a few minutes seeing if we can change that behavior in electron, and if we can't that's fine.

akashnimare:
https://github.com/electron/electron/issues/10371

vsvipul:
This issue has not being fixed yet in https://github.com/electron/electron/issues/10371 . So, we can't change it i guess. VS Code mentions [Ctrl+,] in their label instead of using an accelerator. I am not sure if something like this is what we want @rishig @akashnimare 
Screenshot from 2019-04-09 01-28-23

rishig:
cool, thanks for looking into it. I just noticed that maybe "Log Out Of Organization" might look better as "Log Out of Organization"? 
Other than that, lgtm!

vsvipul:
cool .. changed it . @rishig @akashnimare

akashnimare:
Merged. Thanks, everyone.

PR #716, opened 2019-04-08
[WIP] Add zulip:// URI scheme for navigating within app

kanishk98:
---

**What's this PR do?**

Registers links starting with zulip:// to open with the Zulip app. Redirects users with installed app to the correct narrow and server (only if the domain is already saved) in the desktop app.

Discussion at #470.

**Screenshots?**

deeplinking

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [x] macOS

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -27,7 +27,6 @@ if (isDev) {
 let mainWindow;
 let badgeCount;
 let deepLinkingUrl;
-let execPath;
 
 let isQuitting = false;
 
@@ -41,7 +40,7 @@ if (singleInstanceLock) {
 		if (process.platform === 'win32') {
 			deepLinkingUrl = argv.slice(1);
 			if (mainWindow) {
-				mainWindow.webContents.send('deep-linking-url', deepLinkingUrl);
+				mainWindow.webContents.send('deep-linking-url', deepLinkingUrl[isDev ? 1 : 0]);
 			}
 		}
 
@@ -139,10 +138,11 @@ function createMainWindow() {
 // Decrease load on GPU (experimental)
 app.disableHardwareAcceleration();
 
-if (process.platform === 'win32') {
-	execPath = 'C:\\Program Files\\Zulip\\Zulip.exe';
+if (process.platform === 'win32' && isDev) {
+	app.setAsDefaultProtocolClient('zulip', process.execPath, [path.resolve(process.argv[1])]);
+} else {
+	app.setAsDefaultProtocolClient('zulip');
 }
-app.setAsDefaultProtocolClient('zulip', execPath);
 
 // uri scheme handler for macOS
 app.on('open-url', (event, url) => {
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -39,10 +39,7 @@ if (singleInstanceLock) {
 		// uri scheme handler for windows and linux
 		if (process.platform !== 'darwin') {
 			deepLinkingUrl = argv.slice(1);
-			if (mainWindow) {
-				mainWindow.webContents.focus();
-				mainWindow.webContents.send('deep-linking-url', deepLinkingUrl[(isDev && process.platform === 'win32') ? 1 : 0]);
-			}
+			handleDeepLink(deepLinkingUrl[(isDev && process.platform === 'win32') ? 1 : 0]);
 		}
 
 		if (mainWindow) {
@@ -62,6 +59,13 @@ const APP_ICON = path.join(__dirname, '../resources', 'Icon');
 const iconPath = () => {
 	return APP_ICON + (process.platform === 'win32' ? '.ico' : '.png');
 };
+
+function handleDeepLink(deepLinkingUrl) {
+	if (mainWindow) {
+		mainWindow.webContents.focus();
+		mainWindow.webContents.send('deep-linking-url', deepLinkingUrl);
+	}
+}
 
 function createMainWindow() {
 	// Load the previous state with fallback to defaults
@@ -211,10 +215,7 @@ app.on('ready', () => {
 
 	if (process.platform !== 'darwin') {
 		deepLinkingUrl = process.argv.slice(1);
-		if (mainWindow) {
-			mainWindow.webContents.focus();
-			mainWindow.webContents.send('deep-linking-url', deepLinkingUrl[(isDev && process.platform === 'win32') ? 1 : 0]);
-		}
+		handleDeepLink(deepLinkingUrl[(isDev && process.platform === 'win32') ? 1 : 0]);
 	}
 
 	// Temporarily remove this event
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -36,11 +36,12 @@ const mainURL = 'file://' + path.join(__dirname, '../renderer', 'main.html');
 const singleInstanceLock = app.requestSingleInstanceLock();
 if (singleInstanceLock) {
 	app.on('second-instance', (event, argv) => {
-		// uri scheme handler for windows
-		if (process.platform === 'win32') {
+		// uri scheme handler for windows and linux
+		if (process.platform !== 'darwin') {
 			deepLinkingUrl = argv.slice(1);
 			if (mainWindow) {
-				mainWindow.webContents.send('deep-linking-url', deepLinkingUrl[isDev ? 1 : 0]);
+				mainWindow.webContents.focus();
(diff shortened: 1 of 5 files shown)


andersk:
Is this compatible with the Zulip Android app, which also uses `zulip://` URLs? If we’re going to do this, we should make sure to keep them compatible.

kanishk98:
> If we’re going to do this, we should make sure to keep them compatible.

@andersk can you explain what you mean by compatible here?

timabbott:
The mobile app's existing use of zulip:// URLs is only for an authentication flow.  

But yes, we should coordinate with the mobile team on what URL syntax we want to use for this because the two apps will end up sharing that syntax.

kanishk98:
For the benefit of everyone in this thread, I texted Boris on czo yesterday and asked him to have a look at this PR. Waiting for his suggestions. :)

app/main/index.js
```diff
@@ -129,6 +139,20 @@ function createMainWindow() {
 // Decrease load on GPU (experimental)
 app.disableHardwareAcceleration();
 
+if (process.platform === 'win32') {
+	execPath = 'C:\\Program Files\\Zulip\\Zulip.exe';
```
vsvipul: what if the user installs Zulip at a different path?

kanishk98: > (I've hard-coded the path to the Windows executable here, please let me know if there's a better way to handle this)

That's what I've asked about here, I didn't look into getting the install path yet. Will get to it in an hour or so. Working on another PR.

kanishk98: Also, Vipul, it'd be great if you could check out this PR locally and test it on your system. I'm guessing Linux might have some issues with this sort of feature. (there were some concerns raised in the upstream repo regarding the method I've used).

vsvipul: Sure, will check till evening :)

vsvipul: How do i test this on Ubuntu? Simply prefixing zulip:// to normal URLs doesn't seem to be working. Anything i am missing here?

kanishk98: Don't think so - this method works without any additional work on macOS and Windows. 
Maybe the desktop file needs to be configured properly. Let me check. I'll bother you on czo with requests for testing later tonight. :P

vsvipul: no problem. i'm here.

akashnimare:
I don't think we need to call the `execPath`. Also, you need to register the app's protocol in the package.json. 
https://github.com/oikonomopo/electron-deep-linking-mac-win
https://stackoverflow.com/questions/43912119/open-app-and-pass-parameters-with-deep-linking-using-electron-macos
https://discuss.atom.io/t/enable-deep-linking/27518/7

kanishk98:
> I don't think we need to call the execPath.

Right, I wasn't sure about this earlier. We should note that this will raise an error in development mode when a developer tries to use this feature because then if Zulip isn't installed the conventional way, it won't find an executable. (happened to me while testing)
I ran `npm run dist` and worked fine. I'll remove execPath. 

> Also, you need to register the app's protocol in the package.json.

I don't see the need - the system registers the app's protocol regardless, if we just use the current method. Can you explain why this is required if the feature works both on `npm start` and `npm run dist`?

akashnimare:
> I don't see the need - the system registers the app's protocol regardless, if we just use the current method.

I think this is required for macOS.

https://stackoverflow.com/questions/45809064/registering-custom-protocol-at-installation-process-in-electron-app

kanishk98:
> I think this is required for macOS.

I checked it on a Mac in debug mode, without creating an executable. It worked well then. 
I'll check once more to make sure, and also create an executable for testing. 

I'll post back later tonight (I'll have to borrow a Mac, can't do that immediately)

kanishk98:
> I'll post back later tonight

I checked this on a Mac and the protocol is registered, but since the `open-url` event passes the URL as a string rather than as an array (which is the case on Windows), redirection doesn't happen on macOS. Simple enough problem, I'll push a fix.

kanishk98:
Alright, I've done a couple of things. 
1. Made sure that the scheme works for Windows, both in dev and production. (the API needs different arguments in different cases for Windows)
2. Fixed macOS redirection problems.
3. Added a reload view to make redirection more obvious.

@akashnimare let me know if it works for you now.

abhigyank:
Haven't tested this yet, just out of curiosity from the gif, does this work if the app is not currently running.

kanishk98:
> Haven't tested this yet, just out of curiosity from the gif, does this work if the app is not currently running.

It does, tested that. :)

app/main/index.js
```diff
@@ -34,7 +35,15 @@ const mainURL = 'file://' + path.join(__dirname, '../renderer', 'main.html');
 
 const singleInstanceLock = app.requestSingleInstanceLock();
 if (singleInstanceLock) {
-	app.on('second-instance', () => {
+	app.on('second-instance', (event, argv) => {
+		// uri scheme handler for windows
+		if (process.platform === 'win32') {
+			deepLinkingUrl = argv.slice(1);
+			if (mainWindow) {
+				mainWindow.webContents.send('deep-linking-url', deepLinkingUrl[isDev ? 1 : 0]);
```
akashnimare: Can you explain what's going on here?

kanishk98: Right, so we get the URL passed to the app in the `argv` object and then we need to send that to the renderer process. 
I noticed that the URL is present at the first element of the array when I build the app and it was present at the second element when in development mode. Not sure why, but I can look around if required.

vsvipul:
@kanishk98 i think we only need the URI scheme for the production environment and not for the dev environment. Also, i want to work on this. Ping me if some work is required on this.

kanishk98:
> i think we only need the URI scheme for the production environment and not for the dev environment. 

I agree, but it's not really that big of a hassle to include it in the dev environment. Just an extra condition that makes the experience better for Windows developers too. 

>  Also, i want to work on this. Ping me if some work is required on this.

Thanks a ton! I didn't see this earlier, so I ended up testing out `electron-builder` configs that focus and start the app when the browser receives a zulip:// URL. That seems to work properly right now, and I'm trying to figure how to catch the URL on Linux systems (again, macOS and Windows work well with just the Electron APIs). I'll text you if I get stuck with some Linux stuff.

akashnimare:
@kanishk98 is this ready to review or still WIP?

kanishk98:
This is WIP, I encountered a few bugs in the redirect logic that I haven't fixed yet.

vsvipul:
@kanishk98 i went through the PR. What seems to be the issue here? Maybe, i can help.

kanishk98:
> What seems to be the issue here? Maybe, i can help

That'd be great. Basically the app doesn't redirect to the intended URL if it's not running (the URL just opens up the app and then it stays there). I'm working on this right now, so I'll PM you for some help.

kanishk98:
To update everyone keeping an eye on this PR, there are some problems with `<webview>` that cause bugs with the redirection flow in the app. I've not been able to identify them, mostly because they're pretty erratic across operating systems and I think that it's best if I keep this on hold until work with the move to `BrowserView` is complete. That should address the problems with this PR.

akashnimare:
@kanishk98 are we planning to work on this after BrowserView PR gets merge?

kanishk98:
> are we planning to work on this after BrowserView PR gets merge?

Yeah, I've rechecked the logic for the redirection and the reloading of the `<webview>` (controlled by the `loadURL()` method, not us) is pretty erratic. I'd suggest we wait until that happens because we'll have to change things a bit anyway.

six8:
An alternative to this, or addition, would be a way to paste a URL into the Zulip App. I get emails with Zulip App URLs. If I open them in the browser, it loads the web app. The emails would have to use the `zulip://` scheme instead of `https://` for this PR to work with emails. But that would be confusing if the user isn't using the app.

A way to handle any URL may be to paste the URL into the Zulip App search bar and hit enter and have the Zulip App recognize and go to it.

kanishk98:
> But that would be confusing if the user isn't using the app

Hmm, I'm not sure if I agree with this statement. I'd say that OSes are fairly vocal when redirecting users to desktop apps from URLs - think Windows when it asks if you want to open a Zoom link in the app. Shouldn't that make things obvious to the user?

kanishk98:
@akashnimare in case we're planning to pick this up this week or so, I can fix the conflicts and work with @manavmehta on this.

frunkad:
Just following, if this has any chance of being picked up again? Though 5 years old, this would be quite an important feature for us - lmk if I can help?

PR #717, opened 2019-04-09
shortcut: Allow zoom options from numpad.

### LINKED ISSUE #344: zoom in/out doesn't work for both Ctr -/+ keys
Short keys used for zoom in/out are
Linux/Windows - **CTRL+** for zoom in, **CTRL-** for zoom out

zoom in/out should work for both `-`/`+` keys of keyboard(in case keyboard have two keys for `-`/`+`). 
They have different key code as they are different keys(ex. for `-` 109 and 189), so we have to set event for both key codes.

More discussion about this issue [here](https://chat.zulip.org/#narrow/stream/electron/subject/zoom.20in.2Fout).

vsvipul:
Numpad keys by default aren't supported by electron
for using in menu accelarators. Uses a workaround to
make zoom options work with Numpad keys.

Fixes #344.

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/preload.js
+++ b/app/renderer/js/preload.js
@@ -91,8 +91,15 @@ window.addEventListener('beforeunload', () => {
 
 // electron's globalShortcut can cause unexpected results
 // so adding the reload shortcut in the old-school way
+// Zoom from numpad keys is not supported by electron, so adding it through listeners.
 document.addEventListener('keydown', event => {
 	if (event.code === 'F5') {
 		ipcRenderer.send('forward-message', 'hard-reload');
+	} else if (event.ctrlKey && event.code === 'NumpadAdd') {
+		ipcRenderer.send('forward-message', 'zoomIn');
+	} else if (event.ctrlKey && event.code === 'NumpadSubtract') {
+		ipcRenderer.send('forward-message', 'zoomOut');
+	} else if (event.ctrlKey && event.code === 'Numpad0') {
+		ipcRenderer.send('forward-message', 'zoomActualSize');
 	}
 });


app/renderer/js/preload.js
```diff
@@ -91,8 +91,15 @@ window.addEventListener('beforeunload', () => {
 
 // electron's globalShortcut can cause unexpected results
 // so adding the reload shortcut in the old-school way
+// also, add zoom options from numpad keys
```
abhigyank:
```suggestion
// Zoom from numpad keys is not supported by electron, so adding it through listeners.
```

abhigyank:
Works as expected correctly. I am delighted that we have managed to solve an issue with tag `blocked-upstream `. :rocket:

app/renderer/js/preload.js
```diff
@@ -91,8 +91,15 @@ window.addEventListener('beforeunload', () => {
 
 // electron's globalShortcut can cause unexpected results
 // so adding the reload shortcut in the old-school way
+// Zoom from numpad keys is not supported by electron, so adding it through listeners.
 document.addEventListener('keydown', event => {
 	if (event.code === 'F5') {
```
akashnimare, line 96: Multiple if else, maybe use switch?

akashnimare (changes requested):

app/renderer/js/preload.js
vsvipul: @akashnimare I think adding a switch case will make the code more messy as we also have the Ctrl condition to check in zoom options, so we can go with one of these two, although i'd suggest we stay with if else only. 
1.
```
switch (event.code) {
        case 'F5': 
                 ipcRenderer.send('forward-message', 'hard-reload');
                 break;
	case 'NumpadAdd':  
                 if (event.ctrlKey) {
                     ipcRenderer.send('forward-message', 'zoomIn');
                 }
                 break;
        case 'NumpadSubtract':  
                 if (event.ctrlKey) {
                     ipcRenderer.send('forward-message', 'zoomOut');
                 }
                 break;
        case 'Numpad0':  
                 if (event.ctrlKey) {
                     ipcRenderer.send('forward-message', 'zoomActualSize');
                 }
                 break;
}
```

2.
```
if (event.code == 'F5') {
        ipcRenderer.send('forward-message', 'hard-reload');
} else if (event.ctrlKey) {
        switch (event.code) {
	case 'NumpadAdd':  
                     ipcRenderer.send('forward-message', 'zoomIn');
                     break;
        case 'NumpadSubtract':  
                     ipcRenderer.send('forward-message', 'zoomOut');
                     break;
        case 'Numpad0':
                     ipcRenderer.send('forward-message', 'zoomActualSize');
                     break;
        }
}
```
Which one should we go with ?

akashnimare: Okay, if else soundsss good then.

akashnimare:
Thanks, @vsvipul.

PR #721, opened 2019-04-11
nsis: Allow install without admin previliges.

### LINKED ISSUE #720: Windows: Installer without administrator privileges
The current installer requires administrative privileges to install Zulip. This is probably because it gets installed to `C:/Program Files`.
This can be an issue on corporate PCs where the user doesn't have admin privileges. A lot of applications (e.g. Chrome, Slack, VS Code) therefore allow to run the installer without admin privileges and install to the user directory instead (somewhere in `~/AppData`).

I know that the archive is available (the compression should be changed to .zip, as windows can open that natively without requiring extra software), but it doesn't contain the auto-update feature, which is quite important.

vsvipul:
Allow the nsis installer to install the app to
AppData for a single user so that administrator
previliges are not required while installing.

Fixes #720.

Refer to electron nsis config page [here](https://www.electron.build/configuration/nsis) for more info.
> If oneClick is false and perMachine is false (default): install mode installer page.

**Screenshots?**
Screenshot (203)
Screenshot (202)

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,9 @@
       "publisherName": "Kandra Labs, Inc."
     },
     "nsis": {
-      "allowToChangeInstallationDirectory": true
+      "allowToChangeInstallationDirectory": true,
+      "oneClick": false,
+      "perMachine": false
     }
   },
   "keywords": [


vsvipul:
Also, was getting the following error on running ```npm run dist``` in windows.
```
allowToChangeInstallationDirectory makes sense only for assisted installer (please set oneClick to false)
``` 
Screenshot (201)

Setting oneClick to false fixed this too.

akashnimare:
Merged, thanks. This was a blocker since it could break the auto-update. It doesn't matter now since we're pushing the release with latest electron-builder + updater anyway.

Issue #694, opened 2019-03-31
Individual screens for showing network error


kanishk98:
---

Currently, if there's a problem with the user's network connection, the app loads another tab with an error message. That leaves some pages blank and others working (if the connection is via proxy or some other corner case), and in general, may cause confusion. 

- **Operating System**:
  - [x] Windows
  - [x] Linux/Ubutnu
  - [ ] macOS

- **Suggested solution**:

I think we should show the error messages relevant to the servers themselves. So if a particular server cannot connect to a network, we show an error in the webview for that server alone. Discussion at [this thread](https://chat.zulip.org/#narrow/stream/16-desktop/topic/dismiss-network-error).

kanishk98:
@zulipbot claim

zulipbot:
Hello @kanishk98, you have been unassigned from this issue because you have not updated this issue or any referenced pull requests for over 14 days.

You can reclaim this issue or claim any other issue by commenting `@zulipbot claim` on that issue.

Thanks for your contributions, and hope to see you again soon!

kanishk98:
Working on it. :)

abhigyank:
My point of view on this issue is that, individual screen errors are good if the server doesn't loads not when it has already load and network issue is being handled by the webapp. For there are several features the webapp allows access to even when the webapp is not connected. We don't want to lose on those.
I'd prefer a discussion on czo regarding this before moving onto finalising the PR.

kanishk98:
> if the server doesn't loads not when it has already load and network issue is being handled by the webapp.

That functionality has not been affected. I've attached a screenshot below:

![image](https://user-images.githubusercontent.com/24617297/56137150-cd510580-5fb1-11e9-9f83-5f12a0be9695.png)

This is while running the code from the same PR. Once the server loads into the app, it's the same behaviour as before.

kanishk98:
Tackled and fixed in #723. Thanks for the help, everyone. :)

Issue #591, opened 2018-11-02
Online check ignores proxy settings


schrej:
Edit: The automatic reload on connection errors is caused by the [online check](https://github.com/zulip/zulip-electron/blob/master/app/renderer/js/utils/reconnect-util.js#L34) ignoring proxy settings.

---

It would be a nice addition if the connection error screen would try to reload zulip periodically.

A sane timeout would probably be something like once every minute.
It should do so in the background, so it's not really noticable and also doesn't flash the error screen every minute.
It should be configurable in the settings probably, maybe also the reconnect timeout.

Slack apparently does that as well. I don't have the issue with Slack at least (still in the transition phase, don't blame 😅)

Reason:
If I'm working remotely on my work notebook it connects to a VPN for internet access. This always takes some time and in the meantime zulip will have timed out. If I don't manually check the app, it will stay disconnected and I'll probably miss some messages.

akashnimare:
@schrej yeah, it's a known issue and we need to improve this network flow. Please see the related discussion [here](https://chat.zulip.org/#narrow/stream/16-desktop/topic/no.20internet).

schrej:
Yeah, just noticed that the error screen claims that it will retry, but it doesn't. Adapted the title a little ;)

kanishk98:
So I followed the discussion referenced by @akashnimare and messed around with the code a bit, and I think the following replacement for the `isOnline()` method should work:
```
isOnline() {
		request(testFavicon, (err, response) => {
			if (err) {
				return false;
			} else if (res) {
				return true;
			}
		});
		return false;
	}
```

I was unable to reproduce the issue on my end though, so I thought it best not to submit a PR until further discussion.

akashnimare:
@kanishk98 hmm, can you test your solution and see if it works?

superbiche:
@akashnimare @kanishk98 I've started playing around with this, using `is-online` only if we're not using a proxy, iterating the domains list to test for the favicon, and return true if one domain responds properly. Will let you know when I'm done

abhigyank:
^I don't think iterating over a the domain list is a good solution here. For one the number of domains could be many and two that does a lot of unnecessary queries.

superbiche:
[...]

PR #723, opened 2019-04-12
network: Tackle network issues independently.

kanishk98:
**What's this PR do?**

1. Splits network concerns of each server into its own tab and then implements retry logic with exponential backoff for each of them independently. 
2. Removes `is-online` package and instead uses DOM listeners to decide if connection has failed. 
3. Keeps retrying to reach a server's favicon by using `request-util.js`. By default, there is no limit.
4. Removes need to open a new `FunctionalTab` to show network error. Instead loads error-indicating `network.html` in the webview of the server that failed to connect. 

Please follow the discussions at #591 and #694.
In the former, it was decided that we should try to reach the `domain/static/favicon.ico`, but I'm unsure if that's the best idea (can someone move it from there?). Please let me know what to do there. 

**Screenshots?**

image

(this only shows the UI side of things, my computer's freezing up when I try to record a gif today). Please check that out and let me know if something goes wrong.

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -23,11 +23,6 @@ import CommonUtil = require('./utils/common-util');
 import EnterpriseUtil = require('./utils/enterprise-util');
 import Messages = require('./../../resources/messages');
 
-const logger = new Logger({
-	file: 'errors.log',
-	timestamp: true
-});
-
 interface FunctionalTabProps {
 	name: string;
 	materialIcon: string;
@@ -67,6 +62,11 @@ interface SettingsOptions {
 	dockBouncing?: boolean;
 	loading?: AnyObject;
 }
+
+const logger = new Logger({
+	file: 'errors.log',
+	timestamp: true
+});
 
 const rendererDirectory = path.resolve(__dirname, '..');
 type ServerOrFunctionalTab = ServerTab | FunctionalTab;
@@ -569,6 +569,8 @@ class ServerManagerView {
 	}
 
 	openNetworkTroubleshooting(index: number): void {
+		const reconnectUtil = new ReconnectUtil(this.tabs[index].webview);
+		reconnectUtil.pollInternetAndReload();
 		this.tabs[index].webview.props.url = `file://${rendererDirectory}/network.html`;
 		this.tabs[index].showNetworkError();
 	}
@@ -996,17 +998,7 @@ class ServerManagerView {
 
 window.addEventListener('load', () => {
 	const serverManagerView = new ServerManagerView();
-	const reconnectUtil = new ReconnectUtil(serverManagerView);
 	serverManagerView.init();
-	window.addEventListener('online', () => {
-		reconnectUtil.pollInternetAndReload();
-	});
-
-	window.addEventListener('offline', () => {
-		reconnectUtil.clearState();
-		logger.log('No internet connection, you are offline.');
-	});
-
 	// only start electron-connect (auto reload on change) when its ran
 	// from `npm run dev` or `gulp dev` and not from `npm start` when
 	// app is started `npm start` main process's proces.argv will have
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -370,7 +370,7 @@ class ServerManagerView {
 					}
 					this.showLoading(this.loading[this.tabs[this.activeTabIndex].webview.props.url]);
 				},
-				onNetworkError: this.openNetworkTroubleshooting.bind(this),
+				onNetworkError: (index: number) => this.openNetworkTroubleshooting(index),
 				onTitleChange: this.updateBadge.bind(this),
 				nodeIntegration: false,
 				preload: true
@@ -536,7 +536,7 @@ class ServerManagerView {
 					}
 					this.showLoading(this.loading[this.tabs[this.activeTabIndex].webview.props.url]);
 				},
-				onNetworkError: this.openNetworkTroubleshooting.bind(this),
+				onNetworkError: (index: number) => this.openNetworkTroubleshooting(index),
 				onTitleChange: this.updateBadge.bind(this),
 				nodeIntegration: true,
 				preload: false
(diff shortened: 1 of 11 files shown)


app/renderer/js/utils/reconnect-util.js
```diff
@@ -1,42 +1,63 @@
...
 });
 
 class ReconnectUtil {
-	constructor(serverManagerView) {
-		this.serverManagerView = serverManagerView;
-		this.alreadyReloaded = false;
+	constructor(webview) {
+		this.webview = webview;
+		this.url = webview.props.url;
+		this.exponentialBackoff = backoff.exponential();
```
schrej: The default backoff goes to a maximum of 10s., which is a little too frequent in my opinion.
I'd start with 5 seconds and the fibonacci strategy:
```suggestion
		this.exponentialBackoff = backoff.fibonacci({
                  initialDelay: 5000,
                  maxDelay: 300000
                });
```
that would generate 5s, 5s, 10s, 20s, 30s, 50s, 80s, 130s, 210s, 300s, 300s...

kanishk98: > The default backoff goes to a maximum of 10s., which is a little too frequent in my opinion.

I agree; I'll change that. 

> the fibonacci strategy:

Definitely not a problem, but I'm curious to know why you think that's better than exponential.

schrej: It grows a little bit slower.
5s, 10s, 20s, 40s, 80s, 160s, 300s would be the exponential one with factor 2.

kanishk98: Sure, makes sense. I'll push the changes right away.

schrej:
I've tested it in my proxy environment, it works as intended 👍

kanishk98:
> I've tested it in my proxy environment, it works as intended 👍

Awesome! @akashnimare please review the code whenever you have time and let me know if there's anything else to be done.

abhigyank:
I'd have preferred this to be two different PRs solving two different issues. @kanishk98  can you fix the commits accordingly so its easier to review.

kanishk98:
> I'd have preferred this to be two different PRs solving two different issues.

Sure, that would've been easy to review, but it would also have left this PR incomplete - the user can't connect to certain servers in their proxy environment, but we still open an error tab for them regardless. That's bad UX we should've sought to solve anyway with this PR, and #694's problems are actually just a subset of #591's problems. 
I can certainly drop a few commits and so on, but I don't think it's a good idea to separate the two. Let me know what you think.

kanishk98:
@abhigyank I have added a comment above on why I think this should not be split up into 2 separate PRs. Please let me know if you agree or not. If not, I'd like some guidance on what each child PR should include. Thanks! :)

kanishk98:
Updated PR to TypeScript.

kanishk98:
Fixed merge conflicts. :)

kanishk98:
@akashnimare if you'd like, I can cherrypick the relevant commits from #814 and combine the solutions from both PRs in this one.

app/renderer/js/main.ts
```diff
@@ -370,9 +370,9 @@ class ServerManagerView {
 					}
 					this.showLoading(this.loading[this.tabs[this.activeTabIndex].webview.props.url]);
 				},
-				onNetworkError: this.openNetworkTroubleshooting.bind(this),
+				onNetworkError: (index: number) => this.openNetworkTroubleshooting(index),
 				onTitleChange: this.updateBadge.bind(this),
-				nodeIntegration: false,
+				nodeIntegration: true,
```
akashnimare: @priyank-p, @kanishk98 is there any side effects of this?

priyank-p: I believe that is enables node integration for servers added, if so, I think we should avoid this because it gives access to node api to a url we are loading which isn't our code.

I think this is required for network.js and the workaround for that would be to require it in preload.js; that will allow us to not enable `nodeIntegration`

akashnimare: @kanishk98 ^^

kanishk98: Sure, I'll just make the change. Thanks for pointing this out. :)

akashnimare:
@kanishk98 thanks for making the changes. Since it's an XL PR we need to merge the browserview PR first so that we don't need to make changes there (that one is already pretty big).

kanishk98:
Sure, no problem. I had a look at your comments on the BrowserView PR and I'll be able to let you know how it works on a Mac only later tonight (I don't own one, so I have to wait for my friends).

akashnimare:
Great, thanks.

akashnimare:
@kanishk98 I'm getting the following error when the app reconnects to the internet -

image

akashnimare:
Also, can we make a list item here -


image

kanishk98:
> @kanishk98 I'm getting the following error when the app reconnects to the internet -

That's weird. I've been trying to reproduce this on my end and I can't. Can you text me on czo about what exactly you did to get this error?

> Also, can we make a list item here -

Sure, no problem. I'll do that right away.

akashnimare:
> That's weird. I've been trying to reproduce this on my end and I can't. Can you text me on czo about what exactly you did to get this error?

Follow this - 
* Start the app
* Turn off wifi
* Click on Try now button
* Turn on the wifi
* The app reconnects to the internet
* Check console

akashnimare:
@kanishk98 also getting this -
```
Message:
Cannot read property 'addEventListener' of null

Stacktrace:
TypeError: Cannot read property 'addEventListener' of null
    at NetworkTroubleshootingView.init (/Users/name/dev/zulipwork/zulip-desktop/app/renderer/js/pages/network.js:5:26)

    at window.addEventListener (/Users/name/dev/zulipwork/zulip-desktop/app/renderer/js/preload.js:103:18)

```

kanishk98:
@akashnimare I've added a check so that the error-raising code only runs when we show the network error page. Would you mind taking another look at this on your end to confirm it works without errors?

kanishk98:
I've also changed the error message to fit into `<ul />`. This changes the design a little, but I think it looks good. Let me know what you think. :)

akashnimare:
Merged. Thanks, @kanishk98.

PR #725, opened 2019-04-17
Add "role" key to webview property.

abhigyank:
`this.props.role` is used in `webview.js` but `role` key was missing in the webview properties.
Simple fix, GTM.

### DIFF
--- a/app/renderer/js/main.js
+++ b/app/renderer/js/main.js
@@ -186,6 +186,7 @@ class ServerManagerView {
 				index,
 				tabIndex,
 				url: server.url,
+				role: 'server',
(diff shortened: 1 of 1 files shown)


akashnimare:
We are already passing the role to the tabs. Do we really need this?

abhigyank:
Yeah, it's being used at one place in webview.js, which now shows undefined behavior, and webviews don't have access the properties of the tabs. If they can then it won't be required, but this is a simple solution than trying that in a very complicated way.

akashnimare:
Thanks, @abhigyank.

Issue #727, opened 2019-04-17
Improvements to desktop-app logger 


priyank-p:
There are couples of thing we will need to take care of for the `Logger` we use throughout the app. The logger is located in `app/renderer/js/utils/logger-util.js`, which does two thing it logs the the message in to browser or console and also writes the log in the file.
- First thing we will want to take care of is removing duplicate logs. For example a log one user reported contained:
```log
LOG | 2/22 10:23 |   There is no internet connection, try checking network cables, modem and router.
LOG | 2/22 10:24 |   There is no internet connection, try checking network cables, modem and router.
LOG | 2/22 10:26 |   There is no internet connection, try checking network cables, modem and router.
LOG | 2/22 10:27 |   There is no internet connection, try checking network cables, modem and router.
```
which we can replace with `(4) LOG | 2/22 |   There is no ...`. The key part of doing this would be looping over each log and ignore the timestamp, `2/22 10:23` part of the log, and de-duplicate it that way.

- Second thing would be removing the logs after it exceeds certain lines, we should do this by number of lines instead of file size to keep this feature simple. I think 300-500 lines would be ideal number of lines for a log file to retain. This feature is easy so we can implement this part first.

priyank-p:
@kanishk98 @vsvipul one of you could probably do the second part of limiting log file per certain lines easily. Just split the file read into a array split by newlines are remove the top lines if it exceeds the line size.

kanishk98:
Yeah, I think fixing this is pretty important. I'd love to take this up. 
@vsvipul let's divide the two points of work between ourselves?

vsvipul:
ok sure, i would like to take up the second part. Seems interesting. :)

vsvipul:
@kanishk98 As you are already busy with your open PRs and other important issue like URI Scheme and i18n , i would suggest that i take up this issue completely and work on it.

vsvipul:
@zulipbot claim

kanishk98:
> @kanishk98 As you are already busy with your open PRs and other important issue like URI Scheme and i18n , i would suggest that i take up this issue completely and work on it.

Sure, I guess that's fair - I've been a little swamped lately. Thanks a ton!
Let me know if you need any help.

zulipbot:
Hello @vsvipul, you have been unassigned from this issue because you have not updated this issue or any referenced pull requests for over 14 days.

You can reclaim this issue or claim any other issue by commenting `@zulipbot claim` on that issue.

[...]

PR #729, opened 2019-04-18
logs: Limit the number of lines in log files.

vsvipul:
Limits the numbers of lines in log files to 500.
Fixes a part of #727.

Probably needs more testing before merging.

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/utils/logger-util.js
+++ b/app/renderer/js/utils/logger-util.js
@@ -1,5 +1,6 @@
 const NodeConsole = require('console').Console;
 const fs = require('fs');
+const os = require('os');
 const isDev = require('electron-is-dev');
 const { initSetUp } = require('./default-util');
 const { sentryInit, captureException } = require('./sentry-util');
@@ -28,6 +29,13 @@ class Logger {
 		file = `${logDir}/${file}`;
 		if (timestamp === true) {
 			timestamp = this.getTimestamp;
+		}
+
+		// Trim log according to type of process
+		if (process.type === 'renderer') {
+			requestIdleCallback(() => this.trimLog(file));
+		} else {
+			process.nextTick(() => this.trimLog(file));
 		}
 
 		const fileStream = fs.createWriteStream(file, { flags: 'a' });
@@ -87,6 +95,24 @@ class Logger {
 
 	reportSentry(err) {
 		captureException(err);
+	}
+
+	trimLog(file) {
+		fs.readFile(file, 'utf8', (err, data) => {
+			if (err) {
+				throw err;
+			}
+			const MAX_LOG_FILE_LINES = 500;
+			const logs = data.split(os.EOL);
+			const logLength = logs.length - 1;
+
+			// Keep bottom MAX_LOG_FILE_LINES of each log instance
+			if (logLength > MAX_LOG_FILE_LINES) {
+				const trimmedLogs = logs.slice(logLength - MAX_LOG_FILE_LINES);
+				const toWrite = trimmedLogs.join(os.EOL);
+				fs.writeFileSync(file, toWrite);
+			}
+		});
 	}
 }
 


app/renderer/js/utils/logger-util.js
```diff
@@ -30,6 +30,20 @@ class Logger {
 			timestamp = this.getTimestamp;
 		}
 
+		fs.readFile(file, 'utf8', (err, data) => {
+			if (err) {
+				throw err;
+			}
+			const logs = data.split('\n');
+			// Keep top 300 files of log for each instance
+			const trimLines = logs.slice(Math.max(logs.length - 300, 0));
```
abhigyank: I think we would want to keep the bottom 300 lines i.e. the latest logs.

vsvipul: I think this does just that.. slice removes the top (total-300) lines and leaves the bottom 300.. does it not?

app/renderer/js/utils/logger-util.js
```diff
@@ -30,6 +30,20 @@ class Logger {
...
+			let toWrite = '';
+			trimLines.forEach(line => {
```
abhigyank: I am not sure, but is there a better algorithmic way to this? I think this is of high complexity and could slow the initial startup.

vsvipul: Have to think about this.. what do u think @priyank-p ?

app/renderer/js/utils/logger-util.js
abhigyank: Okay, it does that. I just went by the comments.

vsvipul: Oh sorry, wrote wrongly in the comment.. will update that.

priyank-p: Let's create a `MAX_LOG_FILE_LINES` constant at the top use that here. Feel like that would be better.

priyank-p: Also, let change the `trimLines` variable to `trimmedLogs`.

app/renderer/js/utils/logger-util.js
```diff
@@ -30,6 +30,20 @@ class Logger {
 			timestamp = this.getTimestamp;
 		}
 
+		fs.readFile(file, 'utf8', (err, data) => {
+			if (err) {
+				throw err;
+			}
+			const logs = data.split('\n');
```
priyank-p: Instead of using `\n` here, use `os.EOL` here and elsewere to handle windows line-endings.

app/renderer/js/utils/logger-util.js
```diff
@@ -30,6 +30,20 @@ class Logger {
...
+			// Keep top 300 files of log for each instance
+			const trimLines = logs.slice(Math.max(logs.length - 300, 0));
+			let toWrite = '';
+			trimLines.forEach(line => {
+				toWrite += line + '\n';
+			});
```
priyank-p: You don't need to use a forEach loop here just use `trimmedLogs.join(os.EOL)` here to get the log file in string format.

app/renderer/js/utils/logger-util.js
priyank-p: For the slow down issue, I think it would be one time impact since once we run this code the first time the log file shouldn't be huge. But let's extract this feature into its own function and call it after the `load` event in `main.js`. That would take care of this. Or another thing to try out would be `requestIdleCallback()`.

priyank-p:
@vsvipul this looks good so far, Thanks for picking this up. @akashnimare @abhigyank I think we should increase the log file lines to 500 instead of the current 300 limit. It won't have that much impact on file size. What do you all think?

akashnimare:
> It won't have that much impact on file size. What do you all think?

Yeah, 500 lines are cool too. @priyank-p I was wondering how much this feature will cost us (the reading/removing operation of logs we are doing on startup).

app/renderer/js/utils/logger-util.js
abhigyank: Yeah, I think that should reduce the computational complexity.

app/renderer/js/utils/logger-util.js
abhigyank: If we keep the function to be an asynchronous one, the performance should also be better.
But the only issue could be with async (probably even with its own sync function) being two instances trying to write to the file at the same time.

priyank-p:
I ran some benchmarks on my old laptop (It's not slow but compared to my new one it is) and the result is pretty good. The most extreme case I ran was 1000k lines of logs and trimmed 250K lines. The log file was of approximately 89.82MB (89818 KB exactly) and the trimmed version came out to be 22.45 MB (22452 KB exactly). That is higher than what the user reported on CZO. The time it took was 1.17s to do all the cleanup. (0.77s to read the log file + 0.01s to trim excessive lines + 0.3s rewrite the new log file).

I should also note that we read the log file asynchronously while for the benchmark I used sync so it should be slightly better performance-wise for the app.  I would be great for others to run these benchmark on different os and especially if they have an old computer. I ran mine on windows.

Test Script: https://gist.github.com/priyank-p/026e0f4262788f0b740bdf60574d6890

<details>
<summary>Benchmark Details</summary>

```log
Log file lines: 5000, and MAX_LOG_LINES: 500
$ node test-fs.js 5000 500
Reading log file: 0.921ms
Trimming excessive logs: 0.116ms
Rewriting trimmed log file: 4.136ms

Log file lines: 5000, and MAX_LOG_LINES: 500
$ node test-fs.js 5000 100
Reading log file: 0.901ms
Trimming excessive logs: 0.117ms
Rewriting trimmed log file: 3.670ms

Log file lines: 5000, and MAX_LOG_LINES: 500
$ node test-fs.js 5000 500
Reading log file: 0.839ms
Trimming excessive logs: 0.119ms
Rewriting trimmed log file: 3.881ms

Log file lines: 5000, and MAX_LOG_LINES: 500
$ node test-fs.js 5000 1000
Reading log file: 0.845ms
Trimming excessive logs: 0.115ms
Rewriting trimmed log file: 3.978ms

Log file lines: 5000, and MAX_LOG_LINES: 500
$ node test-fs.js 5000 2500
Reading log file: 0.931ms
Trimming excessive logs: 0.118ms
Rewriting trimmed log file: 4.828ms

Log file lines: 5000, and MAX_LOG_LINES: 500
$ node test-fs.js 10000 2500
Reading log file: 3.209ms
Trimming excessive logs: 0.126ms
Rewriting trimmed log file: 5.732ms

$ node test-fs.js 1000000 250000
Log Lines 1000001
Reading log file: 774.649ms
Trimming excessive logs: 11.182ms
Rewriting trimmed log file: 306.274ms
```

</details>

app/renderer/js/utils/logger-util.js
vsvipul: @abhigyank that is the exact reason why i used the synchronous one here..

vsvipul: @priyank-p if i call it after the load event in ```main.js``` , how do i exactly make it run for all log files. Do i loop over all log files in log directory? Will that be feasible?

priyank-p: Ah didn't think that through. First, extract this feature into its own function then in the constructor check if it is the renderer process. If it renderer process use `requestIdleCallback()` otherwise use `process.nextTick` for main process.

(Can you check if `requestIdleCallback` is available in the main process of Electron? If so we should use that.)

vsvipul: ```requestIdleCallback``` doesn't seem to be working for the main process. Will use process.nextTick.

vsvipul: @priyank-p Using ```request.IdleCallback``` for the renderer process made the app hang up with no domains loaded. While using only process.nextTick works perfectly. I am attaching screenshots for reference. I have pushed the code only for using process.nextTick for both renderer and main process. I suppose it can be used for both. I don't know why requestIdleCallback doesn't work here.
Screenshot from 2019-04-20 04-21-37
Screenshot from 2019-04-20 04-21-15

priyank-p: OK, I see the issue. In the screenshot you attached the code shows `requestIdleCallback(this.trimLog(file))` which is incorrect it should be `requestIdleCallback(() => this.trimLog(file))`. I bet there is error in the console about that you missed. The problem is that you are calling the function in the paramter which return null which is passed in to `requestIdleCallback` which in turn throws error hence the failure of app to load.

vsvipul: oh, let me test that.

vsvipul: @priyank-p works fine now. Thanks for your help. Tested for both main and renderer log files. Pushed the changes.

vsvipul:
@akashnimare @priyank-p please test and review again. Made the required changes.

app/renderer/js/utils/logger-util.js
```diff
@@ -88,6 +95,20 @@ class Logger {
...
+
+	trimLog(file) {
+		fs.readFile(file, 'utf8', (err, data) => {
+			if (err) {
+				throw err;
+			}
+			// Keep bottom MAX_LOG_FILE_LINES of each log instance
+			const MAX_LOG_FILE_LINES = 500;
+			const logs = data.split(os.EOL);
+			const trimmedLogs = logs.slice(Math.max(logs.length - MAX_LOG_FILE_LINES, 0));
```
abhigyank: For some reason `logs.length` is returning one value more than the number of lines (not sure why) , so maybe add a -1 to `logs.length - MAX_LOG_FILE_LINES - 1` along with a comment.
Unless you can find out why thats occurring.

app/renderer/js/utils/logger-util.js
```diff
@@ -88,6 +95,20 @@ class Logger {
...
 	}
+
+	trimLog(file) {
+		fs.readFile(file, 'utf8', (err, data) => {
+			if (err) {
+				throw err;
+			}
+			// Keep bottom MAX_LOG_FILE_LINES of each log instance
+			const MAX_LOG_FILE_LINES = 500;
+			const logs = data.split(os.EOL);
```
abhigyank, line 106: We could probably do a check like `if (logs.length <500)` then don't do the trimming part. Would reduce a bit unnecessary computation and file write.

abhigyank:
Using Priyank's benchmark script I ran a couple of tests on my Ubuntu laptop (old but not very slow) and results were fine. Overall for the script wouldn't make much difference on a day-to-day basis (maybe less than 0.1ms add to the startup). It would probably affect one time "asynchronously" when the app would be updated and the script would run for the first time, that too when the files are of length more than 100K. Reading this file takes time, but since that would be asynchronously happening shouldn't effect much.

app/renderer/js/utils/logger-util.js
vsvipul: Yeah, good thought ! will do.

app/renderer/js/utils/logger-util.js
vsvipul: I think that is because split by EOL is taking a last empty line as the last element in the array.

vsvipul: Made the update.

vsvipul:
@abhigyank i have made the changes requested. Can you please review once more ?

abhigyank:
LGTM. @priyank-p  can you do a final review?

vsvipul:
@akashnimare as @priyank-p is away, can you do a review?

akashnimare:
Works as expected. Merging this for now.

PR #437, opened 2018-03-04
il8n: Add a script that automatically creates locales from template.

priyank-p:
This script automatically builds locales if the template is changed. it sorts the outputted JSON.

PR #730, opened 2019-04-19
Add translation support to menu

kanishk98:
**What's this PR do?**

Gets the `default_language` of the user's realm and updates the app menu accordingly. 
Also updates the `google-translate-api` module used to generate the `.json` files in `app/translations/` to prevent `BAD_REQUEST` errors.

Discussion at #407. 
Please also see #437 and #436. 

**Screenshots?**

link

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

**Further work**

I'll also be pushing changes to the settings page once we decide that this approach is fine.


akashnimare:
Let's have a discussion about this in https://chat.zulip.org/#narrow/stream/16-desktop/topic/spellchecker.

abhigyank:
@kanishk98 can you rebase this?

kanishk98:
@abhigyank sorry, missed this last night. Rebased the PR.

kanishk98:
After some users reporting bugs with their JSON files recently, I decided to take another look at this PR to try and reduce the dependence of this feature on the disk. The last commit broke functionality, so marking this as WIP for now. Will just move back to the disk-based approach if nothing works out. :)

kanishk98:
Alright, so the JSON db errors mostly stemmed from switching tabs and now that we've addressed that, the PR works as expected for me. I think this is ready for review again. Removing WIP flag.

akashnimare:
Can you fix the merge conflicts?

app/renderer/js/main.js
```diff
@@ -621,6 +630,20 @@ class ServerManagerView {
 			});
 		}
 
+		ipcRenderer.on('language-change', (event, { serverLanguage, domain }) => {
+			const index = this.getTabIndex(domain);
+			const oldLanguage = this.tabs[index].props.language;
+			this.tabs[index].props.language = serverLanguage;
+			if (oldLanguage !== serverLanguage) {
```
akashnimare: This is to handle the case where both org have the same language, right?

kanishk98: Uh, yeah, pretty much. More accurately, this is for the case when the app default (English) is not the same as the server language.

akashnimare:
LGTM. @priyank-p quick review?

app/renderer/js/utils/translation-util.js
```diff
@@ -0,0 +1,31 @@
...
+const path = require('path');
+const i18n = require('i18n');
+
+let instance = null;
+
+class TranslationUtil {
+	constructor() {
+		if (instance) {
+			return this;
+		} else {
```
priyank-p: Minor nit: you can just remove this else clause here.

app/renderer/js/utils/translation-util.js
```diff
@@ -0,0 +1,31 @@
...
+			instance = this;
+			i18n.configure({
+				directory: path.join(__dirname, '../../../translations/'),
+				register: this
+			});
+		}
+		return this;
```
priyank-p: Remove this! This is basically useless.

app/main/menu.js
```diff
@@ -441,44 +464,58 @@ class AppMenu {
...
+				label: TranslationUtil.__('Paste and Match Style'),
 				role: 'pasteandmatchstyle'
 			}, {
 				type: 'separator'
 			}, {
+				label: TranslationUtil.__('Select All'),
 				role: 'selectall'
 			}]
 		}, {
-			label: '&View',
```
priyank-p: Why did we have `&` here for?

priyank-p (commented):
LGTM! Left some minor comments

app/main/menu.js
kanishk98: Not sure. Maybe @akashnimare can help out here.

kanishk98:
Pushed the suggested changes.

app/main/menu.js
akashnimare: This came from a standard electron template. Not sure if we need this now.

app/renderer/js/pages/preference/add-certificate.js
```diff
@@ -18,12 +19,12 @@ class AddCertificate extends BaseComponent {
 		return `
 			<div class="settings-card certificates-card">
 				<div class="certificate-input">
-					<div>Organization URL</div>
+					<div>${TranslationUtil.__('Organization URL')}</div>
```
akashnimare: @kanishk98 instead of `TranslationUtil` you can use `t` maybe?

priyank-p:
@kanishk98 I rebased this PR to solve merge conflicts related to typescript can you cross check?

kanishk98:
> I rebased this PR to solve merge conflicts related to typescript can you cross check?

I did, everything looks great. Thanks a ton, @priyank-p!

> @kanishk98 instead of TranslationUtil you can use t maybe?

@akashnimare made the changes and force pushed. 
I think this PR is ready for merging now. Let me know if any other changes are needed.

akashnimare:
@kanishk98 can you update this whenever you get time?

kanishk98:
Fixed merge conflicts and updated PR.

app/renderer/js/utils/translation-util.ts
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+import path = require('path');
+import electron = require('electron');
+import i18n = require('i18n');
+
+let instance: TranslationUtil = null;
```
akashnimare, line 7: @priyank-p, @vsvipul is this correct? We have used following at other places -
```
let instance: null | ConfigUtil = null;
```

app/renderer/js/utils/translation-util.ts
```diff
@@ -0,0 +1,35 @@
...
+
+		instance = this;
+		i18n.configure({
+			directory: path.join(__dirname, '../../../translations/'),
+			register: this
+		});
+	}
+
+	__(phrase: string): string {
+		return i18n.__({ phrase, locale: app.getLocale() });
```
akashnimare: @kanishk98 should we also set `ENGLISH` here as a fallback?

akashnimare:
@vsvipul can you test this on Linux? This works fine for me on macOS.

app/renderer/js/utils/translation-util.ts
priyank-p: Yes, it is! We need to do a sweep in couple of places.

akashnimare:
@kanishk98 merge the conflicts. I think it's GTG and we could include this in the next stable release.

kanishk98:
@akashnimare fixed the conflicts. I also tested it on Linux, so the PR looks good.

app/renderer/js/utils/translation-util.ts
kanishk98: I doubt that it's required because `app.getLocale()` should always work, but it's never bad to be safe, so I added a commit for that anyway. We can drop that if others feel it is unnecessary.

kanishk98: I'm assuming you're talking about the corner case where `app.getLocale()` returns a falsey value.

akashnimare: Correct.

akashnimare:
@kanishk98 thanks for working on this.

PR #732, opened 2019-04-22
settings: Add option to toggle Error Reporting.

### LINKED ISSUE #702: Disable Zulip telemetry?  Client connects to domains other than self-hosted server.
All,

I am using the Zulip desktop client on macOS.  I noticed the client tries to connect to the following domains (even without any content pasted in a channel that's related).  Can the code be scrubbed to make sure the client doesn't send data to any place other than the Zulip server it's connected to?

Domains Zulip client has reached out to:

github.com
google.com
sentry.io
tabbott.net
gravatar.com

vsvipul:
**What's this PR do?**
Adds an option to enable or disable sentry error
reporting under Advanced section in General
Settings. Handles both main and renderer
processes.

Also, changes the domain used to resolve proxy
in proxy-util from google.com to example.com.

Fixes #702.

**Screenshots?**
Screenshot from 2019-04-23 02-13-49

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/index.js
+++ b/app/main/index.js
@@ -159,7 +159,10 @@ app.on('ready', () => {
 	}
 
 	// Initialize sentry for main process
-	sentryInit();
+	const errorReporting = ConfigUtil.getConfigItem('errorReporting');
+	if (errorReporting) {
+		sentryInit();
+	}
 
 	const isSystemProxy = ConfigUtil.getConfigItem('useSystemProxy');
 
@@ -343,6 +346,12 @@ app.on('ready', () => {
 
 	ipcMain.on('realm-icon-changed', (event, serverURL, iconURL) => {
 		page.send('update-realm-icon', serverURL, iconURL);
+	});
+
+	// Using event.sender.send instead of page.send here to
+	// make sure the value of errorReporting is sent only once on load.
+	ipcMain.on('error-reporting', event => {
+		event.sender.send('error-reporting-val', errorReporting);
 	});
 });
 
--- a/app/renderer/js/utils/logger-util.js
+++ b/app/renderer/js/utils/logger-util.js
@@ -5,10 +5,23 @@ const { initSetUp } = require('./default-util');
 const { sentryInit, captureException } = require('./sentry-util');
 
 initSetUp();
-sentryInit();
+
 let app = null;
+let reportErrors = true;
 if (process.type === 'renderer') {
 	app = require('electron').remote.app;
+
+	// Report Errors to Sentry only if it is enabled in settings
+	// Gets the value of reportErrors from config-util for renderer process
+	// For main process, sentryInit() is handled in index.js
+	const { ipcRenderer } = require('electron');
+	ipcRenderer.send('error-reporting');
+	ipcRenderer.on('error-reporting-val', (event, errorReporting) => {
+		reportErrors = errorReporting;
+		if (reportErrors) {
+			sentryInit();
+		}
+	});
 } else {
 	app = require('electron').app;
 }
@@ -86,7 +99,9 @@ class Logger {
 	}
 
 	reportSentry(err) {
-		captureException(err);
+		if (reportErrors) {
+			captureException(err);
+		}
 	}
 }
 
--- a/app/renderer/js/main.js
+++ b/app/renderer/js/main.js
@@ -111,6 +111,7 @@ class ServerManagerView {
 			showNotification: true,
 			autoUpdate: true,
 			betaUpdate: false,
+			errorReporting: true,
 			customCSS: false,
 			silent: false,
 			lastActiveTab: 0,
--- a/app/renderer/js/pages/preference/general-section.js
+++ b/app/renderer/js/pages/preference/general-section.js
@@ -83,6 +83,10 @@ class GeneralSection extends BaseSection {
 				</div>
 				<div class="title">Advanced</div>
 				<div class="settings-card">
+				<div class="setting-row" id="enable-error-reporting">
+					<div class="setting-description">Enable error reporting (requires restart)</div>
+					<div class="setting-control"></div>
+				</div>
 				<div class="setting-row" id="show-download-folder">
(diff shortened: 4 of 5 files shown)


akashnimare:
@abhigyank, @rishig should we show some kind of warning dialog when a user disable this feature? like "This feature helps us improve the app, do you want to disable this?" since error reporting is essential for making the app better.

rishig:
Good question. I don't think it's too important to put a warning dialog; most users who turn it off will know what they're doing, I think. 

Other notes: I would put it in Advanced, rather than Functionality. We also use sentence case for these settings.

akashnimare:
Make sense. @vsvipul can you make the changes?

vsvipul:
@akashnimare sure, on it !

vsvipul:
@akashnimare made the requested changes. Also, handled sentryInit() in both main and renderer process. Review please.

app/renderer/js/utils/logger-util.js
```diff
@@ -5,14 +5,29 @@ const { initSetUp } = require('./default-util');
 const { sentryInit, captureException } = require('./sentry-util');
 
 initSetUp();
-sentryInit();
+
```
akashnimare, line 8: @vsvipul did you check up with @abhigyank for this?

vsvipul: Yes i did, this was done because sentryInit() needs to be done separately for both processes. I've handled that here.

vsvipul:
@akashnimare any changes required for this ?

vsvipul:
@akashnimare a reminder for this. Thanks.

app/renderer/js/utils/logger-util.js
```diff
@@ -5,14 +5,29 @@ const { initSetUp } = require('./default-util');
...
 	app = require('electron').remote.app;
 } else {
 	app = require('electron').app;
 }
 
+// Report Errors to Sentry only if it is enabled in settings
+// Gets the value of reportErrors from config-util for renderer process
+// For main process, sentryInit() is handled in index.js
+let reportErrors = true;
+if (process.type === 'renderer') {
```
akashnimare: We are already using this condition above. So maybe wrap it up there only.

akashnimare (changes requested):

app/renderer/js/utils/logger-util.js
vsvipul: Ok sure, will do.

vsvipul: @akashnimare Made the change. Thanks.

akashnimare (approved):

akashnimare:
LGTM. @abhigyank?

abhigyank:
Currently this would require a restart of the app I guess , as we aren't disabling sentry at runtime. A hackish wat to do this - https://github.com/getsentry/sentry-javascript/issues/2039

Either we could do this on toggling sentry off, or add in the options `(requires restart)`.

akashnimare:
> Either we could do this on toggling sentry off, or add in the options (requires restart).

For now, we could go with `requires restart` option.

vsvipul:
Changed the option text to add ```(requires restart)``` .

app/main/index.js
```diff
@@ -159,7 +159,9 @@ app.on('ready', () => {
 	}
 
 	// Initialize sentry for main process
```
akashnimare, line 161: We could put this here and re-use at other places-
```
const errorReporting = ConfigUtil.getConfigItem('errorReporting');
```

app/main/index.js
```diff
@@ -344,6 +346,11 @@ app.on('ready', () => {
 	ipcMain.on('realm-icon-changed', (event, serverURL, iconURL) => {
 		page.send('update-realm-icon', serverURL, iconURL);
 	});
+
+	ipcMain.on('error-reporting', event => {
+		const errorReporting = ConfigUtil.getConfigItem('errorReporting');
+		event.sender.send('error-reporting-val', errorReporting);
```
akashnimare, line 354: Is there any reason for not using `page.send` here?

akashnimare (changes requested):

app/main/index.js
vsvipul: good thought. Will change.

app/main/index.js
vsvipul: Using page.send, the errorReporting value is given to the ```logger-util``` multiple times. When event.sender.send is used, it is given only once, which should be the case as the logger is loaded only once on application start.

akashnimare: Okay, maybe add this in the comment?

vsvipul: Done.

akashnimare (approved):

app/renderer/js/utils/logger-util.js
```diff
@@ -5,10 +5,23 @@ const { initSetUp } = require('./default-util');
...
 let app = null;
+let reportErrors = true;
 if (process.type === 'renderer') {
 	app = require('electron').remote.app;
+
+	// Report Errors to Sentry only if it is enabled in settings
+	// Gets the value of reportErrors from config-util for renderer process
+	// For main process, sentryInit() is handled in index.js
+	const { ipcRenderer } = require('electron');
+	ipcRenderer.send('error-reporting');
```
akashnimare, line 18: I forgot why we are using this method instead of getting the value from configUtil?

vsvipul: This was done because directly using ConfigUtil here causes a circular dependency (or say a race condition) between the ConfigUtil and the LoggerUtil.

vsvipul:
@akashnimare are we holding this because of issues with autoupdate or #737 ?

abhigyank (approved):

akashnimare:
Merged. Thanks, @vsvipul and @abhigyank 🚀

PR #734, opened 2019-05-02
focus: Fix focus after clicking back button.

vsvipul:
Addresses the issue of the webview not being in focus
after the Back button is clicked. Now, the webview is focused
explicitly by calling focus() on click.

More discussion about this [here](https://chat.zulip.org/#narrow/stream/16-desktop/topic/Back.20arrow.20.2F.20Down.20arrow)

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS


akashnimare:
Thanks, @vsvipul.

PR #748, opened 2019-05-21
system-presence: Add ability to detect system presence.

vsvipul:
**What's this PR do?**
Created two electron-bridge variables.
idle_on_system tells whether the user is idle or not.
last_active_on_system gives the time the user was last
active. Both variables will be used by the webapp to set
user status.

Fixes #352.

Corresponding webapp PR [here](https://github.com/zulip/zulip/pull/12003).

**How to test?**
You can test by console logging the two variables in the setInterval near line 97 of preload.js. They will update based on user activity.

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/index.ts
+++ b/app/main/index.ts
@@ -362,6 +362,22 @@ app.on('ready', () => {
 	ipcMain.on('save-last-tab', (_event: Electron.IpcMessageEvent, index: number) => {
 		ConfigUtil.setConfigItem('lastActiveTab', index);
 	});
+
+	// Update user idle status for each realm after every 15s
+	// Set user idle if no activity for idleThreshold time in seconds
+	// Do a check every idleCheckInterval ms.
+	const idleCheckInterval = 15000;
+	setInterval(() => {
+		const idleThreshold = 60;
+		// Remove typecast to any when querySystemIdleState gets added to electron typings
+		(electron.powerMonitor as any).querySystemIdleState(idleThreshold, (idleState: string) => {
+			if (idleState === 'active') {
+				page.send('set-active');
+			} else {
+				page.send('set-idle');
+			}
+		});
+	}, idleCheckInterval);
 });
 
 app.on('before-quit', () => {
--- a/app/main/index.ts
+++ b/app/main/index.ts
@@ -364,13 +364,17 @@ app.on('ready', () => {
 	});
 
 	// Update user idle status for each realm after every 15s
-	// Set user idle if no activity for idleThreshold time in seconds
-	// Do a check every idleCheckInterval ms.
-	const idleCheckInterval = 15000;
+	const idleCheckInterval = 15 * 1000; // 15 seconds
 	setInterval(() => {
-		const idleThreshold = 60;
-		// Remove typecast to any when querySystemIdleState gets added to electron typings
-		(electron.powerMonitor as any).querySystemIdleState(idleThreshold, (idleState: string) => {
+		// Set user idle if no activity in 1 second (idleThresholdSeconds)
+		const idleThresholdSeconds = 1; // 1 second
+
+		// TODO: Remove typecast to any when types get added
+		// TODO: use powerMonitor.getSystemIdleState when upgrading electron
+		// powerMonitor.querySystemIdleState is deprecated in current electron
+		// version at the time of writing.
+		const powerMonitor = electron.powerMonitor as any;
+		powerMonitor.querySystemIdleState(idleThresholdSeconds, (idleState: string) => {
 			if (idleState === 'active') {
 				page.send('set-active');
 			} else {
--- a/app/renderer/js/electron-bridge.ts
+++ b/app/renderer/js/electron-bridge.ts
@@ -9,7 +9,20 @@ type ListenerType = ((...args: any[]) => void);
 // for the whole file.
 /* eslint-disable @typescript-eslint/camelcase */
 class ElectronBridge extends events {
-	send_notification_reply_message_supported = false;
+	send_notification_reply_message_supported: boolean;
+	idle_on_system: boolean;
+	last_active_on_system: number;
+
+	constructor() {
+		super();
+		this.send_notification_reply_message_supported = false;
+		// Indicates if the user is idle or not
+		this.idle_on_system = false;
+
+		// Indicates the time at which user was last active
+		this.last_active_on_system = Date.now();
+	}
+
 	send_event(eventName: string | symbol, ...args: any[]): void {
 		this.emit(eventName, ...args);
 	}
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -961,6 +961,20 @@ class ServerManagerView {
 		ipcRenderer.on('new-server', () => {
 			this.openSettings('AddServer');
 		});
+
+		ipcRenderer.on('set-active', () => {
+			const webviews: NodeListOf<Electron.WebviewTag> = document.querySelectorAll('webview');
+			webviews.forEach(webview => {
+				webview.send('set-active');
+			});
+		});
+
+		ipcRenderer.on('set-idle', () => {
+			const webviews: NodeListOf<Electron.WebviewTag> = document.querySelectorAll('webview');
+			webviews.forEach(webview => {
+				webview.send('set-idle');
+			});
+		});
 	}
 }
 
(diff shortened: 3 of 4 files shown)


app/renderer/js/preload.js
```diff
@@ -13,9 +13,19 @@ require('./notification');
...
+// so turning the rule off for the whole file.
+/* eslint-disable camelcase */
+
 window.electron_bridge = require('./electron-bridge');
 
+// Indicates if the user is idle or not
+window.electron_bridge.idle_on_system = false;
+
+ // Indicates the time at which user was last active
+window.electron_bridge.last_active_on_system = Date.now();
```
priyank-p: Set this instance variables, in the electron_bridge class constructor. You will need to call `super()` in the first line.

vsvipul: @priyank-p do you want me to construct a subclass of electron_bridge class and use it here in preload.js ?

priyank-p: No, what I am trying to say is initialize the variables in constructor of `electron_bridge` class.

vsvipul: ok. got it.

vsvipul:
@priyank-p @abhigyank i have made some changes. Can you please test and review with corresponding webapp [PR](https://github.com/zulip/zulip/pull/12003). Also, i'm unsure about the webapp side logic. Any help there?

app/renderer/js/preload.js
```diff
@@ -78,6 +82,10 @@ document.addEventListener('DOMContentLoaded', () => {
 			}
 		});
 	}
+	// Update user status after every 15s
+	setInterval(() => {
+		ipcRenderer.send('detect-presence');
+	}, 15000);
```
timabbott: Just to check, I assume this is code that runs only once if you have N organizations?

app/main/index.js
```diff
@@ -358,6 +358,19 @@ app.on('ready', () => {
 	ipcMain.on('save-last-tab', (event, index) => {
 		ConfigUtil.setConfigItem('lastActiveTab', index);
 	});
+
+	// Update user idle status for each realm
+	// Set user idle if no activity for 60s
+	ipcMain.on('detect-presence', event => {
+		const idleThreshold = 60;
```
timabbott: Just to confirm, that number is definitely in seconds, not milliseconds or some other unit?  

(It might make sense to just have the comment mention the units rather than repeating the number, in case we change the number.).

app/renderer/js/preload.js
timabbott: I would have thought we'd just have a single copy of the setInterval() calling `electron.powerMonitor.querySystemIdleState`, and having that send IPC data to each Zulip app window, but maybe I'm missing something here.

app/main/index.js
vsvipul: Yes, it is in seconds! Will add a comment for it!

app/renderer/js/preload.js
vsvipul: Each org has a different preload.js in their webview. So, this is called differently for each org. I couldn't figure out a way to do this only once and then send it to each window. Will look into it more.

abhigyank: @vsvipul  to address this issue - let's move the the `setInterval` to main process so that it occurs once for N orgs and send a IPC data to the `preload.js` directly with `set-active` or `set-idle`. Can you try this out?

vsvipul: @abhigyank I'll try it out. Thanks. This is what i was doing initially, but someone pointed out that the SetInterval should be placed in preload.js and not in the main.

vsvipul: @abhigyank @timabbott  I have moved the logic to main.js and now it runs only once and sets the variable for all webviews. I hope this resolves the issue.

app/main/index.js
```diff
@@ -358,6 +358,19 @@ app.on('ready', () => {
 	ipcMain.on('save-last-tab', (event, index) => {
 		ConfigUtil.setConfigItem('lastActiveTab', index);
 	});
+
+	// Update user idle status for each realm after every 15s
+	// Set user idle if no activity for idleThreshold time in seconds
+	setInterval(() => {
+		const idleThreshold = 60;
```
timabbott: The 60s and 15s seem to not agree?

timabbott: Oh, never mind; might want to make the 15000 as a `const` variable too for readability.

timabbott:
I don't know the electron APIs, so take this with a grain of salt, but this setup looks right to me.

app/main/index.js
vsvipul: Will change.

app/renderer/js/preload.js
abhigyank: The updated change on this looks good to me.

vsvipul:
@timabbott @abhigyank  Is this and the webapp PR good to go then ?

timabbott:
I merged the webapp PR; I think we can integrate this, and then just do some pretty thoughtful evaluation on the desktop app side to make sure it's doing the right thing.  

I wonder if there'd be value in having a developer flag to show the current user's presence status on the window somewhere, which we can just have on while working on the desktop app over the next few weeks to validate whether there are bugs here.  Not sure what the right implementation for that should be, but I'm sure we can come up with something.

(One possible bug that seems fairly likely and thus worth explicitly checking for: Does this correctly set the user as present when a user first logs in?)

abhigyank:
> I wonder if there'd be value in having a developer flag to show the current user's presence status on the window somewhere, which we can just have on while working on the desktop app over the next few weeks to validate whether there are bugs here.

Sounds right  to me. @vsvipul I think you can add a log to the webview console if the app is in development mode with comment to remove it after a few weeks (or before the next release).

>(One possible bug that seems fairly likely and thus worth explicitly checking for: Does this correctly set the user as present when a user first logs in?)

Here logs in could be in two form - the user just logs into the server on the app or opens the app from start. Second the app is open but the user wakes up from sleep and uses the app but doesn't open the app. @vsvipul Can you test out these two instances?

vsvipul:
> Sounds right to me. @vsvipul I think you can add a log to the webview console if the app is in development mode with comment to remove it after a few weeks (or before the next release).

DId this. Now in development mode, we get a status log in console every 15 seconds.
 
> Here logs in could be in two form - the user just logs into the server on the app or opens the app from start. Second the app is open but the user wakes up from sleep and uses the app but doesn't open the app. @vsvipul Can you test out these two instances?

Tested out both. Both work fine.

app/main/index.ts
```diff
@@ -362,6 +362,21 @@ app.on('ready', () => {
...
 		ConfigUtil.setConfigItem('lastActiveTab', index);
 	});
+
+	// Update user idle status for each realm after every 15s
+	// Set user idle if no activity for idleThreshold time in seconds
+	// Do a check every idleCheckInterval ms.
+	const idleCheckInterval = 15000;
+	setInterval(() => {
+		const idleThreshold = 60;
+		(electron.powerMonitor as any).querySystemIdleState(idleThreshold, (idleState: string) => {
```
priyank-p: This is wrong; why do you need to type cast this to `any`?

vsvipul: Apparently, querySystemIdleState has not been added yet to types of powerMonitor. tsc gives the following error-
```
app/main/index.ts:372:25 - error TS2339: Property 'querySystemIdleState' does not exist on type 'PowerMonitor'.

372  	electron.powerMonitor.querySystemIdleState(idleThreshold, (idleState: string) => {
     	                      ~~~~~~~~~~~~~~~~~~~~


Found 1 error.

```

priyank-p: OK, Add a comment here so we can come back to it in future.

vsvipul: Done.

abhigyank:
Looks good to me. Could be merged after a final review by Tim or Akash.

akashnimare:
@vsvipul can you fix the conflicts so that I can review and merge?

vsvipul:
@akashnimare Fixed the conflicts.

akashnimare:
@vsvipul is this ready to merge? Also, can you tell me how to test and what's the desired behaviour? I guess the webapp part is already merged.

vsvipul:
> You can test by console logging the two variables in the setInterval near line 97 of preload.js. They will update based on user activity.

This variables are then picked up by the webapp to set the user status. @akashnimare

priyank-p:
This works pretty well, Thanks @vsvipul for implementing this!

PR #750, opened 2019-05-24
Update server validation logic.

### LINKED ISSUE #573: Stop checking for zulip.ogg if error from the server is a 400 "BAD_REQUEST" error

### LINKED ISSUE #596: No error when trying to add an instance without organisation
**What's wrong** 
Zulip desktop client doesn't display any error or information when trying to add a Zulip instance without any Organisations in it.

**Steps to reproduce**
* Start a fresh, properly configured Zulip instance. **Do not create any organisation**
* Open the Electron app  
* Try to add this instance

**Expected**
Zulip should give a clear, helpful message so the user understands he needs to create the org first / ask the Zulip instance admin to create the org.

**Possible resolutions**
https://github.com/zulip/zulip-electron/blob/925fec71d5542dd124e0b7065023195dac364e12/app/renderer/js/utils/domain-util.js#L215 is looking for the realm icon and not expecting it to be missing.  
Adding a `else` block here to render an alert would suffice.  

**OS**
Mac OS Mojave 10.14.1

I can PR this if needed. 

**EDIT**: I'm leaving https://zulip2.superbiche.co (my testing instance without org) as-is so you can easily reproduce this behavior. I'll shut it down when we've fixed it and checked it works.

kanishk98:
**What's this PR do?**

1. Requests `zulip.ogg` as a fallback option in the case of HTTP 404 errors, preferring to use `getServerSettings` first. Fixes #573 and prevents `typo.zulipchat.com` from getting added. 

Screenshot from 2019-05-25 03-12-23

2. Shows error message when trying to add a Zulip server with no organisations. Fixes #596.

image

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/utils/domain-util.js
+++ b/app/renderer/js/utils/domain-util.js
@@ -10,6 +10,7 @@ const escape = require('escape-html');
 const Logger = require('./logger-util');
 
 const RequestUtil = require(__dirname + '/../utils/request-util.js');
+const Messages = require(__dirname + '/../../../resources/messages.js');
 
 const logger = new Logger({
 	file: `domain-util.log`,
@@ -101,19 +102,50 @@ class DomainUtil {
 		return false;
 	}
 
+	async checkCertError(domain, serverConf, error, silent) {
+		if (silent) {
+			// since getting server settings has already failed
+			return serverConf;
+		} else {
+			// Report error to sentry to get idea of possible certificate errors
+			// users get when adding the servers
+			logger.reportSentry(new Error(error));
+			const certErrorMessage = Messages.certErrorMessage(domain, error);
+			const certErrorDetail = Messages.certErrorDetail();
+
+			const response = await dialog.showMessageBox({
+				type: 'warning',
+				buttons: ['Yes', 'No'],
+				defaultId: 1,
+				message: certErrorMessage,
+				detail: certErrorDetail
+			});
+			if (response === 0) {
+				// set ignoreCerts parameter to true in case user responds with yes
+				serverConf.ignoreCerts = true;
+				try {
+					return await this.getServerSettings(domain, serverConf.ignoreCerts);
+				} catch (err) {
+					if (error === Messages.noOrgsError(domain)) {
+						throw new Error(error);
+					}
+					return serverConf;
+				}
+			} else {
+				throw new Error('Untrusted certificate.');
+			}
+		}
+	}
+
 	// ignoreCerts parameter helps in fetching server icon and
 	// other server details when user chooses to ignore certificate warnings
-	checkDomain(domain, ignoreCerts = false, silent = false) {
+	async checkDomain(domain, ignoreCerts = false, silent = false) {
 		if (!silent && this.duplicateDomain(domain)) {
 			// Do not check duplicate in silent mode
-			return Promise.reject('This server has been added.');
+			throw new Error('This server has been added.');
 		}
 
 		domain = this.formatUrl(domain);
-		const checkDomain = {
-			url: domain + '/static/audio/zulip.ogg',
-			...RequestUtil.requestOptions(domain, ignoreCerts)
-		};
 
 		const serverConf = {
 			icon: defaultIconUrl,
@@ -122,70 +154,29 @@ class DomainUtil {
 			ignoreCerts
 		};
 
-		return new Promise((resolve, reject) => {
-			request(checkDomain, (error, response) => {
-				// If the domain contains following strings we just bypass the server
-				const whitelistDomains = [
-					'zulipdev.org'
-				];
+		try {
+			return await this.getServerSettings(domain, serverConf.ignoreCerts);
+		} catch (err) {
+			// If the domain contains following strings we just bypass the server
+			const whitelistDomains = [
+				'zulipdev.org'
+			];
 
-				// make sure that error is an error or string not undefined
-				// so validation does not throw error.
-				error = error || '';
+			// make sure that error is an error or string not undefined
+			// so validation does not throw error.
+			const error = err || '';
 
-				const certsError = error.toString().includes('certificate');
-				if (!error && response.statusCode < 400) {
-					// Correct
-					this.getServerSettings(domain, serverConf.ignoreCerts).then(serverSettings => {
-						resolve(serverSettings);
-					}, () => {
-						resolve(serverConf);
-					});
-				} else if (domain.indexOf(whitelistDomains) >= 0 || certsError) {
-					if (silent) {
-						this.getServerSettings(domain, serverConf.ignoreCerts).then(serverSettings => {
-							resolve(serverSettings);
-						}, () => {
-							resolve(serverConf);
-						});
-					} else {
-						// Report error to sentry to get idea of possible certificate errors
-						// users get when adding the servers
-						logger.reportSentry(new Error(error));
-						const certErrorMessage = `Do you trust certificate from ${domain}? \n ${error}`;
-						const certErrorDetail = `The organization you're connecting to is either someone impersonating the Zulip server you entered, or the server you're trying to connect to is configured in an insecure way.
-						\nIf you have a valid certificate please add it from Settings>Organizations and try to add the organization again.
-						\nUnless you have a good reason to believe otherwise, you should not proceed.
-						\nYou can click here if you'd like to proceed with the connection.`;
-
-						dialog.showMessageBox({
-							type: 'warning',
-							buttons: ['Yes', 'No'],
-							defaultId: 1,
-							message: certErrorMessage,
-							detail: certErrorDetail
-						}, response => {
-							if (response === 0) {
-								// set ignoreCerts parameter to true in case user responds with yes
-								serverConf.ignoreCerts = true;
-								this.getServerSettings(domain, serverConf.ignoreCerts).then(serverSettings => {
-									resolve(serverSettings);
-								}, () => {
-									resolve(serverConf);
-								});
-							} else {
-								reject('Untrusted Certificate.');
-							}
-						});
-					}
-				} else {
-					const invalidZulipServerError = `${domain} does not appear to be a valid Zulip server. Make sure that \
-					\n (1) you can connect to that URL in a web browser and \n (2) if you need a proxy to connect to the Internet, that you've configured your proxy in the Network settings \n (3) its a zulip server \
-					\n (4) the server has a valid certificate, you can add custom certificates in Settings>Organizations`;
-					reject(invalidZulipServerError);
+			const certsError = error.toString().includes('certificate');
+			if (domain.indexOf(whitelistDomains) >= 0 || certsError) {
+				try {
+					return await this.checkCertError(domain, serverConf, error, silent);
+				} catch (err) {
+					throw err;
 				}
-			});
-		});
+			} else {
+				throw Messages.invalidZulipServerError(domain);
+			}
+		}
 	}
 
 	getServerSettings(domain, ignoreCerts = false) {
@@ -207,9 +198,11 @@ class DomainUtil {
 							alias: escape(data.realm_name),
 							ignoreCerts
 						});
+					} else {
+						reject(Messages.noOrgsError(domain));
 					}
 				} else {
-					reject('Zulip server version < 1.6.');
+					reject(response);
 				}
 			});
 		});
--- a/app/resources/messages.js
+++ b/app/resources/messages.js
@@ -1,0 +1,27 @@
+class Messages {
+	invalidZulipServerError(domain) {
+		return `${domain} does not appear to be a valid Zulip server. Make sure that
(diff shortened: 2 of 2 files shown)


akashnimare:
@kanishk98 what happens in case of `abcd.zulip.org`?

app/renderer/js/utils/domain-util.js
```diff
@@ -101,6 +101,48 @@ class DomainUtil {
...
+				resolve(serverConf);
+			} else {
+				// Report error to sentry to get idea of possible certificate errors
+				// users get when adding the servers
+				logger.reportSentry(new Error(error));
+				const certErrorMessage = `Do you trust certificate from ${domain}? \n ${error}`;
+				const certErrorDetail = `The organization you're connecting to is either someone impersonating the Zulip server you entered, or the server you're trying to connect to is configured in an insecure way.
+				\nIf you have a valid certificate please add it from Settings>Organizations and try to add the organization again.
+				\nUnless you have a good reason to believe otherwise, you should not proceed.
+				\nYou can click here if you'd like to proceed with the connection.`;
```
priyank-p: Why is there a need to use template string literals and `\n`? This will mess up the formatting by adding double new lines.

priyank-p (commented):
Reading through it LGTM! Though we can however split the check for ogg into its own function like you did with for the cert error case since this whole logic seems to getting more complicated and hard to read through and follow.

app/renderer/js/utils/domain-util.js
priyank-p: Also, I suggest you don't use template literal here since the indentation will show up. They are generally best for multiline string on top level.

app/renderer/js/utils/domain-util.js
```diff
@@ -101,6 +101,48 @@ class DomainUtil {
 		return false;
 	}
 
+	checkCertError(domain, serverConf, error, silent) {
+		return new Promise((resolve, reject) => {
```
priyank-p: Make this a async function that way you don't need to return a promise.

app/renderer/js/utils/domain-util.js
```diff
@@ -122,69 +160,59 @@ class DomainUtil {
 			ignoreCerts
 		};
 
-		return new Promise((resolve, reject) => {
-			request(checkDomain, (error, response) => {
+		return new Promise(async (resolve, reject) => {
```
priyank-p: Just make the `checkDomain` function async. You will to throw error where there is `Promise.reject()`.

priyank-p:
@kanishk98 if you interested in `async` conversion type of thing, we can use `request-promise` inside async functions and significantly reduced the callbacks we have and the code really readable and easy to follow, since we don't have to use lots of callback thus reduce the indentation and complex chaining we do.

app/renderer/js/utils/domain-util.js
kanishk98: > Why is there a need to use template string literals and \n?

I can't answer that - I just copied over the messages from the old method. 
I'll change this in a separate commit.

kanishk98:
> what happens in case of abcd.zulip.org?

Thanks for mentioning this! I missed out on a bug - I was checking for certificate errors even if the `error` returned from `getServerSettings()` was `undefined`, so `abcd.zulip.org` got stuck in that. 

I'll change the logic to only check for certificate errors when:
1. Either the `error` message includes `certificate`,
2. The domain is whitelisted.

kanishk98:
> if you interested in async conversion type of thing, we can use request-promise inside async functions and significantly reduced the callbacks we have and the code really readable and easy to follow, since we don't have to use lots of callback thus reduce the indentation and complex chaining we do

@priyank-p that sounds like a good idea. The code's getting difficult to follow, and we should take this up along with our conversion to TypeScript if the others agree. 

Let's open a topic on czo to discuss this?

app/renderer/js/utils/domain-util.js
priyank-p: Ah, can you try seeing if it affects the formatting of the dialog that shows up?

kanishk98: No, there's no change. 

Here's the dialog generated by the PR:

mine

Here's `master`:

Screenshot from 2019-05-28 02-47-49

The difference in the error message is because of the change in the first `request` made.

priyank-p: Cool! The indentation will show up if you don't use `\n`.

kanishk98:
@akashnimare fixed the faulty test case. Here are the URLs I tried before updating this PR, all of which worked as expected:

1. `chat.zulip.org`
2. `abhigyank.zulipdev.org`
3. My own Zulip instance, with no orgs
4. `nucleus.zulipchat.com` (a real organisation)
5. `abcd.zulipchat.com`
6. `abcd.zulip.org`

@priyank-p I'm working on the `async-await` conversions right now, but I'm a little unsure about whether they should be part of this PR or become an independent change. 
I'll write them regardless, let me know what you think.

priyank-p:
You should do the async conversation here for the functions you changed in this PR in a seperate commit. Rest of them you should do it in a new PR.

app/renderer/js/utils/domain-util.js
```diff
@@ -101,54 +102,80 @@ class DomainUtil {
...
+
+	async checkOggFile({ serverConf, domain, whitelistDomains, ignoreCerts, silent }) {
+		const oggCheck = {
+			url: domain + '/static/audio/zulip.ogg',
+			...RequestUtil.requestOptions(domain, ignoreCerts)
+		};
+		try {
+			const response = await requestPromise(oggCheck);
+			if (response.statusCode === 200) {
+				// getting serverSettings has already failed
```
priyank-p: Why does this comment states something failed inside a 200 request condition?
Also I recommend checking range of good HTTP status i.e. 200 to but not including 400 since cached responses return 302 Not Modified.

app/renderer/js/utils/domain-util.js
```diff
@@ -101,54 +102,80 @@ class DomainUtil {
...
+				\n (4) the server has a valid certificate, you can add custom certificates in Settings>Organizations`;
+				return invalidZulipServerError;
+			}
+		}
 	}
 
 	// ignoreCerts parameter helps in fetching server icon and
 	// other server details when user chooses to ignore certificate warnings
-	checkDomain(domain, ignoreCerts = false, silent = false) {
+	async checkDomain(domain, ignoreCerts = false, silent = false) {
```
priyank-p, line 142: I really liked reading the change for this function, this and other changes in this commit looks great. This version if far more readable.

priyank-p (commented):
@kanishk98 Thanks for doing the async migration. It looks great! Left one comment which might not be the change you made directly.

app/renderer/js/utils/domain-util.js
kanishk98: > Why does this comment states something failed inside a 200 request condition?

I wanted to clarify why we're returning `serverConf` instead of `serverSettings`. Since we already tried retrieving the latter and got an error, we need to return `serverConf`. 

> I recommend checking range of good HTTP status i.e. 200 to but not including 400 since cached responses return 302 Not Modified.

That's a good catch, thanks. I'll update the PR.

app/renderer/js/utils/domain-util.js
kanishk98: Awesome, thanks! :)
Let me get a few more pressing issues out of the way and I'll begin work on converting all the code in the app to `async await`. Will let you know when I start.

kanishk98:
@akashnimare I've made the required changes. Please have a look whenever you find time.

app/renderer/js/utils/domain-util.js
abhigyank: Can change the comment to explain this better and include the statement you mentioned.

kanishk98:
As suggested by Tim in https://github.com/zulip/zulip-desktop/issues/573#issuecomment-498536198, I've removed the `.ogg` file check entirely. 

I also noticed that we were using a simple `alert()` for our error message in case the server the user connects with is invalid, so I changed that to a proper error box and changed the grammar a little. 
I also mentioned that we no longer Zulip server versions older than 1.6.

Here's what it looks like:

image

app/renderer/js/utils/domain-util.js
kanishk98: @abhigyank marking this as resolved because we've decided to do away with the .ogg file check entirely, so there's no need to change the comments.

kanishk98:
@akashnimare whenever you have time, please take a look at this again. I think this is ready to merge.

app/renderer/js/utils/domain-util.js
```diff
@@ -122,70 +157,37 @@ class DomainUtil {
...
+			if (domain.indexOf(whitelistDomains) >= 0 || certsError) {
+				try {
+					return await this.checkCertError(domain, serverConf, error, silent);
+				} catch (err) {
+					throw err;
 				}
-			});
-		});
+			} else {
+				const invalidZulipServerError = {
```
akashnimare: @priyank-p @abhigyank does it make sense to define this in a separate file and import it from there to make the code cleaner? Same with `certErrorMessage`, `certErrorDetail`. Code is getting cluttered because of these strings etc.

priyank-p: Yeah, sounds good to me.

kanishk98: Cool. I'll add this file to the `resources` folder.

kanishk98:
@akashnimare made a new file - `resources/messages.js`. We can use this to clean up our code as we go along. Currently it only has the error messages that this PR is concerned with.

akashnimare:
Maybe this could go to `\utils\`? cc - @abhigyank.

kanishk98:
Hmm, I didn't think it was a good idea because the `utils` folder is inside `renderer` and I figured we could also use the same file for the main process. 
Regardless, that's a rather minor change at this stage.

abhigyank:
`resources` seem fine to me. Its not much of a `util` and regardless `util` is getting to many files too! :stuck_out_tongue:

app/renderer/js/pages/preference/new-server-form.js
```diff
@@ -58,7 +59,7 @@ class NewServerForm extends BaseComponent {
 			});
 		}, errorMessage => {
 			this.$saveServerButton.children[0].innerHTML = 'Connect';
-			alert(errorMessage);
+			dialog.showErrorBox(errorMessage.title, errorMessage.content);
```
abhigyank: Have you tested this? Error returned from Line 145 doesn't have title or content keys in it.

abhigyank (changes requested):
Some issue with the error message - 
Screenshot from 2019-06-11 11-40-20

app/renderer/js/pages/preference/new-server-form.js
kanishk98: Ah okay, I must've missed this while testing. 
Thanks, I'll check all error messages and make sure the title and content keys are present in each error message. 

Will edit this commit directly.

kanishk98: Okay, so pretty much the only clean way I know of to do this is to create new `Error` objects and then assign the properties `title` and `content` to them, which I feel is overkill for a feature that wasn't required in the first place (I just didn't like using `alert`). 

If you agree, then I can just drop this commit and go back to using `alert` with simple error strings. 

cc @akashnimare

kanishk98: (Or we could create message objects with `title` and `content` props, `stringify` them when throwing the error, and then parse them on the receiving end, but I think this is really messy).

akashnimare: Alert sounds okay to me.

akashnimare:
@kanishk98 can you take care of the above issue?

kanishk98:
@akashnimare dropped the change from `alert` to `dialog` and fixed the error messages in `messages.js`. 
Let me know if any other changes are required here.

abhigyank (approved):
Wasn't able to test the `noOrgsError` but everything else works and looks fine to me. GTM.

vsvipul:
LGTM. Good though on creating a separate ```messages``` util.

vsvipul (approved):
LGTM. Good thought on creating a separate ```messages``` util.

akashnimare:
Merged. Thanks, everyone 🚀

PR #752, opened 2019-05-30
spellchecker: Improve auto detection of spellchecker language.

kanishk98:
**What's this PR do?**

Adds options for configuring the spellchecker in Settings > General. 

There are 3 levels for settings:
1. Auto-detect and correct (default)
2. Set `spellcheckerLanguage` to the realm language (org-level)
3. Allow user to override (1) and (2) with custom language (app-level). 

**Screenshots?**

Fixed setting, overriding both input text and server language:

Override

Organization-specific setting:

Switch

Auto-detect works as expected. 

  - [x] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/preload.js
+++ b/app/renderer/js/preload.js
@@ -3,7 +3,6 @@
 const { ipcRenderer, shell } = require('electron');
 const SetupSpellChecker = require('./spellchecker');
 
-const ConfigUtil = require(__dirname + '/utils/config-util.js');
 const LinkUtil = require(__dirname + '/utils/link-util.js');
 const params = require(__dirname + '/utils/params-util.js');
 
@@ -47,10 +46,8 @@ document.addEventListener('DOMContentLoaded', () => {
 	// Get the default language of the server
 		const serverLanguage = page_params.default_language; // eslint-disable-line no-undef, camelcase
 		if (serverLanguage) {
-			// Set spellcheker language
-			ConfigUtil.setConfigItem('spellcheckerLanguage', serverLanguage);
 			// Init spellchecker
-			SetupSpellChecker.init();
+			SetupSpellChecker.init(serverLanguage);
 		}
 		// redirect users to network troubleshooting page
 		const getRestartButton = document.querySelector('.restart_get_events_button');
--- a/app/renderer/js/spellchecker.js
+++ b/app/renderer/js/spellchecker.js
@@ -11,11 +11,11 @@ const logger = new Logger({
 });
 
 class SetupSpellChecker {
-	init() {
+	init(serverLanguage) {
 		if (ConfigUtil.getConfigItem('enableSpellchecker')) {
 			this.enableSpellChecker();
 		}
-		this.enableContextMenu();
+		this.enableContextMenu(serverLanguage);
 	}
 
 	enableSpellChecker() {
@@ -26,13 +26,13 @@ class SetupSpellChecker {
 		}
 	}
 
-	enableContextMenu() {
+	enableContextMenu(serverLanguage) {
 		if (this.SpellCheckHandler) {
 			this.SpellCheckHandler.attachToInput();
-
-			const userLanguage = ConfigUtil.getConfigItem('spellcheckerLanguage');
-
-			this.SpellCheckHandler.switchLanguage(userLanguage);
+			this.SpellCheckHandler.switchLanguage(serverLanguage);
+			this.SpellCheckHandler.currentSpellcheckerChanged.subscribe(() => {
+				this.SpellCheckHandler.switchLanguage(this.SpellCheckHandler.currentSpellcheckerLanguage);
+			});
 		}
 
 		const contextMenuBuilder = new ContextMenuBuilder(this.SpellCheckHandler);


akashnimare:
Quick review -

image

In this case, we should disable the choose spellchecker option?

kanishk98:
No, I chose to grey it out because the user should be able to set a new language then. If they do that, then the server-level setting gets disabled. 
And complete disabling of the third option happens when we disable the spellchecker itself.

akashnimare:
For the language options, instead of the code i.e. - 
image


We should show something like -
image

image

kanishk98:
Yeah, I plan to use an npm module for this, like you suggested. [locale-code](https://www.npmjs.com/package/locale-code) would be my choice.

akashnimare:
> And complete disabling of the third option happens when we disable the spellchecker itself.

image

In this case, the third option should be disabled (cursor: not-allowed) and also need to take care of bg color on hover.

akashnimare:
@priyank-p, @abhigyank have a look whenever you have time.

kanishk98:
Alright, I've made some progress:

1. Red lines appear quickly enough for me now, looks like I was making some redundant function calls. 
2. Spellchecker language works as expected now on switching orgs (provided the relevant setting is on)
3. The `not-allowed` cursor and grey background as suggested by @akashnimare for the app-level setting is now present. 

What I'm improving now:
1. 13e5586e109ba23d030183b5831af0c2fa3f0f70 is not the cleanest it can be. I'd really like some suggestions on handling this commit. 
2. Using `electron-spellchecker` to handle fallback locales instead of a `.json` file.
3. Translating locale codes to proper language names. The module I suggested in https://github.com/zulip/zulip-desktop/pull/752#issuecomment-497508840 doesn't seem to be great for that, but I'll continue looking around.

kanishk98:
I think this PR's ready for a review now. 
@priyank-p as discussed earlier, I wrote a script to map locale codes to language names and created a `.json` file. I mapped names to codes as well so we wouldn't have to search through the whole set of languages when saving the spellchecker setting.

I've uploaded said script as a Gist [here](https://gist.github.com/kanishk98/cc5c5fa5cdf65b58b215986a883c74e9) in case you want to check that too.

@rrhythmsharma since you've worked with the spellchecker and were interested in this PR, feel free to have a look and point out problems. :)

priyank-p:
@kanishk98 cool! An suggestion, if you want to do it, you can create a npm package for doing what we need to do. I think other people might be interested in the package if they are using electron-spellcheck.

kanishk98:
> An suggestion, if you want to do it, you can create a npm package for doing what we need to do

Yeah, that's a good idea, thanks! I'll probably do that once this PR's approved so I'm sure I know what I'm doing. :P

app/renderer/js/pages/preference/general-section.js
```diff
@@ -76,10 +78,21 @@ class GeneralSection extends BaseSection {
 						<div class="setting-description">Always start minimized</div>
 						<div class="setting-control"></div>
 					</div>
+				</div>
+				<div class="title">Spellchecker (requires restart)</div>
+				<div class="settings-card">
 					<div class="setting-row" id="enable-spellchecker-option">
-						<div class="setting-description">Enable spellchecker (requires restart)</div>
+						<div class="setting-description">Enable spellchecker</div>
```
abhigyank: Probably add `(auto-detect language)` or something like to that to this statement.

app/renderer/js/main.js
```diff
@@ -432,6 +434,13 @@ class ServerManagerView {
 		} catch (err) {
 		}
 
+		if (ConfigUtil.getConfigItem('useServerLanguage', false)) {
+			const newLanguage = DomainUtil.getServerLanguage(this.tabs[index].webview.props.url);
+			if (newLanguage) {
```
abhigyank: Can there be cases when `newLanguage` might not exist? I see that we are storing the language in preload.js.

abhigyank (commented):
The overall approach looks fine to me. Though we probably need to improve the way we are writing some pieces of code.
I have not gone through the entire changes, hope to do it once I get time soon enough.

akashnimare:
> The overall approach looks fine to me. Though we probably need to improve the way we are writing some pieces of code.

@kanishk98 can you address @abhigyank's comment and go through the code once. The feature works fine for me. @priyank-p needs your input on this too.

app/renderer/js/pages/preference/base-section.js
```diff
@@ -16,6 +16,18 @@ class BaseSection extends BaseComponent {
 		$optionControl.addEventListener('click', clickHandler);
 	}
 
+	generateSettingDropdown(props) {
+		const {$element, value, options, clickHandler} = props;
```
priyank-p: Nit: add spaces around braces.

app/renderer/js/pages/preference/base-section.js
```diff
@@ -38,6 +50,22 @@ class BaseSection extends BaseComponent {
 		}
 	}
 
+	generateDropdownOption(dropdownOption, value) {
+		if (dropdownOption === value) {
+			return `
+				<option class="dropdown-option selected" selected>
```
priyank-p: You can just do, `${dropdownOption === value ? "selected" : ""}` inside the template literal here.

app/renderer/js/preload.js
```diff
@@ -7,6 +7,8 @@ const ConfigUtil = require(__dirname + '/utils/config-util.js');
 const LinkUtil = require(__dirname + '/utils/link-util.js');
 const params = require(__dirname + '/utils/params-util.js');
 
+const fallbackLocales = require(__dirname + '/../../resources/spellcheck/fallback.json');
```
priyank-p: Hmm, so why do you call this fallback? I feel like there could be a better name for this

app/renderer/js/spellchecker.js
```diff
@@ -29,10 +29,15 @@ class SetupSpellChecker {
 	enableContextMenu() {
 		if (this.SpellCheckHandler) {
 			this.SpellCheckHandler.attachToInput();
-
-			this.SpellCheckHandler.currentSpellcheckerChanged.subscribe(() => {
-				this.SpellCheckHandler.switchLanguage(this.SpellCheckHandler.currentSpellcheckerLanguage);
-			});
+			if (ConfigUtil.getConfigItem('useCustomLanguage', false) || ConfigUtil.getConfigItem('useServerLanguage', false)) {
```
priyank-p: I'd move both conditions into a variable and do:
```javascript
const usingCustomLanguage = Config....;
const usingServerLanguague = Config....;
if (usingCustomLanguage || usingServerLanguage) {

```

app/renderer/js/pages/preference/general-section.js
priyank-p: Hmm, I don't think we need to add something like that since in terms of UX. And if we feel strongly about adding something I think "(language is auto-detected)" is better wording for this.

kanishk98: Okay, it's probably better to say more than to say nothing and avoid confusion, so I'll add spellchecker (language is auto-detected by default) for now.

app/renderer/js/main.js
kanishk98: Yes, we're storing the language, but `newLanguage` is fetched using `DomainUtil` and I wanted to be sure that this is not `null` when changing the language to that.

app/renderer/js/preload.js
kanishk98: That's the same language as used by the spellchecker module. I figured that we should use the same thing. 
We can call it `normalizedLocales` if that sounds better to you.

app/renderer/js/pages/preference/base-section.js
kanishk98: This was a pre-existing thing - maybe we should update our style check rules to disallow no spaces around braces?

akashnimare: @kanishk98 we used to have it I think.

app/renderer/js/preload.js
akashnimare: `normalizedLocales` sounds better.

kanishk98:
I've made the required changes. Thanks for the reviews!
@abhigyank I'm going through the code again, will try to improve the structure wherever I can. Otherwise we can probably pick this up in a refactor later if you find any problems.

kanishk98:
@akashnimare I think this PR is pretty much ready now - Priyank's suggestions helped clean up the code too. Let me know what you think.

akashnimare:
Works for me. @rrhythmsharma, @vsvipul can you cross check this on Linux/Windows?

rhythm-sharma:
Sure,  will try to do it asap

rishig:
Added a comment at https://chat.zulip.org/#narrow/stream/16-desktop/topic/spellchecker.20options. 
A small thing is the title should be `Spellchecker (requires app restart)` rather than just requires restart.

kanishk98:
> A small thing is the title should be Spellchecker (requires app restart) rather than just requires restart.

Changed the concerned commit. :)

kanishk98:
@akashnimare merged the settings as discussed on czo. 
The field also grays out and disallows clicking if the spellchecker is disabled.

app/renderer/js/main.js
```diff
@@ -432,6 +434,13 @@ class ServerManagerView {
...
+				ConfigUtil.setConfigItem('spellcheckerLanguage', newLanguage);
```
abhigyank: `activateTab` is called multiple times throughout the app - so does for everytime we would be updating `spellcheckerLanguage` with the domains language? Is this necessary?

abhigyank:
There are two issues I see here - 
* If you choose organisation language - the settings get updated with the language of the last opened server.
* The implementation of `spellchecker.js`  for initialisation depends on `spellcheckerLanguage` for each server - and `spellcheckerLanguage` is being changed on initialisation of servers in `main.js` this would cause for each server to have different languages then expected because both things don't occur in sync.

Can anyone else see these issues?

app/renderer/js/main.js
kanishk98: Well, we need to know in `spellchecker.js` which language to switch to every time the tab is switched. We could also update something like `currentTabIndex` and then get the server language using that in the spellchecker file, but that wouldn't be any different. 

If this setting of org level spellchecker is set, we'll have to change language on a tab switch - I don't see a way around that. (Would love to be told otherwise though).

rishig:
Should we resolve the discussion at https://chat.zulip.org/#narrow/stream/16-desktop/topic/spellchecker.20options before going too far down this path? E.g. there's some chance this should just be a webapp setting.

app/renderer/js/main.js
akashnimare: Can't we get the server language from preload.js using `page_params.default_language` ?

kanishk98: Yeah, that's a good idea. I just checked and the spellchecker works fine with the tab switch with just using `preload.js`. 

Looks like I'll have to refactor this a fair bit, but this should fix Abhigyan's concerns in https://github.com/zulip/zulip-desktop/pull/752#issuecomment-500694499

vsvipul:
Adding to the issues mentioned by @abhigyank , -
1. Instead of saying 'Automatically detect language', we could say 'Detect automatically', so that the the word 'language' isn't cut off.
Screenshot from 2019-06-12 22-15-28
2. The language list should be sorted by name. Currently, this isn't the case.

Tested this, worked fine but there is some delay in case of auto-detection of language.

kanishk98:
Thanks for the reviews!

@vsvipul I've sorted the locale codes according to their language names and also changed the wording of the option as you suggested. 

> Tested this, worked fine but there is some delay in case of auto-detection of language.

Hmm, I noticed that it becomes noticeably slow when you're intentionally typing in garbage, but that's where the other 2 settings for language can help the user. This just feels like a sensible default.

kanishk98:
@akashnimare I've reverted to the old UI and followed Tim and Rishi's suggestion of including only one option, with on meaning autodetect paired with server language. Please review it when you have time.

kanishk98:
Should we need the custom language overrides or additional options again, I've backed up all the related work for that in [here](https://github.com/kanishk98/zulip-desktop/tree/archive-spellchecker) on my fork. We can always refer to that later, depending on user feedback.

akashnimare:
Merged. Thanks, everyone.

app/renderer/js/spellchecker.js
```diff
@@ -26,13 +26,13 @@ class SetupSpellChecker {
...
-	enableContextMenu() {
+	enableContextMenu(serverLanguage) {
 		if (this.SpellCheckHandler) {
 			this.SpellCheckHandler.attachToInput();
-
-			const userLanguage = ConfigUtil.getConfigItem('spellcheckerLanguage');
-
-			this.SpellCheckHandler.switchLanguage(userLanguage);
+			this.SpellCheckHandler.switchLanguage(serverLanguage);
+			this.SpellCheckHandler.currentSpellcheckerChanged.subscribe(() => {
```
vsvipul, line 33: @kanishk98  Can you help me understand what this does? I need to know this for BrowserView. Getting problems due to this.

vsvipul: @kanishk98  Can you help me understand what this does? I need to know this for BrowserView. Getting problems due to this.

kanishk98: Sure. 

This basically tells the spellchecker module to keep track of changes in the input text and if the total body of text is in a different language than the one the module assumed in the beginning, then it applies spellcheck according to the new language.

kanishk98: I think it'll be better for you if you check out the code of this class, but I doubt if the module itself is causing problems. 

Can you tell me what's causing trouble? Maybe I can help.

PR #755, opened 2019-06-15
dev: Use .env file for Sentry DSN.

kanishk98:
* Reads .env file in root folder of repo to get Sentry DSN for builds.

---

**What's this PR do?**

Developers working on the Zulip desktop app often have to build packaged versions of the app, and I personally often forget to replace the `SENTRY_DSN` placeholder string with the actual DSN. This PR reads a `.env` file that'll be ignored by `git` and therefore can be used across branches for development.

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/utils/sentry-util.ts
+++ b/app/renderer/js/utils/sentry-util.ts
@@ -1,11 +1,14 @@
 import { init } from '@sentry/electron';
 
 import isDev = require('electron-is-dev');
+import path = require('path');
+import dotenv = require('dotenv');
+dotenv.config({ path: path.resolve(__dirname, '/../../../../.env') });
 
 export const sentryInit = (): void => {
 	if (!isDev) {
 		init({
-			dsn: 'SENTRY_DSN',
+			dsn: process.env.SENTRY_DSN,
 			// We should ignore this error since it's harmless and we know the reason behind this
 			// This error mainly comes from the console logs.
 			// This is a temp solution until Sentry supports disabling the console logs
--- a/app/package.json
+++ b/app/package.json
@@ -30,6 +30,7 @@
     "@sentry/electron": "0.14.0",
     "adm-zip": "0.4.11",
     "auto-launch": "5.0.5",
+    "dotenv": "8.0.0",
     "electron-is-dev": "0.3.0",
     "electron-log": "2.2.14",
     "electron-spellchecker": "1.1.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -263,6 +263,15 @@
         "@types/node": "*"
       }
     },
+    "@types/dotenv": {
+      "version": "6.1.1",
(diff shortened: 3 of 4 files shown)


priyank-p:
You should also add `.env` to `.gitignore` unless it's already added.

kanishk98:
It's already added.

akashnimare:
We can't use this while building the installers (npm run dist), right?

kanishk98:
We can, that's the reason I sent the PR.

kanishk98:
@akashnimare please let me know if anything else is required here. Thanks.

akashnimare:
Nothing requires from your side. When I last tested, I was unable to build the installers. Gotta test it again.

akashnimare:
@kanishk98 have you tested this in the packaged app?

kanishk98:
> have you tested this in the packaged app?

I had, but I've done that a couple more times to be sure, and I've also added a `path` object that points to the `.env` file (it expects the developer to put it in the root directory). 

Let me know if building the installers works for you this time.

akashnimare:
I don't think you need to add the path. Gotta cross-check though.

kanishk98:
> I don't think you need to add the path. Gotta cross-check though.

Yeah, we don't (in case it helps, the `require('dotenv').config()` command by itself worked for me both in dev and packaged versions). 
Still, I figured it couldn't hurt if the path was right. Might make sure other contributors definitely place the `.env` file in the correct folder.

app/renderer/js/utils/sentry-util.ts
```diff
@@ -1,11 +1,12 @@
 import { init } from '@sentry/electron';
 
 import isDev = require('electron-is-dev');
+require('dotenv').config({ path: __dirname + '/../../../../.env' });
```
akashnimare: @priyank-p, @vsvipul is this correct in ts?

priyank-p: Yes, it is valid though I think linter will complain. Also for consistency let's do:
```suggestion
import path = require('path');
import dotenv = require('dotenv');
dotenv.config({ path: path.resolve(__dirname, '/../../../../.env') });
```

priyank-p: Also, looking at this, let's also import path and pass this path through `path.resolve`.

kanishk98: Oh, I was just about to push the required changes.
Thanks @priyank-p!

akashnimare:
Merged. Thanks, @kanishk98 & @priyank-p.

andersk:
This doesn’t work. One mistake is that `path.resolve(__dirname, '/../../../../.env')` always resolves to `/.env` because `/../../../../.env` is an absolute path, but that isn’t the main problem. Even with that corrected, neither the `.env` file nor the environment variables it provides are available to normal users at runtime.

Fixed in #888.

PR #756, opened 2019-06-15
typescript: Add typescript support.

### LINKED ISSUE #518: Add Typescript as a type checker in the codebase
I think we should add the [flow](https://flow.org/en/) to improve the code quality.

vsvipul:
Adds typescript support using tsc for both main and renderer process.

Need to figure out the lint-js part using xo.

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS


tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compilerOptions": {
+    "target": "es5",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
```
priyank-p: We want to use `esnext` here because we don't want the javascript files to be transpliled for older version, electron always supports newer javascript syntax.

tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+    "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
+    "strict": true,                           /* Enable all strict type-checking options. */
+    "esModuleInterop": true,                   /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */
+    "noImplicitAny": false
```
priyank-p: This should be `true` to make sure we take full advantage of typescript and let it complain about where we should do type annotations.

typings.d.ts
```diff
@@ -0,0 +1 @@
+declare module '*';
```
priyank-p, line 5: Why is this required? The reason I am asking is to know what problem you faced (I know that we need to do something like this to make this work).

app/main/menu.ts
```diff
@@ -211,7 +211,7 @@ class AppMenu {
 	}
 
 	getWindowSubmenu(tabs, activeTabIndex) {
-		const initialSubmenu = [{
+		const initialSubmenu: any[] = [{
```
priyank-p, line 223: We should declare a interface for this. Using a interface would  will make sure any function consuming this will get exactly what it expected.

(The `noImplicitAny` tsconfig.json option will make sure we can't do this.)

priyank-p (commented):
@vsvipul can you split the PR into coherent commits, like each individual commits for: adding dependencies, adding tsconfig.json, migration a file (one commit per file), and for setting up the xo-typescript.

I am looking into why xo-typescript doesn't work!

priyank-p:
@vsvipul I think the xo-typescript thing is not working because you are missing `eslint-config-xo`; the documentation tells us to install that and I think that's what is responsible for adding xo-typescript config to xo. Did not have a chance to test it out yet though :/

vsvipul:
@priyank-p i installed everything mentioned in the documentation.  xo-typescript is working currently but no t with the old js rules, it is using some new rules and some new errors are popping up like these - 
Screenshot from 2019-06-16 10-06-34

tsconfig.json
vsvipul: ok sure!

tsconfig.json
vsvipul: ok gotcha !

typings.d.ts
vsvipul: This is to avoid compiler warning for having declaration files for each and every util we use.

app/main/menu.ts
vsvipul: I changed noImplicitAny to ```true``` but still could do this.

vsvipul:
@priyank-p I have split the commit to multiple commits. Working on migrating other files now.

app/main/menu.ts
```diff
@@ -1,15 +1,15 @@
...
-const { appUpdater } = require('./autoupdater');
+import * as fs from 'fs-extra';
+import * as AdmZip from 'adm-zip';
+import { appUpdater } from './autoupdater';
 
-const ConfigUtil = require(__dirname + '/../renderer/js/utils/config-util.js');
-const DNDUtil = require(__dirname + '/../renderer/js/utils/dnd-util.js');
-const Logger = require(__dirname + '/../renderer/js/utils/logger-util.js');
+import { ConfigUtil } from '../renderer/js/utils/config-util';
+import DNDUtil from '../renderer/js/utils/dnd-util';
```
akashnimare: { DNDUtil } ?

vsvipul: That hasn't been migrated to typescript for now. So, for the sake of testing, left it as it was before. Will change when i migrate DNDUtil to typescript.

app/renderer/js/utils/request-util.ts
```diff
@@ -32,7 +32,7 @@ class RequestUtil {
 		if (certificate) {
 			// To handle case where certificate has been moved from the location in certificates.json
 			try {
-				certificateLocation = fs.readFileSync(certificate);
+				certificateLocation = fs.readFileSync(certificate).toString();
```
akashnimare: Is this needed, `toString()`?

akashnimare (changes requested):

app/renderer/js/utils/system-util.ts
```diff
@@ -1,12 +1,17 @@
...
+let instance: any = null;
 
-let instance = null;
+export interface systemUtil {
+	connectivityERR: string[],
+	userAgent: any
+}
 
-class SystemUtil {
+export class systemUtil {
```
akashnimare: @priyank-p @vsvipul I don't think export is needed here, can be removed.

package.json
```diff
@@ -142,19 +145,26 @@
     "stylelint": "9.10.1",
     "tap-colorize": "1.2.0",
     "tape": "4.8.0",
-    "xo": "0.18.2"
+    "typescript": "^3.5.2",
+    "xo": "^0.24.0"
   },
   "xo": {
```
akashnimare, line 156: @priyank-p, @vsvipul do we have an option to put the `xo` config in a separate file since it's getting longer.

package.json
```diff
@@ -180,12 +190,18 @@
...
     ],
     "envs": [
       "node",
       "browser",
       "mocha"
     ]
+  },
+  "dependencies": {
+    "@types/node": "^12.0.8",
+    "@types/requestidlecallback": "^0.1.1"
```
akashnimare: @vsvipul not sure what's this?

akashnimare (changes requested):

app/renderer/js/utils/request-util.ts
priyank-p: Yes, you need to do that otherwise this would be a buffer. Another way which is think makes more sense is to pass `utf8` as the encoding parameter.

package.json
priyank-p: No, `xo` doesn't provide a way to use a config file. Though we can use a `.eslintrc` I think.

package.json
priyank-p: This was added because @vsvipul was getting a error when he migrated a ts that used this function. I will look into this.

app/main/menu.ts
akashnimare: Okay, sounds good.

priyank-p:
@vsvipul I fixed the linting errors we were getting out of nowhere by upgrading `xo` to `v0.22.0`. It is pretty weird that by just upgrading `xo` we would start to get errors we should look into this, I think the eslint upgrade is what's causing this. Note that there are still linting error in the PR, but these are errors we'll want to fix and are coming from eslint-typescript plugin.

I merged the dependencies (into typescript branch) and tsconfig.json commits with some tweaks. I also added a commit which runs `tsc -w` when we run `npm start` still working on `gulp` part of this.

priyank-p:
(Also, I like to mention if you checkout the PR, you will need to do `git clean -d` otherwise `git` won't let you switch branches.)

app/renderer/js/utils/request-util.ts
```diff
@@ -1,19 +1,19 @@
...
 const logger = new Logger({
 	file: `request-util.log`,
 	timestamp: true
 });
 
-let instance = null;
+let instance: any = null;
 
-class RequestUtil {
+export class requestUtil {
```
priyank-p: @vsvipul why did you rename the class? It should be `RequestUtil`

vsvipul: We needed to export an instance of RequestUtil to make it work in other files, otherwise tsc was returning an error. So renamed the class and named the instance RequestUtil.

priyank-p: We can do that as `export { requestUtil as RequestUtil };`

vsvipul: oh, will change. Thanks.

app/renderer/js/utils/request-util.ts
```diff
@@ -23,7 +23,7 @@ class RequestUtil {
 
 	// ignoreCerts parameter helps in fetching server icon and
 	// other server details when user chooses to ignore certificate warnings
-	requestOptions(domain, ignoreCerts) {
+	requestOptions(domain: string, ignoreCerts: boolean) {
```
priyank-p: I added a new interface here for annotating the return value of this function before merging this commit.

app/renderer/js/utils/request-util.ts
```diff
@@ -49,7 +49,7 @@ class RequestUtil {
 		};
 	}
 
-	formatUrl(domain) {
+	formatUrl(domain: string) {
```
priyank-p: Annotated return type here.

app/renderer/js/utils/request-util.ts
```diff
@@ -49,7 +49,7 @@ class RequestUtil {
...
 		const hasPrefix = (domain.indexOf('http') === 0);
```
priyank-p: The linter was complaining about using `startsWith` here so made that change here

app/renderer/js/utils/request-util.ts
```diff
@@ -1,19 +1,19 @@
...
+import { ConfigUtil } from './config-util';
+import { SystemUtil } from './system-util';
 
 const logger = new Logger({
 	file: `request-util.log`,
 	timestamp: true
 });
 
-let instance = null;
+let instance: any = null;
```
priyank-p: change this to `null | SystemUtil`

app/renderer/js/utils/system-util.ts
```diff
@@ -1,12 +1,17 @@
 'use strict';
 
-const {app} = require('electron').remote;
+import { remote } from 'electron';
+const { app } = remote;
+import * as os from 'os';
 
-const os = require('os');
+let instance: any = null;
```
priyank-p: changed this to `null | SystemUtil`.

app/renderer/js/utils/system-util.ts
priyank-p: Removed the interface and directly annoated the class.

priyank-p: Fixed this.

app/renderer/js/utils/system-util.ts
```diff
@@ -28,22 +33,23 @@ class SystemUtil {
...
 			} else {
 				return 'Windows 10';
 			}
+		} else {
+			return '';
 		}
 	}
 
-	setUserAgent(webViewUserAgent) {
+	setUserAgent(webViewUserAgent: any) {
```
priyank-p: this should be `string`

app/renderer/js/utils/system-util.ts
```diff
@@ -28,22 +33,23 @@ class SystemUtil {
...
+		const platform: string = os.platform();
+		if (platform === 'darwin') {
 			return 'Mac';
-		}
-		if (os.platform() === 'linux') {
+		} else if (platform === 'linux') {
 			return 'Linux';
-		}
-		if (os.platform() === 'win32' || os.platform() === 'win64') {
+		} else if (platform === 'win32' || platform === 'win64') {
```
priyank-p: Removed the `win64` check

app/renderer/js/utils/system-util.ts
```diff
@@ -28,22 +33,23 @@ class SystemUtil {
 	}
 
 	getOS() {
-		if (os.platform() === 'darwin') {
+		const platform: string = os.platform();
```
priyank-p: I remove the `string` declaration there. The function return a type `NodeJS.Platform` which we want to use instead. Genral rule is you shouldn't annotate type that are implicitly annotated.

For example:
```typescript
// Do
const a: string;
a = 'aa';

// Don't
const a: string = 'aa';
```

app/renderer/js/utils/request-util.ts
```diff
@@ -1,19 +1,19 @@
-const fs = require('fs');
-const Logger = require('./logger-util');
+import * as fs from 'fs';
+import { Logger } from './logger-util';
```
priyank-p: This should be just `import Logger from './logger-util'`

priyank-p:
@vsvipul posted the thing I changed when merging the request-util and logger-util commits above.

app/renderer/js/utils/system-util.ts
vsvipul: Got it! Will keep in mind from now on.

app/renderer/js/utils/system-util.ts
vsvipul: win64 doesn't exist! So, was remove therefore. See https://nodejs.org/api/os.html#os_os_platform .

priyank-p:
Merged all of the commits with changes to the typescript branch. Thanks @vsvipul for working on this, we just finished migration of the whole `utils` directory which is awesome.

Next, let finish migrating the rest of the three files left in `app/main` directory.

abhigyank:
This is great work @vsvipul @priyank-p 🎉

PR #761, opened 2019-06-19
db: Fix JSON db errors on tab switch.

kanishk98:
---

**What's this PR do?**

Fixes errors with the `settings.json` file related to problem found [here](https://chat.zulip.org/#narrow/pm-with/53,4241,4922,8973-group/near/756485). 

We'd ideally like to see a more complete solution for this later on, this just prevents errors in a very frequent situation. 

**Screenshots?**

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/main.js
+++ b/app/renderer/js/main.js
@@ -409,8 +409,8 @@ class ServerManagerView {
 	activateLastTab(index) {
 		// Open all the tabs in background, also activate the tab based on the index
 		this.activateTab(index);
-		// Save last active tab
-		ConfigUtil.setConfigItem('lastActiveTab', index);
+		// Save last active tab via main process to avoid JSON DB errors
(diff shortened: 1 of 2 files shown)


priyank-p:
LGTM!

app/renderer/js/main.js
```diff
@@ -410,7 +410,7 @@ class ServerManagerView {
 		// Open all the tabs in background, also activate the tab based on the index
 		this.activateTab(index);
 		// Save last active tab
```
vsvipul: A comment explaining why this was required in another process would be good ! LGTM.

kanishk98: Good idea, thanks. We can remove the comment once the db problems are fixed.

akashnimare:
Merged. Thanks, everyone.

PR #764, opened 2019-06-20
typescript: Migrate app/ dir to typescript.

vsvipul:
  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/autoupdater.ts
+++ b/app/main/autoupdater.ts
@@ -1,18 +1,20 @@
 'use strict';
-const { app, dialog, shell } = require('electron');
-const { autoUpdater } = require('electron-updater');
-const isDev = require('electron-is-dev');
+import { app, dialog, shell } from 'electron';
+import { autoUpdater } from 'electron-updater';
+import * as isDev from 'electron-is-dev';
+import log from 'electron-log';
 
-const ConfigUtil = require('./../renderer/js/utils/config-util.js');
+import { linuxUpdateNotification } from './linuxupdater';	// Required only in case of linux
 
-function appUpdater(updateFromMenu = false) {
+import ConfigUtil = require('../renderer/js/utils/config-util');
+
+export function appUpdater(updateFromMenu = false): void {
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "lib": ["dom", "dom.iterable", "esnext"],
     "module": "commonjs",
     "esModuleInterop": true,
+    "skipLibCheck": true,
```
vsvipul: This was done to prevent scanning of node_modules by tsc. It was returning an error in some node_module.

package.json
```diff
@@ -193,9 +193,11 @@
           "padding-line-between-statements": 0,
           "quotes": 0,
           "unicorn/catch-error-name": 0,
+          "unicorn/no-console-spaces": 0,
```
vsvipul: Prevented statements like these, with trailing spaces inside console.log statements- 
```
console.log("Result : ");

File: package.json [hunk 3]
```diff
@@ -193,9 +193,11 @@
...
           "node/no-deprecated-api": 0,
           "@typescript-eslint/member-ordering": "never",
-          "@typescript-eslint/restrict-plus-operands": "never"
+          "@typescript-eslint/restrict-plus-operands": "never",
+          "import/default": 0
```
vsvipul: Prevents a default module export being required by every module.

tsconfig.json
priyank-p: Hmm, I feel like we need this, otherwise we will be missing types like `Electron`. Let me take a look at this.

package.json
priyank-p: We will probably need to remove this later once we are done with migration

priyank-p:
Merged the first three commits here.

priyank-p:
@vsvipul I made a change to the `globalAny` thing. I extended the global interface to add the property we use. Merged Thanks!

PR #765, opened 2019-06-20
typescript: Migrate components dir to typescript.

vsvipul:
**What's this PR do?**
Migrate components dir to typescript.

Fixed all errors except one-
```
app/renderer/js/components/tab.ts:19:49 - error TS2339: Property 'template' does not exist on type 'Tab'.

19  	this.$el = this.generateNodeFromTemplate(this.template());
    	                                              ~~~~~~~~
```
There was no template function in Tab class.

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/components/tab.ts
+++ b/app/renderer/js/components/tab.ts
@@ -1,6 +1,7 @@
 'use strict';
 
-import BaseComponent = require('./base');
+import WebView from './webview';
+import BaseComponent from './base';
 
 // TODO: TypeScript - Type annotate props
 interface TabProps {
@@ -9,7 +10,7 @@ interface TabProps {
 
 class Tab extends BaseComponent {
 	props: TabProps;
-	webview: any;
+	webview: WebView;
 	$el: Element;
 	constructor(props: TabProps) {
 		super();
@@ -25,7 +26,7 @@ class Tab extends BaseComponent {
 	}
 
 	isLoading(): boolean {
-		return this.webview.isLoading;
+		return this.webview.$el.isLoading();
 	}
 
 	activate(): void {
--- a/app/renderer/js/components/webview.ts
+++ b/app/renderer/js/components/webview.ts
@@ -1,22 +1,35 @@
 'use strict';
 
-const path = require('path');
-const fs = require('fs');
+import path from 'path';
+import fs from 'fs';
+import { remote } from 'electron';
 
-const ConfigUtil = require(__dirname + '/../utils/config-util.js');
-const SystemUtil = require(__dirname + '/../utils/system-util.js');
-const { app, dialog } = require('electron').remote;
+import ConfigUtil = require('../utils/config-util');
+import SystemUtil = require('../utils/system-util');
+import BaseComponent = require('../components/base');
+import handleExternalLink = require('../components/handle-external-link');
 
-const BaseComponent = require(__dirname + '/../components/base.js');
-const handleExternalLink = require(__dirname + '/../components/handle-external-link.js');
+const { app, dialog } = remote;
 
 const shouldSilentWebview = ConfigUtil.getConfigItem('silent');
+
+// TODO: TypeScript - Type annotate WebViewProps.
+interface WebViewProps {
+	[key: string]: any;
+}
+
 class WebView extends BaseComponent {
-	constructor(props) {
+	props: any;
+	zoomFactor: number;
+	badgeCount: number;
+	loading: boolean;
+	customCSS: string;
+	$webviewsContainer: DOMTokenList;
+	$el: Electron.WebviewTag;
+	constructor(props: WebViewProps) {
 		super();
 
 		this.props = props;
-
 		this.zoomFactor = 1.0;
 		this.loading = true;
 		this.badgeCount = 0;
@@ -24,7 +37,7 @@ class WebView extends BaseComponent {
 		this.$webviewsContainer = document.querySelector('#webviews-container').classList;
 	}
 
-	template() {
+	template(): string {
 		return `<webview
 					class="disabled"
 					data-tab-id="${this.props.tabIndex}"
@@ -38,14 +51,14 @@ class WebView extends BaseComponent {
 				</webview>`;
 	}
 
-	init() {
-		this.$el = this.generateNodeFromTemplate(this.template());
-		this.props.$root.appendChild(this.$el);
+	init(): void {
+		this.$el = this.generateNodeFromTemplate(this.template()) as Electron.WebviewTag;
+		this.props.$root.append(this.$el);
 
 		this.registerListeners();
 	}
 
-	registerListeners() {
+	registerListeners(): void {
 		this.$el.addEventListener('new-window', event => {
 			handleExternalLink.call(this, event);
 		});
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -194,6 +194,7 @@
           "quotes": 0,
           "unicorn/catch-error-name": 0,
           "unicorn/no-console-spaces": 0,
+          "unicorn/regex-shorthand": 0,
```
vsvipul: Added to avoid linting error on line 141 of webview.ts.

priyank-p: @vsvipul we actually want this rule for improved readability of regex. The rule is also fixable so we can automate this.

app/renderer/js/components/handle-external-link.ts
```diff
@@ -1,12 +1,14 @@
...
+import LinkUtil = require('../utils/link-util');
+import DomainUtil = require('../utils/domain-util');
+import ConfigUtil = require('../utils/config-util');
+
+const { shell, app } = remote;
 
 const dingSound = new Audio('../resources/sounds/ding.ogg');
 
-function handleExternalLink(event) {
+function handleExternalLink(this: any, event: any): void {
```
priyank-p: I would add a todo comment here, to remind us to figure out a way to passed correct type here; we probably need to use a generic.

app/renderer/js/components/tab.ts
```diff
@@ -1,9 +1,12 @@
 'use strict';
 
-const BaseComponent = require(__dirname + '/../components/base.js');
+import BaseComponent = require('../components/base');
 
 class Tab extends BaseComponent {
-	constructor(props) {
+	props: any;
```
priyank-p: Should be `object`, and we should also add a todo for this.

app/renderer/js/components/tab.ts
```diff
@@ -1,9 +1,12 @@
...
+	webview: any;
```
priyank-p: We should also figure out type of this. Shouldn't be `any`.

priyank-p: Ah, so this should be `WebView` class.

app/renderer/js/components/server-tab.ts
```diff
@@ -17,24 +19,23 @@ class ServerTab extends Tab {
...
-		this.$badge = this.$el.getElementsByClassName('server-tab-badge')[0];
+		this.$badge = this.$el.querySelectorAll('.server-tab-badge')[0];
 	}
 
-	updateBadge(count) {
+	updateBadge(count: number): void {
 		if (count > 0) {
-			const formattedCount = count > 999 ? '1K+' : count;
-
+			const formattedCount: string = count > 999 ? '1K+' : count.toString();
```
priyank-p: We don't need to explicitly type annotate to string here.

app/renderer/js/components/functional-tab.ts
```diff
@@ -1,9 +1,10 @@
 'use strict';
 
-const Tab = require(__dirname + '/../components/tab.js');
+import Tab = require('../components/tab');
 
 class FunctionalTab extends Tab {
-	template() {
+	$closeButton: any;
```
priyank-p: This should be `Element` or `HTMLElement`?

app/renderer/js/components/functional-tab.ts
```diff
@@ -1,9 +1,10 @@
 'use strict';
```
priyank-p: Also you forgot to do `export = FunctionTab;` at the bottom.

app/renderer/js/components/webview.ts
```diff
@@ -1,30 +1,38 @@
 'use strict';
 
-const path = require('path');
-const fs = require('fs');
+import * as path from 'path';
+import * as fs from 'fs';
```
priyank-p: Remove the `* as` since we don't need it there.

app/renderer/js/components/webview.ts
```diff
@@ -1,30 +1,38 @@
...
 
-const BaseComponent = require(__dirname + '/../components/base.js');
-const handleExternalLink = require(__dirname + '/../components/handle-external-link.js');
+const { app, dialog } = remote;
 
 const shouldSilentWebview = ConfigUtil.getConfigItem('silent');
+
 class WebView extends BaseComponent {
-	constructor(props) {
+	props: any;
```
priyank-p, line 22: should be `object` instead of `any`.

app/renderer/js/components/webview.ts
```diff
@@ -1,30 +1,38 @@
...
+	zoomFactor: number;
+	badgeCount: number;
+	loading: boolean;
+	customCSS: string;
+	$webviewsContainer: any ;
```
priyank-p: Hmm, shouldn't this be `Element` or `HTMLElement`?

package.json
vsvipul: @priyank-p Great. I'll work on fixing this in webview then.

priyank-p (commented):
@vsvipul left bunch of comments can you address those? About the error you are facing, I think that should be `this.super.template()`.

app/renderer/js/components/tab.ts
priyank-p: Add a todo comment here to remind use that we need to annotate this to `WebView`.

app/renderer/js/components/webview.ts
vsvipul: Should be ```DOMTokenList```

package.json
vsvipul: Fixed this in the last commit. Please check if if is correct. Worked on regex for the first time.

app/renderer/js/components/tab.ts
vsvipul: @priyank-p I have tried using WebView here. Please check and tell if it is the correct way or not.

app/renderer/js/components/tab.ts
```diff
@@ -29,7 +31,7 @@ class Tab extends BaseComponent {
 	}
 
 	isLoading(): boolean {
-		return this.webview.isLoading;
+		return this.webview.$el.isLoading();
```
priyank-p, line 29: Can you explain why this change is correct? Is this a bug from before?

priyank-p:
@vsvipul merged everything except the one commit for which I left a question above.

app/renderer/js/components/tab.ts
vsvipul: @priyank-p Yes... ```isLoading``` is a property of ```<webview>``` type but here ```this.webview``` was of type ```WebView``` class which we created. ```$el``` is of type ```<webview>``` . If this is correct, then yes, this should be a bug from before.. @abhigyank and @akashnimare Can you please confirm this?

abhigyank: From documentation looks correct. Does this log correct value when called?

vsvipul: When i checked this, i coudn't exactly test out its working, but on console.logging previous code return ```undefined```, while this returns ```boolean``` values.

priyank-p:
@vsvipul Merged thanks! Next you can either work on `preference` directory or the `resources/messages.js` which should be really quick. I am working on `notfication` directory and I should be done soon.

These were the [changes](https://github.com/zulip/zulip-desktop/compare/adfab3db946cf254f29009b51fbacbb5470c4ee7..dae26b6b3cb493148f38cb069bd0a12215fd8aca) is made for the whole PR before merging. Just minor things.

(Also last couple of PR have been really great!)

PR #768, opened 2019-06-22
components: Remove unused isLoading function from Tab.

priyank-p:
Before we were returning this.webview.isLoading, which is always undefined
since we don't define any isLoading property on class WebView. This was
caught during migration to typescript.


akashnimare:
@priyank-p looks like we are not using this function anymore, so we should remove this?

priyank-p:
Yeah, remove the function entirely; should have a git grep before making a PR.

akashnimare:
Cool, thanks. Merged.

PR #769, opened 2019-06-22
typescript: Migrate notifications directory to typescript.

priyank-p:


app/renderer/js/notification/helpers.ts
```diff
@@ -1,30 +1,31 @@
...
+import Logger from '../utils/logger-util';
 
 const logger = new Logger({
 	file: 'errors.log',
 	timestamp: true
 });
 
 // Do not change this
-const appId = 'org.zulip.zulip-electron';
+export const appId = 'org.zulip.zulip-electron';
```
vsvipul, line 10: Why do we need to export this ?

app/renderer/js/notification/darwin-notifications.ts
```diff
@@ -39,45 +44,45 @@ class DarwinNotification {
...
 	}
 
-	static requestPermission() {
+	static requestPermission(): void {
 		return; // eslint-disable-line no-useless-return
 	}
 
 	// Override default Notification permission
-	static get permission() {
+	static get permission(): 'granted' | 'denied' {
```
vsvipul: We can create an ENUM type here?

app/renderer/js/notification/index.ts
```diff
@@ -1,12 +1,12 @@
...
 
-const {
-  remote: { app }
-} = require('electron');
+import { remote } from 'electron';
 
-const params = require('../utils/params-util.js');
-const DefaultNotification = require('./default-notification');
-const { appId, loadBots } = require('./helpers');
+import * as params from '../utils/params-util.js';
```
vsvipul: .js can be remove here.

typings.d.ts
```diff
@@ -4,13 +4,14 @@
 // are not supported
 declare module '*';
 
-declare var page_params: object;
+declare var page_params: any;
```
vsvipul, line 7: Why was this change required?

typings.d.ts
```diff
@@ -9,9 +9,27 @@ declare var page_params: any;
 // since requestIdleCallback didn't make it into lib.dom.d.ts yet
 declare function requestIdleCallback(callback: Function, options?: object): void;
 
+// Patch Notification object so we can implement our side
+// of Notification classes which we export into zulip side through
+// preload.js; if we don't do his extending Notification will throw error.
+// Relevant code is in app/renderer/js/notification/default-notification.ts
+// and the relevant function is requestPermission.
+declare var PatchedNotification: {
```
vsvipul: Why is this declared globally? Can't it be done inside the file where used?

app/renderer/js/notification/helpers.ts
priyank-p: We use it in `default-notification`.

app/renderer/js/notification/darwin-notifications.ts
priyank-p: I am pretty sure I put `NotificationPermission` here 😕. Also, I don't think enum is a good idea here.

typings.d.ts
priyank-p: Because you can't access `page_parmas` property if the type is `object`; that's why I do an interface with `[key: string]: any;` for most of code and add todo so we can come back to it later instead of putting any. In this case though we can't and don't need to annotate `page_params` so any is fine.

app/renderer/js/notification/darwin-notifications.ts
```diff
@@ -39,45 +44,45 @@ class DarwinNotification {
...
 		}
 
 		if (event === 'reply') {
-			replyHandler = handler;
+			replyHandler = handler as ReplyHandler;
 		}
 	}
 
-	notificationHandler({ response }) {
+	notificationHandler({ response }: { response: string }): void {
```
vsvipul: Can be separated as a type outside, but this seems fine too.

app/renderer/js/notification/helpers.ts
```diff
@@ -139,18 +145,8 @@ function parseReply(reply) {
...
 }
 
-function setupReply(id) {
+export function setupReply(id: string): void {
 	const { narrow } = window;
 	const narrowByTopic = narrow.by_topic || narrow.by_subject;
 	narrowByTopic(id, { trigger: 'notification' });
 }
-
-module.exports = {
```
vsvipul, line 148: Can't we follow the same thing we did earlier and do something like -
```
exports = { 1, 2 } 
```
here ?

app/renderer/js/notification/darwin-notifications.ts
priyank-p: Yeah, I doing that will make it more readable.

app/renderer/js/notification/helpers.ts
priyank-p: I prefer not to do that just for the sake of convention relating to this ES6 syntax.

PR #770, opened 2019-06-22
typescript: Migrate pages dir to typescript.

vsvipul:
  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS


app/renderer/js/pages/preference/add-certificate.ts
```diff
@@ -1,20 +1,28 @@
...
+const { dialog } = remote;
 
 class AddCertificate extends BaseComponent {
-	constructor(props) {
+	// TODO: TypeScript - Here props should be object type
+	props: any;
+	_certFile: string;
+	$addCertificate: Element | null;
+	addCertificateButton: Element | null;
+	serverUrl: Element | null;
```
priyank-p: Make this `HTMLInputElement` since we know it's a input element. That way we don't need to type cast this to `HTMLInputElement`.

app/renderer/js/pages/preference/add-certificate.ts
```diff
@@ -52,35 +60,35 @@ class AddCertificate extends BaseComponent {
...
-	addHandler() {
+	addHandler(): void {
 		const showDialogOptions = {
 			title: 'Select file',
 			defaultId: 1,
 			properties: ['openFile'],
 			filters: [{ name: 'crt, pem', extensions: ['crt', 'pem'] }]
 		};
-		dialog.showOpenDialog(showDialogOptions, selectedFile => {
+		dialog.showOpenDialog(showDialogOptions as OpenDialogOptions, selectedFile => {
```
priyank-p: It's better to do `const showDialogOptions: OpenDialogOptions` above where the variable is declared. Also typescript should automatically be type-checking for `showDialogOptions` anyway so we probably don't need to do this?

app/renderer/js/pages/preference/add-certificate.ts
```diff
@@ -52,35 +60,35 @@ class AddCertificate extends BaseComponent {
...
 
-	initListeners() {
+	initListeners(): void {
 		this.addCertificateButton.addEventListener('click', () => {
 			this.addHandler();
 		});
 
 		this.serverUrl.addEventListener('keypress', event => {
-			const EnterkeyCode = event.keyCode;
+			const EnterkeyCode = (event as KeyboardEvent).keyCode;
```
priyank-p: You should do `(event: KeyboardEvent)` above; `as` is generally used for type assertion.

app/renderer/js/pages/preference/badge-settings.ts
```diff
@@ -1,10 +1,10 @@
 'use strict';
-const electron = require('electron');
-const { app } = require('electron');
+import { app } from 'electron';
 
-const ConfigUtil = require(__dirname + '/../../utils/config-util.js');
+import electron = require('electron');
```
priyank-p, line 4: Minor suggestion here, you can also do `import electron, { app } from 'electron';` I am pretty sure.

app/renderer/js/pages/preference/connected-org-section.ts
```diff
@@ -1,18 +1,27 @@
...
+import AddCertificate = require('./add-certificate');
+import FindAccounts = require('./find-accounts');
 
 class ConnectedOrgSection extends BaseSection {
-	constructor(props) {
+	// TODO: TypeScript - Here props should be object type
+	props: any;
+	$serverInfoContainer: Element | null;
+	$existingServers: Element | null;
+	$newOrgButton: Element | null;
```
priyank-p: Since we know this a button make this a `HTMLButtonElement` so later if we needed to do say `.click` somebody doesn't need to fiddle with this.

app/renderer/js/pages/preference/find-accounts.ts
```diff
@@ -24,15 +29,15 @@ class FindAccounts extends BaseComponent {
...
 	}
 
-	init() {
+	init(): void {
 		this.$findAccounts = this.generateNodeFromTemplate(this.template());
-		this.props.$root.appendChild(this.$findAccounts);
+		this.props.$root.append(this.$findAccounts);
 		this.$findAccountsButton = this.$findAccounts.querySelector('#find-accounts-button');
-		this.$serverUrlField = this.$findAccounts.querySelectorAll('input.setting-input-value')[0];
+		this.$serverUrlField = this.$findAccounts.querySelectorAll('input.setting-input-value')[0] as HTMLInputElement;
```
priyank-p, line 36: Do we really need to do `as HTMLInputElement` here? I feel like this should be unnessacary since we declared above that this variable is in fact type of `HTMLInputElement`.

app/renderer/js/pages/preference/find-accounts.ts
```diff
@@ -42,14 +47,14 @@ class FindAccounts extends BaseComponent {
...
-	initListeners() {
+	initListeners(): void {
 		this.$findAccountsButton.addEventListener('click', () => {
 			this.findAccounts(this.$serverUrlField.value);
 		});
 
 		this.$serverUrlField.addEventListener('click', () => {
 			if (this.$serverUrlField.value === 'zulipchat.com') {
-				this.$serverUrlField.setSelectionRange(0, 0);
+				(this.$serverUrlField.setSelectionRange(0, 0));
```
priyank-p: This is syntax error, since there is a open paren here?

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -354,30 +359,30 @@ class GeneralSection extends BaseSection {
...
+	customCssDialog(): void {
 		const showDialogOptions = {
 			title: 'Select file',
 			defaultId: 1,
 			properties: ['openFile'],
 			filters: [{ name: 'CSS file', extensions: ['css'] }]
 		};
 
-		dialog.showOpenDialog(showDialogOptions, selectedFile => {
+		dialog.showOpenDialog(showDialogOptions as OpenDialogOptions, selectedFile => {
```
priyank-p: See my comment above for the similar thing in previous commit.

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -389,51 +394,51 @@ class GeneralSection extends BaseSection {
...
 		});
 	}
 
-	showCustomCSSPath() {
+	showCustomCSSPath(): void {
 		if (!ConfigUtil.getConfigItem('customCSS')) {
-			const cssPATH = document.getElementById('remove-custom-css');
-			cssPATH.style.display = 'none';
+			const cssPATH = document.querySelector('#remove-custom-css');
+			(cssPATH as HTMLElement).style.display = 'none';
```
priyank-p: Do `const cssPATH: HTMLElement = ...` in the previous line.

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -389,51 +394,51 @@ class GeneralSection extends BaseSection {
...
 		}
 	}
 
-	removeCustomCSS() {
-		const removeCSSButton = document.getElementById('css-delete-action');
+	removeCustomCSS(): void {
+		const removeCSSButton = document.querySelector('#css-delete-action');
 		removeCSSButton.addEventListener('click', () => {
-			ConfigUtil.setConfigItem('customCSS');
+			ConfigUtil.setConfigItem('customCSS', undefined);
```
priyank-p: Let's make this `""`, a blank string, since we later pull out this value an inject it into the style tag anyways. cc @akashnimare

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -389,51 +394,51 @@ class GeneralSection extends BaseSection {
...
-	downloadFolderDialog() {
+	downloadFolderDialog(): void {
 		const showDialogOptions = {
 			title: 'Select Download Location',
 			defaultId: 1,
 			properties: ['openDirectory']
 		};
 
-		dialog.showOpenDialog(showDialogOptions, selectedFolder => {
+		dialog.showOpenDialog(showDialogOptions as OpenDialogOptions, selectedFolder => {
```
priyank-p: Ditto.

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -389,51 +394,51 @@ class GeneralSection extends BaseSection {
...
 			if (selectedFolder) {
 				ConfigUtil.setConfigItem('downloadsPath', selectedFolder[0]);
 				const downloadFolderPath = document.querySelector('.download-folder-path');
-				downloadFolderPath.innerText = selectedFolder[0];
+				(downloadFolderPath as HTMLElement).innerText = selectedFolder[0];
```
priyank-p: Ditto.

app/renderer/js/pages/preference/new-server-form.ts
```diff
@@ -36,21 +42,20 @@ class NewServerForm extends BaseComponent {
...
+	initForm(): void {
 		this.$newServerForm = this.generateNodeFromTemplate(this.template());
-		this.$saveServerButton = this.$newServerForm.getElementsByClassName('server-save-action')[0];
+		this.$saveServerButton = this.$newServerForm.querySelectorAll('.server-save-action')[0];
 		this.props.$root.innerHTML = '';
-		this.props.$root.appendChild(this.$newServerForm);
-
-		this.$newServerUrl = this.$newServerForm.querySelectorAll('input.setting-input-value')[0];
+		this.props.$root.append(this.$newServerForm);
+		this.$newServerUrl = this.$newServerForm.querySelectorAll('input.setting-input-value')[0] as HTMLInputElement;
```
priyank-p, line 55: See my comment about this in previous commit.

app/renderer/js/pages/preference/preference.ts
```diff
@@ -1,24 +1,28 @@
...
+import NetworkSection = require('./js/pages/preference/network-section');
+import ConnectedOrgSection = require('./js/pages/preference/connected-org-section');
+import ShortcutsSection = require('./js/pages/preference/shortcuts-section');
 
 class PreferenceView extends BaseComponent {
+	$sidebarContainer: Element;
+	$settingsContainer: Element;
+	nav: typeof Nav;
+	section: typeof ServersSection | typeof GeneralSection |
+		typeof NetworkSection | typeof ConnectedOrgSection | typeof ShortcutsSection;
```
priyank-p: Let extract all of this out to a variable, i.e. `type Section = typeof ...` and use it there.

app/renderer/js/pages/preference/preference.ts
```diff
@@ -77,32 +81,32 @@ class PreferenceView extends BaseComponent {
 	}
 
 	// Handle toggling and reflect changes in preference page
-	handleToggle(elementName, state) {
+	handleToggle(elementName: string, state: boolean): void {
 		const inputSelector = `#${elementName} .action .switch input`;
 		const input = document.querySelector(inputSelector);
 		if (input) {
-			input.checked = state;
+			(input as HTMLInputElement).checked = state;
```
priyank-p: Ditto.

app/renderer/js/pages/preference/preference.ts
```diff
@@ -77,32 +81,32 @@ class PreferenceView extends BaseComponent {
...
 		}
 	}
 
-	registerIpcs() {
-		ipcRenderer.on('switch-settings-nav', (event, navItem) => {
+	registerIpcs(): void {
+		ipcRenderer.on('switch-settings-nav', (_event: Event, navItem: string) => {
```
priyank-p, line 94: I have been meaning to turn off this rule that make use `_` before unused parameter. Can you look into this?

app/renderer/js/pages/preference/shortcuts-section.ts
```diff
@@ -0,0 +1,341 @@
...
+
+class ShortcutsSection extends BaseSection {
+	// TODO: TypeScript - Here props should be object type
+	props: any;
+	constructor(props: any) {
+		super();
+		this.props = props;
+	}
+
+	templateMac(): string {
```
priyank-p, line 18: Can you add a todo here briefly explaining that we need to de-duplicate the `templateMac()` and `templateWinLin()` here. Both of them have same template literal but and the only thing diffrent about them is the `userOSKey` which we could pass as argument.

priyank-p (commented):
@vsvipul posted a bunch of comments. Can you address those?

app/renderer/js/pages/preference/add-certificate.ts
vsvipul: tsc was giving an error on this. So, added this then. I'll try ```const showDialogOptions: OpenDialogOptions``` .

vsvipul:
```
'defaultId' does not exist in type 'OpenDialogOptions'.
```
This was the problem here. So doing it as i did it was necessary.

app/renderer/js/pages/preference/add-certificate.ts
vsvipul: Wasn't required. Removed it.

app/renderer/js/pages/preference/general-section.ts
vsvipul: defaultId is not a property in OpenDialogOptions. Should i remove it?

app/renderer/js/pages/preference/new-server-form.ts
vsvipul: This is required because querySelector returns Element type.

app/renderer/js/pages/preference/preference.ts
vsvipul: We have done this while migrating other files as well. Will have to change it after everything is done later, if we want to do this.

app/renderer/js/pages/preference/shortcuts-section.ts
vsvipul: No, the template literal is not same. I just compared them. One is 157 lines, while other is 141. Still, added a comment.

vsvipul:
@priyank-p I have made some changes. Can you please do a review again ?

app/renderer/js/pages/preference/badge-settings.ts
vsvipul: Gave the powerMonitor error again after using this.

app/renderer/js/pages/preference/add-certificate.ts
priyank-p: TypeScript here is correct `defaultId` is only avalible for `showMessageBox` not`showOpenDialog` so we should remove it. Also can you submit this patch to master?

app/renderer/js/pages/preference/badge-settings.ts
priyank-p: Oh, ok!

app/renderer/js/pages/preference/general-section.ts
priyank-p: Yes, I mentioned this above for similar thing.

app/renderer/js/pages/preference/new-server-form.ts
priyank-p: OK, sounds good!

app/renderer/js/pages/preference/preference.ts
priyank-p: Actually, we can leave the already done as is with `_`, if we turn this lint rule off we just don't have to keep doing this.

app/renderer/js/pages/preference/shortcuts-section.ts
priyank-p: Hmm cool! Either way I suspect it majority of it is the same. Thanks for adding the comment.

app/renderer/js/pages/preference/add-certificate.ts
vsvipul: OK, great.! will do.

app/renderer/js/pages/preference/preference.ts
vsvipul: Ok, we can then turn this off in a later commit!

app/renderer/js/pages/preference/find-accounts.ts
vsvipul: Then, we get this-
```
Type 'Element' is missing the following properties from type 'HTMLInputElement': accept, align, alt, autocomplete, and 164 more.ts(2740)
```

priyank-p: OK, so we do need this.

app/renderer/js/pages/preference/connected-org-section.ts
```diff
@@ -57,26 +66,24 @@ class ConnectedOrgSection extends BaseSection {
 		}
 
 		this.$newOrgButton.addEventListener('click', () => {
-			// We don't need to import this since it's already imported in other files
-			// eslint-disable-next-line no-undef
 			ipcRenderer.send('forward-message', 'open-org-tab');
```
priyank-p, line 69: I wonder how this didn't throw an error before!

priyank-p:
@vsvipul I extend the todo comment a bit for the `shortcut-section.ts` and merged this along with other PRs. We just finished the migration of all the files to typescript.

Next thing, we should work on the gulp thing we talked about, then work on making `npm start` workflow more robuts and lastly the moving the compiled js files to `src`. Once again @vsvipul thanks for working on this!

PR #771, opened 2019-06-23
typescript: Migrate js/ and resources/ to typescript.

vsvipul:
  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/electron-bridge.ts
+++ b/app/renderer/js/electron-bridge.ts
@@ -1,17 +1,19 @@
-const events = require('events');
-const { ipcRenderer } = require('electron');
+import events from 'events';
+import { ipcRenderer } from 'electron';
+
+type ListenerType = ((...args: any[]) => void);
 
 // we have and will have some non camelcase stuff
 // while working with zulip so just turning the rule off
 // for the whole file.
-/* eslint-disable camelcase */
+/* eslint-disable @typescript-eslint/camelcase */
 class ElectronBridge extends events {
-	send_event(...args) {
-		this.emit(...args);
+	send_event(eventName: string | symbol, ...args: any[]): void {
+		this.emit(eventName, ...args);
 	}
 
-	on_event(...args) {
-		this.on(...args);
+	on_event(eventName: string, listener: ListenerType): void {
+		this.on(eventName, listener);
 	}
 }
 
@@ -37,4 +39,6 @@ electron_bridge.on('realm_icon_url', iconURL => {
 // functions zulip side will emit event using ElectronBrigde.send_event
 // which is alias of .emit and on this side we can handle the data by adding
 // a listener for the event.
-module.exports = electron_bridge;
+export = electron_bridge;
+
+/* eslint-enable @typescript-eslint/camelcase */
--- a/app/renderer/js/feedback.ts
+++ b/app/renderer/js/feedback.ts
@@ -1,7 +1,11 @@
-const { app } = require('electron').remote;
-const path = require('path');
-const fs = require('fs');
-const SendFeedback = require('@electron-elements/send-feedback');
+import { remote } from 'electron';
+import path from 'path';
+import fs from 'fs';
+import SendFeedback from '@electron-elements/send-feedback';
+
+const { app } = remote;
+
+type SendFeedbackType = typeof SendFeedback;
 
 // make the button color match zulip app's theme
 SendFeedback.customStyles = `
@@ -22,7 +26,7 @@ button {
 `;
 
 customElements.define('send-feedback', SendFeedback);
-const sendFeedback = document.querySelector('send-feedback');
+const sendFeedback: SendFeedbackType = document.querySelector('send-feedback');
 const feedbackHolder = sendFeedback.parentElement;
 
 // customize the fields of custom elements
@@ -38,9 +42,9 @@ sendFeedback.useReporter('emailReporter', {
 	email: 'akash@zulipchat.com'
 });
 
-feedbackHolder.addEventListener('click', e => {
-  // only remove the class if the grey out faded
-  // part is clicked and not the feedback element itself
+feedbackHolder.addEventListener('click', (e: Event) => {
+	// only remove the class if the grey out faded
+	// part is clicked and not the feedback element itself
 	if (e.target === e.currentTarget) {
 		feedbackHolder.classList.remove('show');
 	}
@@ -56,7 +60,7 @@ const dataDir = app.getPath('userData');
 const logsDir = path.join(dataDir, '/Logs');
 sendFeedback.logs.push(...fs.readdirSync(logsDir).map(file => path.join(logsDir, file)));
 
-module.exports = {
+export = {
 	feedbackHolder,
 	sendFeedback
 };
--- a/app/renderer/js/preload.ts
+++ b/app/renderer/js/preload.ts
@@ -1,11 +1,18 @@
 'use strict';
 
-const { ipcRenderer, shell } = require('electron');
-const SetupSpellChecker = require('./spellchecker');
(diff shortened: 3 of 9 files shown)


app/renderer/js/electron-bridge.ts
```diff
@@ -1,17 +1,17 @@
...
 class ElectronBridge extends events {
-	send_event(...args) {
-		this.emit(...args);
+	send_event(eventName: string | symbol, ...args: any[]): void {
+		this.emit(eventName, ...args);
 	}
 
-	on_event(...args) {
-		this.on(...args);
+	on_event(eventName: string, listener: ((...args: any[]) => void)): void {
```
priyank-p: Minor suggestion here, let move the type into a variable for listener and use it here, it will look more cleaner that way.

app/renderer/js/feedback.ts
```diff
@@ -22,7 +24,7 @@ button {
 `;
 
 customElements.define('send-feedback', SendFeedback);
-const sendFeedback = document.querySelector('send-feedback');
+const sendFeedback: any = document.querySelector('send-feedback');
```
priyank-p: This should be `HTMLElement` or equivalent

typings.d.ts
```diff
@@ -4,7 +4,18 @@
 // are not supported
 declare module '*';
 
-declare var page_params: object;
+interface PageParamsObject extends object{
```
priyank-p: `extends object` is useless here?

typings.d.ts
```diff
@@ -4,7 +4,18 @@
...
 
 // since requestIdleCallback didn't make it into lib.dom.d.ts yet
 declare function requestIdleCallback(callback: Function, options?: object): void;
+
+interface MyWindow extends Window {
+    electron_bridge: any;
+    tray: any;
+    $: any;
+    lightbox: any;
+}
```
priyank-p: Missing eol here.

typings.d.ts
```diff
@@ -4,7 +4,18 @@
...
+interface PageParamsObject extends object{
+    realm_uri: string;
+    default_language: string;
+}
+declare var page_params: PageParamsObject;
 
 // since requestIdleCallback didn't make it into lib.dom.d.ts yet
 declare function requestIdleCallback(callback: Function, options?: object): void;
+
+interface MyWindow extends Window {
```
priyank-p: Let's call this `ZulipWebWindow` or something that makes it clear it's zulip web app window.

app/renderer/js/preload.ts
```diff
@@ -1,51 +1,55 @@
...
+
+// eslint-disable-next-line @typescript-eslint/camelcase
 window.electron_bridge = require('./electron-bridge');
 
-const logout = () => {
+const logout = (): void => {
 	// Create the menu for the below
-	document.querySelector('.dropdown-toggle').click();
+	const dropdown = document.querySelector('.dropdown-toggle');
+	(dropdown as HTMLElement).click();
```
priyank-p: Sorry it this is being brought of a lot but let's move the `HTMLElement` declaration to variable or do `as HTMLElement` if needed for others. Reason being of we need to use `dropdown` variable again in future someone will need ot do `(dropdown as HTMLElement)` over and over again which we should avoid.

app/renderer/js/preload.ts
```diff
@@ -1,51 +1,55 @@
...
 
 	const nodes = document.querySelectorAll('.dropdown-menu li:last-child a');
-	nodes[nodes.length - 1].click();
+	(nodes[nodes.length - 1] as HTMLElement).click();
```
priyank-p: Minor suggestion, you can use do `HTMLAnchorElement` here. Also I think we can also do `NodeList<HTMLAnchorArray>[]` or similar on `nodes variable` if you can figure this out.

app/renderer/js/preload.ts
```diff
@@ -1,51 +1,55 @@
...
 	// Additional check
-	if (node.text.trim().toLowerCase() === 'keyboard shortcuts (?)') {
-		node.click();
+	if (node.textContent.trim().toLowerCase() === 'keyboard shortcuts (?)') {
+		(node as HTMLElement).click();
 	} else {
 		// Atleast click the dropdown
-		document.querySelector('.dropdown-toggle').click();
+		const dropdown = document.querySelector('.dropdown-toggle');
+		(dropdown as HTMLElement).click();
```
priyank-p: Ditto.

app/renderer/js/preload.ts
```diff
@@ -1,51 +1,55 @@
...
 
-const shortcut = () => {
+const shortcut = (): void => {
 	// Create the menu for the below
 	const node = document.querySelector('a[data-overlay-trigger=keyboard-shortcuts]');
 	// Additional check
-	if (node.text.trim().toLowerCase() === 'keyboard shortcuts (?)') {
-		node.click();
+	if (node.textContent.trim().toLowerCase() === 'keyboard shortcuts (?)') {
+		(node as HTMLElement).click();
```
priyank-p: ditto

app/renderer/js/preload.ts
```diff
@@ -1,51 +1,55 @@
...
+		(dropdown as HTMLElement).click();
 	}
 };
 
-process.once('loaded', () => {
-	global.logout = logout;
-	global.shortcut = shortcut;
+process.once('loaded', (): void => {
+	(global as any).logout = logout;
+	(global as any).shortcut = shortcut;
```
priyank-p: For the `global` here, let's not type cast to any instead take the approach I took to `main.ts` for global: https://github.com/zulip/zulip-desktop/blob/typescript/app/main/index.ts#L19-L23

app/renderer/js/preload.ts
```diff
@@ -61,7 +65,7 @@ document.addEventListener('DOMContentLoaded', () => {
 		}
 		// Open image attachment link in the lightbox instead of opening in the default browser
 		const { $, lightbox } = window;
-		$('#main_div').on('click', '.message_content p a', function (e) {
+		$('#main_div').on('click', '.message_content p a', function (this: any, e: Event) {
```
priyank-p: For this `HTMLElement` should also be good.

app/renderer/js/tray.ts
```diff
@@ -206,7 +204,7 @@ function toggleTray() {
 	}
 	const selector = 'webview:not([class*=disabled])';
 	const webview = document.querySelector(selector);
-	const webContents = webview.getWebContents();
+	const webContents = (webview as WebviewTag).getWebContents();
```
priyank-p: Move the `as WebviewTag` to the variable declared above :)

priyank-p (approved):
@vsvipul This PR looks good though left minor comment that should be quick for you to do.

app/renderer/js/preload.ts
```diff
@@ -1,51 +1,55 @@
...
+	(global as any).logout = logout;
+	(global as any).shortcut = shortcut;
 });
 
 // To prevent failing this script on linux we need to load it after the document loaded
-document.addEventListener('DOMContentLoaded', () => {
+document.addEventListener('DOMContentLoaded', (): void => {
 	if (params.isPageParams()) {
 	// Get the default language of the server
-		const serverLanguage = page_params.default_language; // eslint-disable-line no-undef, camelcase
```
priyank-p, line 48: This change could be the reason you are getting linting error in CI!

vsvipul:
@priyank-p I have made all the suggested changes and forced push. Also, you were right in case of the linting error. It is fixed now. Thanks for the review.

priyank-p:
This is merged in typescript branch, I guess this didn't automatically merge because it merge conflicted with other thing. Thank @vsvipul!

PR #773, opened 2019-06-25
gulp: Add tasks for compile and run from src.

vsvipul:
**What's this PR do?**

**Screenshots?**

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS


gulpfile.js
```diff
@@ -5,6 +5,27 @@ const electron = require('electron-connect').server.create({
...
+const ts = require("gulp-typescript");
+const tsProject = ts.createProject("tsconfig.json");
+const paths = {
+    files: [
+        'app/**/*',
+        '!app/**/*.ts'
+    ]
+};
+
+gulp.task("copy-non-ts", function () {
```
priyank-p: We don't need to do this here, since we plan to move the regular `tsc` to output to `src`.

priyank-p: We can manually remove already outputted files by doing `git clean -xf`.

priyank-p (approved):
@vsvipul this LGTM! Haven't tried out locally yet though.

gulpfile.js
vsvipul: But, doesn't ```tsc``` just compile .ts to .js . This is for other files like css and html and images.

priyank-p: Ah, ok makes sense!

package.json
```diff
@@ -31,7 +31,8 @@
     "lint-js": "xo",
     "test": "npm run lint-html && npm run lint-css && npm run lint-js",
     "test-e2e": "gulp test-e2e",
-    "dev": "gulp dev & nodemon --watch app/main --watch app/renderer --exec 'npm test' -e html,css,js",
+    "compile": "gulp compile",
+    "dev": "gulp dev & nodemon --watch app/main --watch app/renderer --exec 'npm test' ",
```
akashnimare: A gentle reminder that `dev` script was broken when I last checked.

vsvipul:
@priyank-p Made some changes. Review please.

package.json
vsvipul: Fixed this. I think the nodemon part was not required. The processes are automatically reloaded from the gulp scripts in gulpfile.js .

priyank-p:
@vsvipul I tweaked this to make sure that we run `tsc` if there are files left to compiles otherwise the app throws an error. Thanks, @vsvipul!

PR #774, opened 2019-06-25
typescript: Migrate main.js to typescript.

vsvipul:
---

**What's this PR do?**

**Screenshots?**

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -1,53 +1,124 @@
 'use strict';
 
-const { ipcRenderer, remote, clipboard, shell } = require('electron');
-const isDev = require('electron-is-dev');
+import { ipcRenderer, remote, clipboard, shell } from 'electron';
+import isDev from 'electron-is-dev';
+import escape from 'escape-html';
+import electronConnect from 'electron-connect';
+import feedbackHolder from './js/feedback';
 
 const { session, app, Menu, dialog } = remote;
-const escape = require('escape-html');
 
-require(__dirname + '/js/tray.js');
-const DomainUtil = require(__dirname + '/js/utils/domain-util.js');
-const WebView = require(__dirname + '/js/components/webview.js');
-const ServerTab = require(__dirname + '/js/components/server-tab.js');
-const FunctionalTab = require(__dirname + '/js/components/functional-tab.js');
-const ConfigUtil = require(__dirname + '/js/utils/config-util.js');
-const DNDUtil = require(__dirname + '/js/utils/dnd-util.js');
-const ReconnectUtil = require(__dirname + '/js/utils/reconnect-util.js');
-const Logger = require(__dirname + '/js/utils/logger-util.js');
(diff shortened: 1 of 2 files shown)


app/renderer/js/main.ts
```diff
@@ -1,53 +1,119 @@
 'use strict';
 
-const { ipcRenderer, remote, clipboard, shell } = require('electron');
-const isDev = require('electron-is-dev');
+import { ipcRenderer, remote, clipboard, shell } from 'electron';
+import isDev from 'electron-is-dev';
+import escape from 'escape-html';
+import electronConnect from 'electron-connect';
```
priyank-p, line 6: This change is wrong; `electron-connect` is not installed when the app is running in production.

app/renderer/js/main.ts
```diff
@@ -1,53 +1,119 @@
...
 	timestamp: true
 });
 
+interface FunctionalTabProps {
+	name: string;
+	materialIcon: string;
+	url: string;
+}
+
+interface StringIndexType {
```
priyank-p: I'd name this `AnyObject` or something similar.

app/renderer/js/main.ts
```diff
@@ -1,53 +1,119 @@
...
+	[key: string]: any;
+}
+
+interface SettingsOptions {
+	[key: string]: any;
```
priyank-p: I remove this `[key: string]: any`. The reason you used it was because of the `for` loop which I solved by doing:
```typescript
for (const i in settingOptions) {
    const setting = i as keyof SettingsOptions;
    if (ConfigUtil.getConfigItem(i) === null) {
        ConfigUtil.setConfigItem(i, settingOptions[setting]);
    }
}
```

priyank-p:
@vsvipul I am seeing error in the console when I hover over sidebar when testing this PR can you debug?

vsvipul:
@priyank-p Somehow, using ```querySelectorAll()``` on line 109 instead of ```getElementsByClassName()``` seems to be causing this problem. Coudn't find any solution for this. Should we remove the xo rule ?

app/renderer/js/main.ts
vsvipul: What should be a solution for this? As xo does not allow imports in the middle of the program.

priyank-p: I used `require` instead.

priyank-p:
OK, this is a ugly bug in current setup which is hard to fix. The problem is by implementation detail both `document.querySelector` and `document.getElementsByClassName` differ, for better or worse I think the `querySelector` here is correct. So the issue here is we query for elements in the `constructor` but the problem is the `.server-tooltip` doesn't show up in the dom until the `init` method is called but with previous code we still get the dom elements somehow. I am going to add a `TODO` about this and come back to it later.

priyank-p:
@vsvipul merged thanks! We just finished the migration to typescript which is exciting. Let's start testing out the app to make sure we didn't introduce any regressions.

PR #776, opened 2019-06-27
typescript: Fix bunch of things.

priyank-p:
Removed, `declared module *` and the added fixes required.

@vsvipul can you take a quick look? This also fixes on the bug you encountered while testing the installer.

### DIFF
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -4,7 +4,6 @@ import { ipcRenderer, remote, clipboard, shell } from 'electron';
 import path from 'path';
 import isDev from 'electron-is-dev';
 import escape from 'escape-html';
-import electronConnect from 'electron-connect';
 import { feedbackHolder } from './feedback';
 
 const { session, app, Menu, dialog } = remote;
@@ -902,7 +901,7 @@ window.addEventListener('load', () => {
 	// `--no-electron-connect`
 	const mainProcessArgv = remote.getGlobal('process').argv;
 	if (isDev && !mainProcessArgv.includes('--no-electron-connect')) {
-		electronConnect.client.create();
+		require('electron-connect').client.create();
 	}
 });
 
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -1,25 +1,26 @@
 'use strict';
 
 import { ipcRenderer, remote, clipboard, shell } from 'electron';
+import path from 'path';
 import isDev from 'electron-is-dev';
 import escape from 'escape-html';
 import electronConnect from 'electron-connect';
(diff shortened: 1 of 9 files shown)


app/renderer/js/main.ts
```diff
@@ -4,22 +4,22 @@ import { ipcRenderer, remote, clipboard, shell } from 'electron';
...
 import electronConnect from 'electron-connect';
-import feedbackHolder from './js/feedback';
+import { feedbackHolder } from './feedback';
 
 const { session, app, Menu, dialog } = remote;
 
 // eslint-disable-next-line import/no-unassigned-import
-require('./js/tray');
-
-import DomainUtil = require('./js/utils/domain-util');
```
vsvipul, line 14: Why this used to work earlier? Was there a relative addressing at place here, this process used to run from index.html, that's why ?

app/renderer/js/main.ts
```diff
@@ -66,6 +66,8 @@ interface SettingsOptions {
 	loading?: AnyObject;
 }
 
+type ServerOrFunctionalTab = ServerTab | FunctionalTab;
```
vsvipul, line 70: When i used something like this, i got an error saying ```ServerTab is a class, can't be used as a type```

app/renderer/main.html
```diff
@@ -55,6 +55,11 @@
...
     </body>
-    <script src="js/main.js"></script>
+    
+    <script>
+        // we don't use src='./js/main' in the script tag because
+        // it messes up require module path resolution
+        require('./js/main');
+    </script>
     <script>require('./js/shared/preventdrag.js')</script>
 </html>
```
vsvipul, line 65: an eol is missed here.

vsvipul:
@priyank-p I am still facing this error on running the AppImage from the installers created-
Screenshot from 2019-06-27 19-46-41

In dev environment, everything is working fine. This is only for the AppImage.

app/renderer/js/main.ts
priyank-p: Because we were using `<script src="js/main.js">` and so the current way of importing didn't work out well for typescript so I change it to `<script>require('./js/main.js')</script>` which makes both world happy :)

app/renderer/js/main.ts
priyank-p: Yeah that was because typescript didn't know about the file; I though it was because we used `declare module *`

vsvipul: Oh, great!

vsvipul (approved):
LGTM!

priyank-p:
This was already merged so closing. Thanks @vsvipul for the review.

PR #779, opened 2019-06-27
typescript: Fix incorrect import statements.

vsvipul:
  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/menu.ts
+++ b/app/main/menu.ts
@@ -1,14 +1,11 @@
 'use strict';
-import * as path from 'path';
-
 import { app, shell, BrowserWindow, Menu, dialog } from 'electron';
-
-import * as fs from 'fs-extra';
-import * as AdmZip from 'adm-zip';
-
-import * as DNDUtil from '../renderer/js/utils/dnd-util';
 import { appUpdater } from './autoupdater';
 
+import AdmZip = require('adm-zip');
+import fs = require('fs-extra');
+import path = require('path');
+import DNDUtil = require('../renderer/js/utils/dnd-util');
 import Logger = require('../renderer/js/utils/logger-util');
 import ConfigUtil = require('../renderer/js/utils/config-util');
 
--- a/app/main/autoupdater.ts
+++ b/app/main/autoupdater.ts
@@ -1,11 +1,10 @@
 'use strict';
 import { app, dialog, shell } from 'electron';
 import { autoUpdater } from 'electron-updater';
-import * as isDev from 'electron-is-dev';
-import log from 'electron-log';
-
 import { linuxUpdateNotification } from './linuxupdater';	// Required only in case of linux
 
+import log = require('electron-log');
+import isDev = require('electron-is-dev');
 import ConfigUtil = require('../renderer/js/utils/config-util');
 
 export function appUpdater(updateFromMenu = false): void {
--- a/app/main/index.ts
+++ b/app/main/index.ts
@@ -1,13 +1,12 @@
 'use strict';
-import path from 'path';
-import fs from 'fs';
-import windowStateKeeper from 'electron-window-state';
-import isDev from 'electron-is-dev';
-
 import { sentryInit } from '../renderer/js/utils/sentry-util';
 import { appUpdater } from './autoupdater';
 import { setAutoLaunch } from './startup';
 
+import windowStateKeeper = require('electron-window-state');
+import path = require('path');
+import fs = require('fs');
+import isDev = require('electron-is-dev');
 import electron = require('electron');
 const { app, ipcMain } = electron;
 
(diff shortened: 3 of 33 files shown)


priyank-p:
@vsvipul thanks for looking into this. I decided to do an audit of all the imports just to be sure we didn't break anything. The way I got all the imports was by `git grep 'import.*from.*\;' | cut -f2 -d ':' | sort | uniq -c` which narrow everything down pretty easily. This problem is only with using `import * as` or star imports as I warned earlier.

I would like one more set of eyes looking through this just to be sure so here are all the imports:
```javascript
git grep 'import.*from.*\;' | cut -f2 -d ':' | sort | uniq -c

  1 import * as AdmZip from 'adm-zip';
  1 import * as AutoLaunch from 'auto-launch';
  1 import * as DNDUtil from '../renderer/js/utils/dnd-util';
  5 import * as fs from 'fs';
  1 import * as fs from 'fs-extra';
  1 import * as isOnline from 'is-online';
  2 import * as os from 'os';
  1 import * as params from '../utils/params-util';
  3 import * as path from 'path';
  1 import * as url from 'url';
  1 import * as wurl from 'wurl';
  2 import { app } from 'electron';
  1 import { app, dialog, shell } from 'electron';
  1 import { app, Notification } from 'electron';
  1 import { app, shell, BrowserWindow, Menu, dialog } from 'electron';
  1 import { appId, loadBots } from './helpers';
  2 import { appUpdater } from './autoupdater';
  1 import { autoUpdater } from 'electron-updater';
  1 import { Console as NodeConsole } from 'console'; // eslint-disable-line node/prefer-global/console
  1 import { feedbackHolder } from './feedback';
  1 import { focusCurrentServer } from './helpers';
  1 import { init } from '@sentry/electron';
  2 import { initSetUp } from './default-util';
  9 import { ipcRenderer } from 'electron';
  1 import { ipcRenderer, remote } from 'electron';
  1 import { ipcRenderer, remote, clipboard, shell } from 'electron';
  1 import { ipcRenderer, remote, OpenDialogOptions } from 'electron';
  1 import { ipcRenderer, remote, WebviewTag, NativeImage } from 'electron';
  1 import { ipcRenderer, shell } from 'electron';
  1 import { linuxUpdateNotification } from './linuxupdater';       // Required only in case of linux
  6 import { remote } from 'electron';
  1 import { remote, ipcRenderer } from 'electron';
  1 import { remote, OpenDialogOptions } from 'electron';
  1 import { sentryInit } from '../renderer/js/utils/sentry-util';
  1 import { sentryInit, captureException } from './sentry-util';
  1 import { setAutoLaunch } from './startup';
  3 import { shell } from 'electron';
  1 import { SpellCheckHandler, ContextMenuListener, ContextMenuBuilder } from 'electron-spellchecker';
  1 import BaseComponent from './base';
  2 import ConfigUtil from '../utils/config-util';
  1 import ConfigUtil from './config-util';
  1 import DefaultNotification from './default-notification';
  2 import escape from 'escape-html';
  1 import events from 'events';
  5 import fs from 'fs';
  1 import fs from 'fs-extra';
  4 import JsonDB from 'node-json-db';
  1 import log from 'electron-log';
  1 import Logger from '../utils/logger-util';
  1 import MacNotifier from 'node-mac-notifier';
  9 import path from 'path';
  2 import request from 'request';
  1 import semver from 'semver';
  1 import SendFeedback from '@electron-elements/send-feedback';
  1 import SetupSpellChecker from './spellchecker';
  1 import url from 'url';
  1 import WebView from './webview';
  1 import windowStateKeeper from 'electron-window-state';
```

priyank-p:
Also, it seems like there are lots of imports but half of them are `electron` and our code imports so we don't need to check those. I just didn't remove them from the list :)

app/main/menu.ts
```diff
@@ -4,7 +4,7 @@ import * as path from 'path';
 import { app, shell, BrowserWindow, Menu, dialog } from 'electron';
 
 import * as fs from 'fs-extra';
-import * as AdmZip from 'adm-zip';
+import AdmZip from 'adm-zip';
```
vsvipul: @priyank-p Where is this documented that we can import it this way? I suggest a better way would be to do all imports with commonjs syntax as ```import x = require('x')``` .This won't cause any problems.

priyank-p: @vsvipul If you feel strong about using commonjs import syntax feel free to update this :). The `import AdmZip from 'adm-zip';` is a ES6 default import statement meaning if the `amd-zip` module does `export default` or `module.exports.default` we will be able to import it. The module do that as we know, but typescript adds `__importDefault` helper which takes care of this.

vsvipul: Ok, will change to commonjs syntax then.

vsvipul:
@priyank-p  I have made some changes. Can you test it out?

priyank-p:
Merged thanks @vsvipul! We didn't need to use the commonjs import syntax for `fs`, and `path` and other node.js modules but doesn't hurt to do it either.

PR #772, opened 2019-06-24
syntax: Minor bug fixes.

vsvipul:
These were found out during migration to typescript. 

**You have tested this PR on:**
  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

PR #782, opened 2019-06-30
Use TypeScript type checker for zulip-desktop codebase.

### LINKED ISSUE #518: Add Typescript as a type checker in the codebase

priyank-p:
Me and @vsvipul have done lot of testing to make sure we didn't break anything. There are no significant code changes introduced during the migration, if it was needed because of bug fixes we also sent out patches to be included in master; see #772, #768 for example. That said, we'll still want all of use to double-check that there are no issues by testing on packaged app.

Fixes: #518


app/main/index.ts
```diff
@@ -1,21 +1,25 @@
...
-const windowStateKeeper = require('electron-window-state');
-const isDev = require('electron-is-dev');
-const appMenu = require('./menu');
-const { appUpdater } = require('./autoupdater');
+import { sentryInit } from '../renderer/js/utils/sentry-util';
+import { appUpdater } from './autoupdater';
+import { setAutoLaunch } from './startup';
+
+import windowStateKeeper = require('electron-window-state');
+import path = require('path');
```
kanishk98, line 7: Is there a specific reason for using `require` or `from` (or the other way around?)
It would make things much simpler if we decided on using only one.

app/main/index.ts
```diff
@@ -341,21 +345,21 @@ app.on('ready', () => {
...
 
-	ipcMain.on('realm-icon-changed', (event, serverURL, iconURL) => {
+	ipcMain.on('realm-icon-changed', (_event: Electron.IpcMessageEvent, serverURL: string, iconURL: string) => {
 		page.send('update-realm-icon', serverURL, iconURL);
 	});
 
 	// Using event.sender.send instead of page.send here to
 	// make sure the value of errorReporting is sent only once on load.
-	ipcMain.on('error-reporting', event => {
+	ipcMain.on('error-reporting', (event: Electron.IpcMessageEvent) => {
```
kanishk98, line 358: Shouldn't this be `_event` as well?

app/renderer/js/components/functional-tab.ts
```diff
@@ -1,9 +1,10 @@
 'use strict';
 
-const Tab = require(__dirname + '/../components/tab.js');
+import Tab = require('./tab');
```
kanishk98, line 3: Nice work with the paths! Thanks for this.

kanishk98 (approved):
LGTM. Tested it, works well on my system. 
Left a couple of questions below.

package.json
```diff
@@ -124,13 +124,16 @@
     "InstantMessaging"
   ],
   "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "^1.10.2",
+    "@typescript-eslint/parser": "^1.10.2",
```
kanishk98: We'd want exact version numbers here, right?
Forgot to comment this the last time around.

app/main/index.ts
vsvipul: This was a xo rule which we removed later on so _ is not required for unused parameters now.

abhigyank:
The app compiles and runs smoothly for me in dev. I see a decrease in startup time during development for the app. I will do a bit more testing in dev to do a final check and then also do a check on the packaged app soon.

gulpfile.js
```diff
@@ -5,16 +5,37 @@ const electron = require('electron-connect').server.create({
...
+
+const glob = require('glob');
+const { execSync } = require('child_process');
+
+const baseFilePattern = 'app/+(main|renderer)/**/*';
+const globOptions = { cwd: __dirname };
+const jsFiles = glob.sync(baseFilePattern + '.js', globOptions);
+const tsFiles = glob.sync(baseFilePattern + '.ts', globOptions);
+if (jsFiles.length !== tsFiles.length) {
+	const npx = process.platform === 'win32' ? 'npx.cmd' : 'npx';
```
abhigyank, line 19: Is this not required `tools/run-dev`? There we seem to be directly using `npx`.

abhigyank (approved):
LGTM! Packaged version packaged Linux AppImage, works as expected. Didn't get any glitch or issue due to migrations,everything is smooth. Good job :rocket:

app/main/index.ts
priyank-p: @kanishk98 yes! So the problem is we cannot use the from syntax to import modules that either export a class or function (eventhough for node both are functions!). So instead of having a mix we decided to use the require syntax for node modules and for our own code we still use from.

gulpfile.js
priyank-p: We do use for `tools/run-dev` but it's not visible; the `run` function takes care of this.

priyank-p:
@abhigyank I wonder how much does the startup decrease? We didn't do anything special in that regard.

app/main/index.ts
kanishk98: Ah right, makes sense. Thanks!

akashnimare:
@priyank-p can you fix those merge conflicts so that I can pull and test it locally?

package.json
priyank-p: Done during rebase!

akashnimare:
Initial observation -
* `CMD + Q` does quit the app but not the editor. I have to use `CTRL + C` in order to quit the `npm start` command.

priyank-p:
I pushed a commit which kills the `npm start`, `tools/run-dev`, if either `tsc` or the app is killed.

abhigyank:
> @abhigyank I wonder how much does the startup decrease? We didn't do anything special in that regard.

Observable but not very significant. Not sure how I can measure that though. Even I don't see a reason for decrease, but I feel it is probably due to the way we were importing things and how the compiled js imports things.

On another note - how do we merge this PR? Do we merge all commits to preserve the incremental approach or we can club a few similar commits into one to make commits show incremental approach but fixes commits are in a single commit kind of thing?

vsvipul (approved):

priyank-p:
About how to merge this PR -- yes we want all the commit history to be intact, grouping changes is a good idea but that would be just unnecessary work we have to do.

app/main/autoupdater.ts
```diff
@@ -1,18 +1,19 @@
...
 
-function appUpdater(updateFromMenu = false) {
+export function appUpdater(updateFromMenu = false): void {
 	// Don't initiate auto-updates in development
 	if (isDev) {
 		return;
 	}
 
 	if (process.platform === 'linux' && !process.env.APPIMAGE) {
-		const { linuxUpdateNotification } = require('./linuxupdater');
```
akashnimare, line 15: @priyank-p, @vsvipul I put it here because I thought we'll only require/import this on Linux.

app/main/index.ts
```diff
@@ -1,21 +1,25 @@
...
-const electron = require('electron');
-const windowStateKeeper = require('electron-window-state');
-const isDev = require('electron-is-dev');
-const appMenu = require('./menu');
-const { appUpdater } = require('./autoupdater');
+import { sentryInit } from '../renderer/js/utils/sentry-util';
+import { appUpdater } from './autoupdater';
+import { setAutoLaunch } from './startup';
+
+import windowStateKeeper = require('electron-window-state');
```
akashnimare, line 6: I think I have read it somewhere that it's good practice to put the app dependencies first (path, fs, other modules).

priyank-p: Yes, it is generally better to organize the imports.

app/main/autoupdater.ts
priyank-p: OK, this change was made because linting usually complains of assigning value through require statement in typescript unless it's a side effect require i.e. `require('some-module')` not `const someModule = require('some-module')`. Since there is nothing in the `linux-updater.js` that causes error if the os is not linux I think we should be fine with this as-is?

akashnimare: Yeah, there are no side-effects. I was just worried about the performance since we are importing this on other systems also.

app/main/index.ts
priyank-p: We had a linting rule which is why this was changed.

app/package.json
```diff
@@ -37,7 +37,7 @@
     "electron-window-state": "5.0.3",
     "escape-html": "1.0.3",
     "is-online": "7.0.0",
-    "node-json-db": "0.9.1",
+    "node-json-db": "0.9.2",
```
akashnimare, line 40: @priyank-p, @vsvipul this is safe, right?

vsvipul: This only adds the typings required for typescript. So, it shouldn't be a problem. This was the only change as seen from the version history [here](https://github.com/Belphemur/node-json-db/releases).

app/renderer/js/components/tab.ts
```diff
@@ -1,44 +1,44 @@
...
+	$el: Element;
+	constructor(props: TabProps) {
 		super();
 
 		this.props = props;
 		this.webview = this.props.webview;
-
-		this.init();
-	}
-
```
akashnimare, line 14: @vsvipul, @priyank-p can you guys explain why we have removed this?

vsvipul: this.template() used in init() function was a property of its child classes - ServerTab and FunctionalTab . So this init() function was moved to both these subclasses.

app/renderer/js/pages/network.ts
```diff
@@ -1,20 +1,21 @@
...
-	init() {
+	init(): void {
 		this.$reconnectButton.addEventListener('click', () => {
 			ipcRenderer.send('forward-message', 'reload-viewer');
 		});
 	}
 }
 
-window.onload = () => {
+window.addEventListener('load', () => {
```
akashnimare, line 18: Is there any reason for changing this?

vsvipul: It is an xo linter rule to prefer eventListener instead of onload.

app/renderer/js/pages/preference/add-certificate.ts
```diff
@@ -1,20 +1,28 @@
 'use-strict';
 
-const { dialog } = require('electron').remote;
-const path = require('path');
+import { remote, OpenDialogOptions } from 'electron';
```
akashnimare, line 3: Not sure if electron has this api `OpenDialogOptions` ?

app/renderer/js/pages/preference/connected-org-section.ts
```diff
@@ -27,21 +36,21 @@ class ConnectedOrgSection extends BaseSection {
...
 
 		const servers = DomainUtil.getDomains();
 		this.props.$root.innerHTML = this.template();
 
-		this.$serverInfoContainer = document.getElementById('server-info-container');
-		this.$existingServers = document.getElementById('existing-servers');
-		this.$newOrgButton = document.getElementById('new-org-button');
-		this.$addCertificateContainer = document.getElementById('add-certificate-container');
-		this.$findAccountsContainer = document.getElementById('find-accounts-container');
+		this.$serverInfoContainer = document.querySelector('#server-info-container');
```
akashnimare, line 49: Can someone explain why we have moved to `querySelector` instead of `getElementById`?

vsvipul: This is also an xo rule to prefer querySelector and querySelectorAll instead of getElementById and getElementByClass.

app/renderer/js/pages/preference/find-accounts.ts
```diff
@@ -1,16 +1,21 @@
 'use-strict';
 
-const { shell } = require('electron');
+import { shell } from 'electron';
 
-const BaseComponent = require(__dirname + '/../../components/base.js');
+import BaseComponent = require('../../components/base');
```
akashnimare, line 5: This path will be automatically resolved on Windows, right?

app/renderer/js/pages/preference/add-certificate.ts
```diff
@@ -55,12 +63,12 @@ class AddCertificate extends BaseComponent {
...
 			dialog.showErrorBox('Error', `Please, ${serverUrl === '' ?
-      'Enter an Organization URL' : 'Choose certificate file'}`);
+				'Enter an Organization URL' : 'Choose certificate file'}`);
 		}
 	}
 
-	addHandler() {
-		const showDialogOptions = {
+	addHandler(): void {
+		const showDialogOptions: OpenDialogOptions = {
```
vsvipul, line 71: Using Electron.OpenDialogOptions here and removing the wrong import here should fix this. I think the import here was added automatically by the text editor. @priyank-p FYI.

app/renderer/js/pages/preference/add-certificate.ts
vsvipul: Made another comment in this file.

app/renderer/js/shared/preventdrag.ts
```diff
@@ -5,13 +5,13 @@
 // It doesn't affect the compose box so that users can still
 // use drag and drop event to share files etc
 
-const preventDragAndDrop = () => {
+function preventDragAndDrop(): void {
```
akashnimare: We should be consistent here i.e. with es6.

app/renderer/js/tray.ts
```diff
@@ -1,17 +1,15 @@
 'use strict';
-const path = require('path');
-
-const electron = require('electron');
-
-const { ipcRenderer, remote } = electron;
+import { ipcRenderer, remote, WebviewTag, NativeImage } from 'electron';
```
akashnimare, line 2: Can somebody explain `WebviewTag`, `NativeImage`?

vsvipul: @priyank-p Same here, as in the OpenDialogOptions. Even though, this is also a right way to do this. I think we'll have to prefer ```Electron.type``` method to avoid unnecesary imports.

app/renderer/js/utils/default-util.ts
```diff
@@ -1,6 +1,6 @@
-const fs = require('fs');
+import fs = require('fs');
 
-let app = null;
+let app: Electron.App;
```
akashnimare: How does this work? Two doubts here -
* Should it be `null`?
* We haven't imported electron here?

app/renderer/js/utils/domain-util.ts
```diff
@@ -168,25 +170,21 @@ class DomainUtil {
 
 			const certsError = error.toString().includes('certificate');
 			if (domain.indexOf(whitelistDomains) >= 0 || certsError) {
-				try {
-					return await this.checkCertError(domain, serverConf, error, silent);
```
akashnimare, line 172: I guess we don't need await here?

app/renderer/js/components/tab.ts
priyank-p: See commit message of this [commit](https://github.com/zulip/zulip-desktop/pull/782/commits/340023e498e8765981c02447b10a4f8bee359228) to see the relevant changes also.

app/renderer/js/utils/logger-util.ts
```diff
@@ -1,37 +1,57 @@
...
+	timestamp?: any;
+	file?: string;
+	level?: boolean;
+	logInDevMode?: boolean;
+}
 
 initSetUp();
 
-let app = null;
+let app: Electron.App = null;
```
akashnimare, line 24: This is what we should use in other places also.

app/renderer/js/pages/network.ts
priyank-p: Also, it's generally better practice to use `addEventListener`

app/renderer/js/pages/preference/add-certificate.ts
priyank-p: Yes it not a api, the `OpenDialogOptions` is a type that typescript can import.

app/renderer/js/pages/preference/find-accounts.ts
priyank-p: Yes, see https://github.com/zulip/zulip-desktop/pull/782/commits/8b7953343ded2e2a04218856ff42299283fe5bc7 for more details.

app/renderer/js/pages/preference/add-certificate.ts
priyank-p: I think it's fine to leave this as is since `Electron.OpenDialogOptions` is too verbose.

app/renderer/js/utils/default-util.ts
priyank-p: The `let app: Electron.App` tells typescript that it will be an electron app instance when initialized so this is correct. If assigned some other value to it later on it will complain.

app/renderer/js/utils/domain-util.ts
priyank-p: Yeah!

package.json
```diff
@@ -21,18 +21,21 @@
     "node": ">=6.0.0"
   },
   "scripts": {
-    "start": "electron app --disable-http-cache --no-electron-connect",
+    "start": "node tools/run-dev",
+    "start-app": "electron app --disable-http-cache --no-electron-connect",
```
akashnimare: We should remove this? Also, when I run `npm start` for the first time it exits the command (I guess it only compiles) and doesn't start the app.

typings.d.ts
```diff
@@ -0,0 +1,54 @@
+declare module 'adm-zip';
```
akashnimare, line 1: Can you guys explain the use case of this file i.e. `typings.d.ts`?

priyank-p: We define common types we are likely to use in many places here, like `page_params` etc., so like a global types file. Also, since there are not types of some modules we need to use this file also.

package.json
priyank-p: That the bug I fixed last time, can you rebase and try again to check again?
(Also you can do `git clean -xf` to remove compiled files.)

vsvipul: @priyank-p we can add this ```git clean -xf``` here as a script. I think that will be good.

priyank-p: Sure!

app/renderer/js/pages/preference/add-certificate.ts
akashnimare: Got it.

app/renderer/js/pages/preference/find-accounts.ts
akashnimare: Got it, good commit message btw.

app/renderer/js/utils/default-util.ts
akashnimare: We should use `let app: Electron.App = null;` here?

app/renderer/js/shared/preventdrag.ts
akashnimare: This can be done as a follow up I think.

akashnimare:
@priyank-p, @vsvipul I have gone through the code once and left a few comments. Let me know once you make the changes as discussed above so that I can review + merge it.

app/renderer/js/shared/preventdrag.ts
priyank-p: Done!

app/renderer/js/utils/default-util.ts
priyank-p: Done! I changed this one instance to `let app: Electron.App = null;`. Also, I realized I missed answering the second bullet, `Electron` comes from the electron typescript declaration file which means that it is automatically know by typescript without importing it.

package.json
priyank-p: Done! For the `git clean` script, I added that as `git clean app/*.js -xf` so it would only remove typescript compiled files and nothing else like `.vscode` etc.

app/main/index.ts
```diff
@@ -234,7 +238,7 @@ app.on('ready', () => {
 	});
 
 	ipcMain.on('clear-app-settings', () => {
-		global.mainWindowState.unmanage(mainWindow);
+		globalPatched.mainWindowState.unmanage();
```
akashnimare, line 241: We haven't passed the window here. Is this intentional?

vsvipul: This was wrongly declared before. We don't need to pass anything to unmanage here. Passing of mainWindow is only required in a manage call.

vsvipul: You can refer to [this](https://github.com/mawie81/electron-window-state/blob/master/index.js).

app/renderer/js/notification/index.ts
```diff
@@ -1,12 +1,11 @@
 'use strict';
 
-const {
-  remote: { app }
-} = require('electron');
+import { remote } from 'electron';
+import * as params from '../utils/params-util';
```
akashnimare, line 4: Is this correct?

vsvipul: Yes, it is. What's wrong with this ?

app/renderer/js/pages/preference/shortcuts-section.ts
```diff
@@ -0,0 +1,344 @@
...
+import BaseSection = require('./base-section');
+
+class ShortcutsSection extends BaseSection {
+	// TODO: TypeScript - Here props should be object type
+	props: any;
+	constructor(props: any) {
+		super();
+		this.props = props;
+	}
+	// TODO - Deduplicate templateMac and templateWinLin functions. In theory
```
akashnimare, line 14: Agree on this. We should fix this as a follow-up.

app/renderer/js/preload.ts
```diff
@@ -1,50 +1,61 @@
...
+const globalPatched = global as PatchedGlobal;
 
 // eslint-disable-next-line import/no-unassigned-import
 require('./notification');
 
 // Prevent drag and drop event in main process which prevents remote code executaion
 require(__dirname + '/shared/preventdrag.js');
 
-// eslint-disable-next-line camelcase
+declare let window: ZulipWebWindow;
```
akashnimare, line 22: What's the use case of this?

app/renderer/js/utils/system-util.ts
```diff
@@ -27,29 +31,30 @@ class SystemUtil {
...
 			}
+		} else {
+			return '';
 		}
 	}
 
-	setUserAgent(webViewUserAgent) {
-		this.userAgent = 'ZulipElectron/' + app.getVersion() + ' ' + webViewUserAgent;
+	setUserAgent(webViewUserAgent: string): void {
+		this.userAgent = `ZulipElectron/'${app.getVersion()}${webViewUserAgent}`;
```
akashnimare: @priyank-p, @vsvipul can you guys make sure this is correct?

app/main/index.ts
akashnimare: Got it, thanks.

akashnimare:
Went through the code again and left a few comments/doubts there. This work as expected on macOS. We can merge this once I test this on Windows.

app/renderer/js/preload.ts
vsvipul: We need to declare the window variable before using it. This type is declared in typings.d.ts file.

akashnimare:
We are using `@types/node` and this is what the electron docs say -

> If you were already using third-party definitions like @types/electron and @types/node, you should remove them from your Electron project to prevent any collisions.

https://electronjs.org/blog/typescript

akashnimare:
Just tested this on Windows, works well.

app/renderer/js/preload.ts
akashnimare: Got it.

app/renderer/js/utils/system-util.ts
akashnimare: I think this should be 
> `ZulipElectron/${app.getVersion()} ${webViewUserAgent}`

akashnimare (changes requested):

app/renderer/js/utils/system-util.ts
vsvipul: @priyank-p Can you fix this please ?

priyank-p:
I removed `@types/node` packages. It wouldn't cause any collision because internally electron just extends it but since we don't need it we can remove it. The `@types/electron` however will cause collision as the docs suggest.

akashnimare (approved):

akashnimare:
@priyank-p, @vsvipul great work guys. Thanks for working on this. @abhigyank, @kanishk98 thanks for reviewing 🚀

PR #783, opened 2019-06-30
reconnect-util: Gracefully reload current webview.

### LINKED ISSUE #767: Zulip drafts discarded when wifi drops
When wifi drops out while working in the OSX Zulip client, a refresh occurs and any WIP drafts are discarded. Video showing the issue here: https://www.dropbox.com/s/2r5xz7x64yytncn/zulip_deletes_drafts.mp4

priyank-p:
Before we were destroying, removing the webview elements, when we
need to reload webviews which cause unwanted side-effect for example
drafts not saving because no beforeunload event handler isn't ran.

Fixes: https://github.com/zulip/zulip-desktop/issues/767.

  - [X] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/utils/reconnect-util.js
+++ b/app/renderer/js/utils/reconnect-util.js
@@ -35,7 +35,7 @@ class ReconnectUtil {
 					.then(online => {
 						if (online) {
 							if (!this.alreadyReloaded) {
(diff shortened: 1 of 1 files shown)


akashnimare:
Great work @priyank-p 👌

PR #784, opened 2019-06-30
sidebar: Remove superfluous padding on sidebar.

priyank-p:
**What's this PR do?**
There is extra top padding of 27px which isn't needed. I don't know if we should add `2px` padding instead of none but whatever we had before was way too much?

**Screenshots?**
Before:
image
image

After:
image

### DIFF
--- a/app/renderer/css/main.css
+++ b/app/renderer/css/main.css
@@ -18,7 +18,7 @@ body {
 .toggle-sidebar {
     background: rgba(34, 44, 49, 1.000);
     width: 54px;
-    padding: 27px 0 20px 0;
+    padding: 0 0 20px 0;
     justify-content: space-between;
     display: flex;
     flex-direction: column;


vsvipul:
I think that was intentionally given spacing to allow some considerable distance from the menu bar.

akashnimare:
> I think that was intentionally given spacing to allow some considerable distance from the menu bar.

Yeah, correct the design is intentional. This also matches with the header of web app.

image

priyank-p:
OK, I get that we want more spacing from the menu but the current spacing is way too much.

(Closing, for now, we can come back it if needed)

timabbott:
I'm not sure that there's value in having it line up with the menu bar?  I think it'd probably look fine if it didn't.

@rishig what are your thoughts?

rishig:
My intuition is that less space would be better. What do other products do?

PR #785, opened 2019-07-01
Add missing transitive dependencies

snowe2010:
**What's this PR do?**

zulip was using a dependency on fs-extra that was resolving transitively
through electron. This adds the dependency explicitly. This is safer and
allows tools like pnpm to function.

https://medium.com/pnpm/pnpms-strictness-helps-to-avoid-silly-bugs-9a15fb306308

  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [x] macOS

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,6 +131,7 @@
     "electron-builder": "20.40.2",
     "electron-connect": "0.6.2",
     "electron-debug": "1.4.0",
+    "fs-extra": "8.1.0",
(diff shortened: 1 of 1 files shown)


priyank-p:
@akashnimare this should be good to merge :)

akashnimare:
Merged. Thanks, @snowe2010 :)

PR #793, opened 2019-07-17
BrowserView: Remove webview and initialize BrowserView.

vsvipul:
**What's this PR do?**
Implements BrowserView with similar functionality as webview. 
View is a class which we will use, it extends electron's BrowserView
class. ViewManager will be used to create, manage and delete views
in the main process. Removes all usage of webview and adds
an initial implementation of BrowserView.

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS


app/main/view.ts
```diff
@@ -0,0 +1,90 @@
+'use strict';
+
+import { BrowserView } from 'electron';
+
+import ConfigUtil = require('../renderer/js/utils/config-util');
+const shouldSilentWebview = ConfigUtil.getConfigItem('silent');
+
+class View extends BrowserView {
+	// TODO: Typescript - Create an interface for props here
```
priyank-p: Since you know what the props are going to be why this `TODO`? Just create the interface here.

vsvipul: Done.

app/main/view.ts
```diff
@@ -0,0 +1,95 @@
+'use strict';
+
+import { BrowserView } from 'electron';
+
+import ConfigUtil = require('../renderer/js/utils/config-util');
+const shouldSilentWebview = ConfigUtil.getConfigItem('silent');
+
+export interface PropsType {
```
priyank-p: Name this `ViewProps`.

app/main/view.ts
```diff
@@ -0,0 +1,95 @@
...
+export class View extends BrowserView {
+	props: PropsType;
+	index: number;
+	url: string;
+	zoomFactor: number;
+	loading: true;
+	badgeCount: number;
+	customCSS: boolean;
+
+	constructor(props: PropsType) {
```
priyank-p: Just a syntactic sugar you can do here: `constructor(public props: ViewProps)` and then remove `this.props = props;` below.

vsvipul: Done.

app/main/view.ts
vsvipul: Done.

vsvipul:
I think this is ready. @akashnimare @abhigyank . Should i wait for this to be merged or add subsequent commits on top of this PR ?

akashnimare:
>  add subsequent commits on top of this PR

Do this.

abhigyank:
I think you need to rebase with zulip:browserview branch otherwise the two extra commits from master would also come in the changes.

vsvipul:
> I think you need to rebase with zulip:browserview branch otherwise the two extra commits from master would also come in the changes.

I think it will be better to keep this branch updated with master as well, so that we don't face merge conflicts while merging into master later.

app/main/viewmanager.ts
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	forwardMessageAll(name: string, ...args: any[]): void {
+		for (const id in this.views) {
+			this.views[id].webContents.send(name, ...args);
+		}
+	}
+}
+
+export = new ViewManager();
```
priyank-p, line 165: Do `export default const ViewManager = new ViewManager();` or something else but not this `export =` syntax. We need to stop using this.

app/renderer/js/components/server-tab.ts
```diff
@@ -10,9 +10,8 @@ class ServerTab extends Tab {
 
 	template(): string {
 		return `<div class="tab" data-tab-id="${this.props.tabIndex}">
-					<div class="server-tooltip" style="display:none">${this.props.name}</div>
```
abhigyank, line 13: Any reason to remove this?
I see you removed all tooltips and moved to title. Any reason? Can we do the CSS for the title?

abhigyank:
This is some really good work. One important observation that I have - Moving to BrowserView shifts almost everything to the main process. Most of the functions surrounding webview previously was in `renderer/main.js` since webview was in renderer. In this change, we have kept the structure same and in places where webview functions are called we are using ipc to move those calls to the main process - even for functions like create, reload, load, switch etc. Now that browserView resides in the main process, would it make sense for us move all these important browserView actions entirely to the main process from renderer/main instead of doing ipc calls?
I am not sure if ipc calls use much CPU or memory, but I think it would make sense to have the the basic initialisation actions moved to main process.

app/renderer/js/components/server-tab.ts
vsvipul: No, we can't! Unfortunately we can't layer any HTML/CSS over BrowserViews, so this will not be possible.

app/main/viewmanager.ts
vsvipul: > I am not sure if ipc calls use much CPU or memory, but I think it would make sense to have the the basic initialisation actions moved to main process.

Yes, the initialization can be moved to main process, but switching tabs, and everything else, will still require ipc Calls.

vsvipul:
> I am not sure if ipc calls use much CPU or memory, but I think it would make sense to have the the basic initialisation actions moved to main process.

@abhigyank  Yes, the initialization can be moved to main process, but switching tabs, and everything else, will still require ipc Calls.

app/main/view.ts
```diff
@@ -0,0 +1,93 @@
+'use strict';
```
priyank-p, line 1: Minor change, but the commit message for this commit could be:
```
BrowserView: Add View wrapper class for webview.

View class extends the BrowserView class and provides interface
similar to of webview. Other functions can be later when required.
```

app/main/viewmanager.ts
```diff
@@ -0,0 +1,128 @@
+'use strict';
```
priyank-p, line 1: The commit message for this commit should be:
```
BrowserView: Add ViewManger for managing multiple View classes.

The ViewManager class makes it easier to manager multiple Views.
```

because the `ViewManager` class is not a wrapper. The summary can be extended.

priyank-p: Can you also add more details in commit message how does it handle multiple view
exactly in more detail, like resize etc.?

app/main/viewmanager.ts
```diff
@@ -0,0 +1,128 @@
...
+
+		ipcMain.on('view-destroy', (e: Event, index: number) => {
+			this.destroy(index);
+		});
+
+		ipcMain.on('view-destroy-all', () => {
+			this.destroyAll();
+		});
+
+		ipcMain.on('forward-message-view', (e: Event, name: string, ...params: any[]) => {
```
priyank-p: The event name should be `forward-view-message`?

app/main/viewmanager.ts
```diff
@@ -0,0 +1,128 @@
...
+	selectedIndex: number;
+
+	constructor() {
+		this.views = {};
+		this.selectedIndex = 0;
+		this.registerIpcs();
+	}
+
+	registerIpcs(): void {
+		ipcMain.on('view-create', (e: Event, props: ViewProps) => {
```
priyank-p: This sounds, and reads for that matter, better as `create-view`, `select-view` and `destroy-view`?

app/main/viewmanager.ts
```diff
@@ -0,0 +1,128 @@
...
+
+		ipcMain.on('view-select', (e: Event, index: number) => {
+			this.select(index);
+		});
+
+		ipcMain.on('view-destroy', (e: Event, index: number) => {
+			this.destroy(index);
+		});
+
+		ipcMain.on('view-destroy-all', () => {
```
priyank-p: I guess since I am posting review about event names, `destroy-all-views` here?

app/main/viewmanager.ts
```diff
@@ -0,0 +1,128 @@
...
+			if (typeof fn === 'function') {
+				// Type checking requires spread elements to match up with a rest parameter.
+				// So, using a workaround here.
+				(fn as any)(...params);
+			}
+		});
+	}
+
+	create(props: ViewProps): void {
+		if (this.views[props.index]) {
```
priyank-p, line 81: I think I mentioned this before but, the allowing a programmer to pass and track IDs to pass into this `index` field here is very brittle; it seems to call for trouble. Can we make this auto generated and make `readonly` property?

app/main/viewmanager.ts
```diff
@@ -0,0 +1,128 @@
...
+			const { url } = view;
+			view.webContents.loadURL(url);
+		}
+		mainWindow.setBrowserView(view);
+		view.webContents.focus();
+		this.fixBounds();
+	}
+
+	fixBounds(): void {
+		const SIDEBAR_WIDTH = 54;
```
priyank-p, line 110: I'd add a comment here mentioning that any updates here should be reflected to css and other way around.

priyank-p (commented):
@vsvipul I reviewed the first two commits and left minor comments. I don't like the passing index programmatically and having to manage it; I do rather the code handle that. Let's come up with a better model here, I guess we can come up with a better model if you describe what exactly it's needed for.

vsvipul:
> @vsvipul I reviewed the first two commits and left minor comments. I don't like the passing index programmatically and having to manage it; I do rather the code handle that. Let's come up with a better model here, I guess we can come up with a better model if you describe what exactly it's needed for.

Index is needed to create a view with that particular index so that we can easily remove and change to that view using only the index later, nothing else is required. What better model do you propose?

priyank-p:
@vsvipul OK, here is what we can do, we can have a internal variable in `view.t` say `viewIndex` that is set to 0 at first. Then when we create a view in `constructor` we can do `this.index = viewIndex;` and increment it `viewIndex++`. That way we can access the index and do whatever we want like remove, select but we just don't have to pass it in.

app/main/index.ts
```diff
@@ -262,6 +263,10 @@ app.on('ready', () => {
 		page.send('toggle-autohide-menubar', showMenubar, true);
 	});
 
+	ipcMain.on('toggle-sidebar', () => {
```
priyank-p: Do we need to do similar thing for app window resize (or is it done already)?

app/main/menu.ts
```diff
@@ -89,7 +89,7 @@ class AppMenu {
 			accelerator: process.platform === 'darwin' ? 'Alt+Command+I' : 'Ctrl+Shift+I',
 			click(_item: any, focusedWindow: any) {
 				if (focusedWindow) {
-					focusedWindow.webContents.toggleDevTools();
+					focusedWindow.webContents.openDevTools({mode: 'undocked'});
```
priyank-p: How do we handle the toggle behavior here, since this just seems to open devtools.

app/main/view.ts
```diff
@@ -17,7 +17,7 @@ export class View extends BrowserView {
 	index: number;
 	url: string;
 	zoomFactor: number;
-	loading: true;
```
priyank-p: Wow, I am surprised typescript didn't complain here.

app/main/viewmanager.ts
```diff
@@ -41,12 +41,9 @@ class ViewManager {
...
 		ipcMain.on('call-view-function', (e: Event, name: string, ...params: any[]) => {
-			const fn = this.views[this.selectedIndex][name as keyof View];
-			if (typeof fn === 'function') {
-				// Type checking requires spread elements to match up with a rest parameter.
-				// So, using a workaround here.
-				(fn as any)(...params);
-			}
+			// Type checking requires spread elements to match up with a rest parameter.
+			// So, using a workaround here.
+			(this.views[this.selectedIndex] as any)[name as keyof View](...params);
```
priyank-p: I wonder if `(this.views[this.selectedIndex] as View)` work here?

app/main/view.ts
```diff
@@ -45,6 +45,34 @@ export class View extends BrowserView {
 				this.webContents.setAudioMuted(true);
 			});
 		}
+
+		this.webContents.addListener('did-navigate-in-page', () => {
+			const isSettingsPage = this.url.includes('renderer/preference.html');
+			if (isSettingsPage) {
+				return;
+			}
+			this.canGoBackButton();
```
priyank-p: Is this function call for showing the go back button if we can, or to be more specific enable the button. If this is the case I wonder if `maybeEnableGoBackButton` would be a better name.

app/main/viewmanager.ts
```diff
@@ -67,6 +64,7 @@ class ViewManager {
 			return;
 		}
 		this.selectedIndex = index;
+		mainWindow.setBrowserView(null);
```
priyank-p: What is the story behind this change?

priyank-p (commented):
Left some more comments, I like how we get to change `tab.webview.props.name` -> `tabs.props.name`.

app/main/view.ts
vsvipul: I have used the same name as in webview.ts. Should i change this ?

app/main/view.ts
vsvipul: Removed loading completely from here and moved it to main.ts

app/main/menu.ts
vsvipul: Yes, it opens DevTools in a new window. We can't toggle it. This was the case earlier for Opening devtools for active tab. But, not we are showing devtools for active tab in the same window and for zulip app in a new window, because zulip app devtools are overlapped by the browser views.

app/main/index.ts
vsvipul:
```
view.setAutoResize({ width: true, height: true });
```
This line in fixBounds() ensured automatic resizing with window.

app/main/viewmanager.ts
vsvipul: Unsetting the old view by doing this and then setting the new view ensures that there are no glitches while switching. Can be removed too, but i guess this is better.

app/main/viewmanager.ts
vsvipul: No, can't. Getting this error-
```
Cannot invoke an expression whose type lacks a call signature. Type 'string | number | boolean | ViewProps | WebContents | (() => boolean) | ((event: string, listener: Function) => View) | ((event: string, listener: Function) => View) | ((event: string, listener: Function) => View) | ... 27 more ... | (() => string[])' has no compatible call signatures.ts(2349)
```

app/main/view.ts
priyank-p: Sure, `maybeEnableGoBackButton` sounds better.

app/main/index.ts
priyank-p: Great!

app/main/menu.ts
priyank-p: So, there is no way to toggle behavior like we had before?

app/main/viewmanager.ts
priyank-p: OK, I'd add a comment stating that here.

vsvipul: Did you add anything earlier too? I did a force push recently.

app/main/menu.ts
vsvipul: No, we can't toggle something that opens in a new window. We coudn't toggle the tools for active tab earlier, now we can't toggle for zulip app, and vice versa.

app/main/view.ts
vsvipul: Ok i'll change the name here.

app/main/viewmanager.ts
priyank-p: What do you mean by "Did you add anything"?

app/main/view.ts
abhigyank: `maybeEnableGoBackButton` sounds like a hackish function name to me.

app/main/index.ts
```diff
@@ -261,6 +264,10 @@ app.on('ready', () => {
 		page.send('toggle-autohide-menubar', showMenubar, true);
 	});
 
+	ipcMain.on('fix-bounds', () => {
```
akashnimare, line 269: Needs to document this.

app/main/menu.ts
akashnimare: This `undocked` thing open the devtools, right?

app/main/view.ts
```diff
@@ -0,0 +1,174 @@
+'use strict';
+
+import { BrowserView, BrowserWindow, app } from 'electron';
+
+import ConfigUtil = require('../renderer/js/utils/config-util');
+const shouldSilentWebview = ConfigUtil.getConfigItem('silent');
```
akashnimare: Should we set a backup value for this? In case the json gets corrupted? i.e.
```
const shouldSilentWebview = ConfigUtil.getConfigItem('silent') || false;
```

app/main/view.ts
```diff
@@ -0,0 +1,174 @@
...
+export interface ViewProps {
+	index: number;
+	url: string;
+	name: string;
+	nodeIntegration: boolean;
+	preload: boolean;
+}
+
+export class View extends BrowserView {
+	index: number;
```
akashnimare, line 22: Should we make an interface for this also?

app/main/menu.ts
vsvipul: Yes, it opens them in a new window, as we can't show them inside the window now.

app/main/view.ts
vsvipul: I picked it up from webview.ts . Is this necessary? If yes, then i'll change it.

akashnimare: Leave it for now then.

vsvipul:
@akashnimare @abhigyank @priyank-p @kanishk98 Can you pull and test this? I think this PR is complete if the last commit is good to go. (Still removing webview file is left).

app/main/viewmanager.ts
vsvipul: ```export default const ViewManager = new ViewManager();``` This doesn't seem to be working. Can you try this?

app/main/index.ts
vsvipul: Will do after the testing and bug fixing is done.

app/main/view.ts
vsvipul: Not required since we haven't used it anywhere.

app/main/viewmanager.ts
priyank-p: OK, so I think I know why this throws error. The correct way to write this would be:
```typescript
type ViewCallable = (...params: any[]) -> any;
(this.views[this.selectedIndex][name as keyof View] as ViewCallable)(...params);
```

priyank-p: Since we do this in couple of places I do split this off into it's own function and also check if it a function before calling it too. So something like:
```typescript
callViewFunction(index: number, name: keyof View, ...params; any[]) {
    const viewFunction = this.views[index][name];
    if (typeof viewFunction === 'function') {
        type ViewCallable = (...args: any[]) -> any;
        (viewFunction as ViewCallable)(...params);
    } else {
         // error reporting here
    }
}
```

app/main/viewmanager.ts
priyank-p: So I made a mistake here, this could be either `export default new ViewManager()` or
```typescript
const viewManager = new ViewManager();
export default viewManager();

// then in index.ts
import ViewManager from './viewmanager';
```

I prefer the second one just because `export default new ViewManager();` looks awkward to me (but that just me).

app/main/index.ts
```diff
@@ -279,7 +285,7 @@ app.on('ready', () => {
 		AppMenu.setMenu(props);
 		const activeTab = props.tabs[props.activeTabIndex];
 		if (activeTab) {
-			mainWindow.setTitle(`Zulip - ${activeTab.webview.props.name}`);
+			mainWindow.setTitle(`Zulip - ${activeTab.props.name}`);
```
abhigyank, line 291: It seems on the BrowserView, it doesn't auto decode encoded text (in case of abhigyank.zulipdev.org) so we see escaped HTML on the title. For temporary fix we can decode the title here with a comment and create an issue upstream to decode the url. It used to work webview so should work here as well.

abhigyank:
Bug - If you resize the window, all other servers than the active one gets reloaded. Same is the case with hiding sidebar or menu bar. If you come back to the active server after switching that reloads too. Though I don't feel that its a full reload or something cause its quite fast. 

Rest tab switching and everything are working fine and smooth. I couldn't find any bugs. Haven't gone through the implementation and code yet but seems to be working okay. @priyank-p Maybe you can give it a try if you can find something broken.

app/main/index.ts
vsvipul: Fixed this in the latest commit. We weren't decoding it when we were pushing it to tabs.

vsvipul:
> Bug - If you resize the window, all other servers than the active one gets reloaded. Same is the case with hiding sidebar or menu bar. If you come back to the active server after switching that reloads too. Though I don't feel that its a full reload or something cause its quite fast.

This is known. Still looking into what is causing this. Couldn't find what is causing this weird behavior @priyank-p  .

vsvipul:
> Bug - If you resize the window, all other servers than the active one gets reloaded. Same is the case with hiding sidebar or menu bar. If you come back to the active server after switching that reloads too. Though I don't feel that its a full reload or something cause its quite fast.

This has something to do with our preload script. Try logging out of all orgs and do the same. It will be fixed. It isn't something related to BrowserView. It is something in our own preload script (specifically spellchecker i think, which is causing this problem). What should i do about it?

app/main/viewmanager.ts
vsvipul: I have made a separate function in the last commit, but couldn't make this work. Get webcontents not defined error when i use this. See if you can make it work.

akashnimare:
> This has something to do with our preload script. Try logging out of all orgs and do the same. It will be fixed. It isn't something related to BrowserView. It is something in our own preload script (specifically spellchecker i think, which is causing this problem). What should i do about it?

@vsvipul can you reproduce this on master?

vsvipul:
> @vsvipul can you reproduce this on master?

No, i can't reproduce this on master. What i meant to say was that, it has something to with spellchecker and BrowserView compatibility with each other that seems to causing this. I've been trying to fix this from the past week, but couldn't fix it. If someone can please help, it'd be great!

kanishk98:
> What i meant to say was that, it has something to with spellchecker and BrowserView compatibility with each other that seems to causing this.

@vsvipul what makes you say that? I commented out the spellchecker related actions in `preload.ts` but that didn't solve the bug. 
I think it's probably something else - I'll let you know if I find something. :)

vsvipul:
> @vsvipul what makes you say that? I commented out the spellchecker related actions in `preload.ts` but that didn't solve the bug.
> I think it's probably something else - I'll let you know if I find something. :)

Sorry, i just tested it again and i guess i was wrong! @akashnimare  Is this required to be fixed before merging this PR? I'll be adding the documentation today.

app/renderer/js/components/handle-external-link.ts
```diff
@@ -22,8 +20,6 @@ function handleExternalLink(this: any, event: any): void {
 	} = LinkUtil.isInternal(domainPrefix, url);
 
 	if (isWhiteListURL) {
-		event.preventDefault();
```
akashnimare, line 25: Can't remember exactly why we added this but I think it has something to do with the race condition where the link was being opened in the browser + the desktop app. Please cross-check it.

app/renderer/js/components/handle-external-link.ts
```diff
@@ -8,10 +8,8 @@ const { shell, app } = remote;
 
 const dingSound = new Audio('../resources/sounds/ding.ogg');
 
-// TODO: TypeScript - Figure out a way to pass correct type here.
-function handleExternalLink(this: any, event: any): void {
-	const { url } = event;
```
akashnimare, line 13: I'm guessing it's a legacy code which we forgot to remove?

app/main/index.ts
vsvipul: DOne.

app/renderer/js/components/handle-external-link.ts
vsvipul: Yes, typescript detected that here, and it is passed correctly now.

app/renderer/js/components/handle-external-link.ts
vsvipul: For what type of link do i need to test this?

app/main/viewmanager.ts
```diff
@@ -0,0 +1,161 @@
...
+
+		ipcMain.on('forward-message-all', (e: Event, name: string, ...params: any[]) => {
+			this.forwardMessageAll(name, ...params);
+		});
+
+		ipcMain.on('call-view-function', (e: Event, name: string, ...params: any[]) => {
+			this.callViewFunction(this.selectedIndex, name, ...params);
+		});
+
+		ipcMain.on('call-specific-view-function', (e: Event, index: number, name: string, ...params: any[]) => {
```
akashnimare, line 53: Please document this newly added IPCs -
> forward-message-all
> call-view-function
> call-specific-view-function
> forward-view-message

@priyank-p happy with the naming of ipc functions?

vsvipul: Documented these.

app/renderer/js/components/tab.ts
```diff
@@ -27,17 +23,14 @@ class Tab extends BaseComponent {
...
 		this.$el.classList.add('active');
-		this.webview.load();
 	}
 
 	deactivate(): void {
 		this.$el.classList.remove('active');
-		this.webview.hide();
 	}
 
 	destroy(): void {
```
akashnimare, line 35: Are we using these listeners anywhere?
> activate
> deactivate
> destroy

If yes, please document this.

vsvipul: Done.

app/main/viewmanager.ts
abhigyank: I think we can rename `forward-view-message` to `forward-message-view` and `forward-message-all` to `forward-message-all-views`. Similarly the `forwardMessageAll` to `forwardMessageAllViews`. @akashnimare thoughts?

app/main/index.ts
```diff
@@ -108,10 +110,12 @@ function createMainWindow(): Electron.BrowserWindow {
 	win.setTitle('Zulip');
 
 	win.on('enter-full-screen', () => {
+		ViewManager.fixBounds();
```
akashnimare, line 113: Can you document the bound thing here and what it do etc?

app/main/viewmanager.ts
akashnimare: @vsvipul is this being taken care of?

app/renderer/js/components/handle-external-link.ts
akashnimare: Try to test the following links -
* Internal url - Click on internal zulip link like send a message to yourself with something like https://chat.zulip.org/#narrow/stream/112-desktop-issues and click the same.
* Click on an attachment link.
* Some external URL.

app/main/index.ts
vsvipul: I have documented this here. https://github.com/zulip/zulip-desktop/blob/e4f025166f882ffa9bd6fcde73fe19d4d7572d76/app/main/viewmanager.ts#L106

akashnimare:
This is getting pretty close to being merged. @vsvipul please take care of the conflicts. @abhigyank, @kanishk98, @priyank-p would be great if you guys can test the PR from your side.

akashnimare:
We have to look for the potential issues in electron-builder which could affect our app since we are using v21 of electron-builder -
https://github.com/electron-userland/electron-builder/issues/4185

akashnimare:
@vsvipul  I get this after running `npm start` -

https://cl.ly/08d596abfc9a

Tried reinstalling node modules and `npm run clean-ts-files` but no luck.

vsvipul:
@akashnimare This is working fine for me. Wonder what's causing the problem for you. @kanishk98  @abhigyank  @priyank-p  can you guys test once again as well? Let's wrap this up.

akashnimare:
@vsvipul I'm still stuck to that page. For me, even the setting page is not loading. No error in console.

image

app/renderer/js/components/handle-external-link.ts
akashnimare: @vsvipul Is this being tested?

akashnimare:
@kanishk98 can you check if this PR works for you on macOS? This works for me on Linux but something is wrong on macOS.

kanishk98:
Yeah, it doesn't work on macOS for me either. (screenshot attached below)
I remember it working on Windows last time I tried it, though. @vsvipul are we using Windows/Linux specific APIs somewhere?

Screenshot 2019-09-18 at 5 36 29 PM

akashnimare:
In that case, we should check with electron v5 and see if they have made any changes.

akashnimare:
@kanishk98 let's try to find out which commit broke the app on macOS.

kanishk98:
@akashnimare I'm doing that right now. I don't think it's something to do with Electron, though - this seems to stem from a bug in `activateTab()`. I'm trying to fix it right now, will text you and Vipul on czo if I find something.

akashnimare:
@kanishk98 any luck in debugging?

kanishk98:
> any luck in debugging?

Yeah, the app breaks at commit 9c4a3de835d0e7f20948e07982d980f03bb2dfc7. I'm currently working to see where the Mac-specific problem is.

kanishk98:
Ah okay, the solution is almost too simple:

In `viewmanager.select()`, we need to reverse the order of function calls from 

```ts
this.fixBounds();
mainWindow.setBrowserView(view);
```

to

```ts
mainWindow.setBrowserView(view);
this.fixBounds();
```

@vsvipul can you make the changes? I just tried this out on a Mac - it should solve our problem.

akashnimare:
@kanishk98 are you sure this works for you? Can you post a gif here?

kanishk98:
> Can you post a gif here?

I can't post a gif right now (it's not my Mac that I tested this on), but you can fetch my branch [browserview](https://github.com/kanishk98/zulip-desktop/tree/browserview) onto your system and try it out. Do let me know if it doesn't work.

akashnimare:
@vsvipul can you fix the merge conflicts?

vsvipul:
@akashnimare working on it right now. Sorry for the late response. Was a little busy this week.

vsvipul:
I am having some difficulty porting the new notification option to browser view.

kanishk98:
> I am having some difficulty porting the new notification option to browser view.

I think I can help with that. Can you tell me what the issue is?

vsvipul:
@kanishk98 How do i set the enabled state in the rendered process for that setting because the current URL in browser view is in the main procesS?

kanishk98:
I can think of the following methods:

1. Use a synchronous IPC call (probably the simplest, but I'd avoid this). 
2. Save a `loggedIn` flag in `domain.json` for all orgs that you can read when deciding whether `enabled` is true or not. Said flag is updated when the URL of the view changes. 
3. I'm a little doubtful about this one, but we can also maintain an instance variable in `ViewManager` that maps server URLs to `loggedIn` status. Updated in the same way as (2). 

Since I'm also working on this right now, I'll try what works on my end and let you know. Text me on czo if you have any further questions.

akashnimare:
@kanishk98, @vsvipul awesome. Let's try to merge this asap since it's a blocker for other PRs.

vsvipul:
@akashnimare  Apparently, rebasing it automatically messes up my whole branch. The title property which I added gets removed, function names get changed due to which many ipc Calls stop working. I'll have to do this commit by commit myself. So, it is going to take some time.

akashnimare:
@vsvipul maybe reset the branch locally and git pull + rebase and fix merge conflicts? There are three files mainly where the conflicts thing happening.

vsvipul:
@akashnimare Yeah, i did exactly that, but that still messes up a lot of stuff. I am doing it but will take some time.

akashnimare:
@vsvipul let me know if you need help. Gotta merge this asap since a lot of other PRs are still pending.

vsvipul:
@akashnimare As I am currently a little busy with academic work, I have given access to my branch to @kanishk98 a few days back. He is working on rebasing and migrating the notification option he added earlier.

akashnimare:
Cool, thanks for the update!

vsvipul:
Closing this in favor of #831 .

andersk:
You can't close #793 in favor of #793. You mean #831.

vsvipul:
Yes @andersk . Thanks. Updated.

PR #804, opened 2019-08-02
preference: Fix visually broken delete custom css button.

### LINKED ISSUE #801: Fix alignment and look of Delete CSS Button in Settings
---

- **Operating System**:
  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS
Here, we can see the the Delete button is not aligned with other options and is also bigger than other buttons.
Screenshot from 2019-07-31 03-46-01

priyank-p:
There were couple of issue solved by this commit:
  - The button had extra 10px margin to right; this comes from .action class
  - The "Delete" text and icon were not aligned
  - The button looks bigger than other buttons

Both of which made the button look broken. We fix first by changing
.css-delete-action to #css-delete-action in css file; git grep shows
this is only used once and is used as id not a class. Then we use
display flex to fix the alinging of icon and text. We add box-sizing border-box
which solves the bigger width issue. We remove extra 10px margin that comes from
.action.

Fixes: #801.

**Screenshots?**
image

### DIFF
--- a/app/renderer/js/pages/preference/general-section.ts
+++ b/app/renderer/js/pages/preference/general-section.ts
@@ -104,7 +104,6 @@ class GeneralSection extends BaseSection {
 				<button class="custom-css-button green">${t.__('Upload')}</button>
 			</div>
 			<div class="setting-row" id="remove-custom-css">
-		"S
 				<div class="setting-description">
 					<div class="selected-css-path" id="custom-css-path">${ConfigUtil.getConfigItem('customCSS')}</div>
 				</div>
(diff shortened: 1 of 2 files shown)


app/renderer/js/pages/preference/general-section.ts
```diff
@@ -104,7 +104,6 @@ class GeneralSection extends BaseSection {
 				<button class="custom-css-button green">${t.__('Upload')}</button>
 			</div>
 			<div class="setting-row" id="remove-custom-css">
-		"S
```
priyank-p, line 107: This was accidentally included by @kanishk98 in the translation PR 😄!

akashnimare: Nice catch :)

akashnimare:
Merged. Thanks, @priyank-p, @PrakharUniyal.

PR #809, opened 2019-08-14
Added mute-per-organization support 

muskankhedia:
---

**What's this PR do?**

Solves: #623 

This feature enables the user to mute a particular organization, on muting the organization the user will not be displayed the badge further for that organization. User can mute the organization in 2 ways: 

1. Using the per-organization context menu.
2. Clicking the mute button in the organization settings option 

In order to implement this feature, 2 properties are added in *domain.json*

1. **muted:** stores boolean value
2. **badgeCount:** stores the webview badgecount when the organization was muted

**Screenshots?**
Aug-14-2019-6_02-PM

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS


app/renderer/js/main.ts
```diff
@@ -743,6 +743,33 @@ class ServerManagerView {
 						});
 					}
 				},
+				{
+					label: (DomainUtil.getDomain(index).muted ? 'Unmute' : 'Mute'),
+					click: () => {
+						console.warn("HIII");
+						console.log(index);
+						console.warn(DomainUtil.getDomain(index).muted);
```
kanishk98: Looks like you left over some `console` related actions here. Please remove them and then amend the commit.

app/renderer/js/main.ts
```diff
@@ -743,6 +743,30 @@ class ServerManagerView {
...
+							if (response === 0) {
+								if (DomainUtil.getDomain(index).muted) {
+									// server is already muted
+									this.tabs[index].webview.updateBadgeCount(DomainUtil.getDomain(index).badgeCount);
+									DomainUtil.updateBadgeCount(index, -1);
+									DomainUtil.updateMute(index, false);
+								} else {
+									DomainUtil.updateBadgeCount(index, this.tabs[index].webview.badgeCount);
+									DomainUtil.updateMute(index, true);
+									this.tabs[index].webview.updateBadgeCount(0);
```
kanishk98: Why do we update the `badgeCount` to 0 here?

app/renderer/js/main.ts
```diff
@@ -961,6 +985,19 @@ class ServerManagerView {
 		ipcRenderer.on('new-server', () => {
 			this.openSettings('AddServer');
 		});
+
+		ipcRenderer.on('mute-org', (event: Event, index: number) => {
+			if (DomainUtil.getDomain(index).muted) {
+				// server is already muted
+				this.tabs[index].webview.updateBadgeCount(DomainUtil.getDomain(index).badgeCount);
+				DomainUtil.updateBadgeCount(index, -1);
+				DomainUtil.updateMute(index, false);
```
kanishk98: Let's define a function that takes care of this update instead of writing the same code repeatedly?

app/renderer/js/main.ts
```diff
@@ -961,6 +985,19 @@ class ServerManagerView {
 		ipcRenderer.on('new-server', () => {
 			this.openSettings('AddServer');
 		});
+
+		ipcRenderer.on('mute-org', (event: Event, index: number) => {
+			if (DomainUtil.getDomain(index).muted) {
+				// server is already muted
+				this.tabs[index].webview.updateBadgeCount(DomainUtil.getDomain(index).badgeCount);
+				DomainUtil.updateBadgeCount(index, -1);
```
kanishk98: Can you explain the -1 to me here? As far as I understand, you're saving the `badgeCount` in `domain.json` so that you can instantly show the correct message count when the user decides to unmute an organisation. 
Is there a problem with saving the current `badgeCount` instead of -1?

app/renderer/js/components/webview.ts
```diff
@@ -7,6 +7,7 @@ import ConfigUtil = require('../utils/config-util');
 import SystemUtil = require('../utils/system-util');
 import BaseComponent = require('../components/base');
 import handleExternalLink = require('../components/handle-external-link');
+import DomainUtil = require("../utils/domain-util");
```
kanishk98: Nitpicking here, but we use single quotes throughout the app. 
I think we've missed a test in `xo`, I'll check this asap.

app/renderer/js/components/webview.ts
```diff
@@ -147,7 +148,16 @@ class WebView extends BaseComponent {
 		});
 	}
 
+	updateBadgeCount(count: number): void {
+		this.badgeCount = count;
+		this.props.onTitleChange();
```
kanishk98, line 153: Would it be a good idea to keep updating `domain.json` with the correct badge count here too so that when the user switches on notifications again, they get the correct number immediately even if the page hasn't loaded?

kanishk98 (changes requested):
@muskankhedia thanks for your work! I see you're new to Zulip, and I'm very glad that you chose to start by contributing to such a major feature. 

The PR is pretty good, and I've left some comments that I'd like you to consider. 

Just a heads-up, though: we've been having discussions regarding the mute feature, and we will most probably choose a feature where a server-side setting and a desktop app feature work directly with each other (instead of locally, as the PR stands now). I'll let you know here or on czo when we start work with that, and you can follow [this](https://chat.zulip.org/#narrow/stream/16-desktop/topic/Mute.20an.20entire.20org.3F) thread for now.

app/renderer/js/main.ts
muskankhedia: Yaa.. it as been removed in the second commit.

app/renderer/js/main.ts
muskankhedia: Here the badge-count in the webview has been updated to zero, in order to hide the badge from the organization.

app/renderer/js/main.ts
muskankhedia: Yaa, the badgecount is updated in domain.json to instantly show the correct message count when the user decides to unmute an organization, updating the value to -1 denotes that the particular organization is not muted. This is maintained to avoid confusion and to keep a check if the value is updated on muting and unmuting an organization.

app/renderer/js/components/webview.ts
muskankhedia: corrected it!

app/renderer/js/components/webview.ts
muskankhedia: the `updateBadgeCount()` in `webview.ts` is used by the muting the feature, so the value at the `domain.json` will have no effect on adding it to the function. Further, the function itself is dependent on the called parameters, so there will be no changes.

app/renderer/js/main.ts
muskankhedia: I have added the functions in the `main.ts` file, the functions are: `muteOrg()` and `unMuteOrg()`

kanishk98:
Thanks again for your work, @muskankhedia! While we were originally thinking of a desktop-only setting for muting organisations, we've since moved to a more integrated approach where such settings are governed by the web app and so all clients can pick up on that. 
[Akash's message](https://chat.zulip.org/#narrow/stream/16-desktop/topic/Mute.20an.20entire.20org.3F/near/779552) handles it pretty clearly. 

For future contributions, I'd suggest you use either GitHub issues or Zulip itself to discuss features before you write them. I've made this mistake many times too, but it'll only slow you down by forcing you to revise features continuously. 

Hope that's not a problem, and welcome to Zulip!

muskankhedia:
> Thanks again for your work, @muskankhedia! While we were originally thinking of a desktop-only setting for muting organisations, we've since moved to a more integrated approach where such settings are governed by the web app and so all clients can pick up on that.
> [Akash's message](https://chat.zulip.org/#narrow/stream/16-desktop/topic/Mute.20an.20entire.20org.3F/near/779552) handles it pretty clearly.
> 
> For future contributions, I'd suggest you use either GitHub issues or Zulip itself to discuss features before you write them. I've made this mistake many times too, but it'll only slow you down by forcing you to revise features continuously.
> 
> Hope that's not a problem, and welcome to Zulip!

Thanks, for further contributions I will surely keep a note on this.

app/renderer/js/components/webview.ts
vsvipul: THis might not be a very good idea. THe badgecount in the title changes when we scroll down a page and accessing the json-db this many times, is not a good idea.

vsvipul:
As it has already been decided to do a version of this integrated with the webapp, i think this can be closed for now. @kanishk98  you can work with @muskankhedia to complete the PR.

app/renderer/js/components/webview.ts
muskankhedia: The json-db will be accessed only at the times of muting and unmuting an organization. Once its unmuted, it will update the value according to the present number of unread messages.

akashnimare:
@muskankhedia thanks for working on this. Feel free to collaborate with @kanishk98 on this :)

PR #812, opened 2019-08-21
sidebar: Add option to open notification setting from the context menu.

kanishk98:
---

**What's this PR do?**

Adds an option to show notifications settings for an org to the menu. 
If "none" is chosen in unread count summary, the app does not show a badge count or desktop notifications from that organisation. 

UI suggestion [here](https://chat.zulip.org/#narrow/stream/16-desktop/topic/Mute.20an.20entire.20org.3F/near/779966). 
The topic for the above message also has our discussion concerning the web app. 

Corresponding web app PR at [#13044](https://github.com/zulip/zulip/pull/13044). 

**Problems**

While the current implementation follows the discussed method, I think it'll be much better if we chose to make a native settings page to handle notifications and connected that with the same settings on the web app. I'm not sure why, but the notification settings modal is slow to load and sometimes fails altogether. Although we show the rotating loading icon in the sidebar, I don't think it's good enough. 

Let me know what you guys think. Please do review the `notify` field addition though. 

@muskankhedia it'll be great if you can have a look too. :)

  - [x] Windows
  - [x] Linux/Ubuntu (do desktop notifications get muted on Linux? They didn't for me.)
  - [ ] macOS


muskankhedia:
@kanishk98 Thanks for mentioning me in this PR. I had a look at it and understood the algo you are following to implement mute property in zulip desktop app.

app/renderer/js/main.ts
```diff
@@ -426,6 +426,11 @@ class ServerManagerView {
 		return this.tabs[this.activeTabIndex].webview.props.url;
 	}
 
+	shouldCurrentActiveServerNotify(): boolean {
+		const { notify } = DomainUtil.getDomain(this.activeTabIndex);
+		return notify === undefined ? false : notify;
```
akashnimare: This looks confusing. Can you make it a bit clearer (add a comment or two/make the code better).

kanishk98:
Just noticed a couple of problems with this PR. I'll mark this as WIP for now and review the discussions we had regarding this feature on czo.

akashnimare:
@kanishk98 I tested this and found out -
* It takes a while to open the notification page since we are loading a new src. This also affects the notification count and app hangs for a few seconds which is bad.
* Instead of loading a new src I would suggest we should click on the setting element using the JS (same as logout thing) that way we can load that page instantly. 

Clicking on setting page via JS -

akashnimare:
@kanishk98 can you make the changes?

akashnimare:
This is how it looks now -

https://cl.ly/e3ef3f323945

FYI, @rishig.

timabbott:
Looks great to me as well!  Thanks for adding this.

PR #814, opened 2019-08-28
domains: Prompt user on old server invalidation.

### LINKED ISSUE #813: The app should check whether it's actually loading Zulip
- **Operating System**:
  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS 

Easier for me to just show you:
image

Basically what happened here was that Riot got pushed to the URL that once held a Zulip instance. Now instead of giving an error that it's not connected to Zulip, the Zulip app is running Riot instead. 

A similar issue ocurrs when you right click an image in Zulip and press View Image, which opens the image, forcing you to look for the back button.

The funny part? Riot works just fine. I've yet to run into any issues.

kanishk98:
Fixes #813.

On app load, we call `checkDomain()` every time we check for updates to the server icon and can detect if the new URL (possibly caused by corruption of `domain.json` or the server admin moving the address of the Zulip server) is not a legal Zulip org. This PR prompts the user to remove such an organisation and re-add the same after checking the URL. 

We also add an `override` parameter for this, so if there's an enterprise user facing this issue, they can also remove that organisation. This does allow them to remove an organisation using a hack - they could go to `domain.json` (which they can access since it's in their user folder), change the URL to say `google.com`, then have the app prompt them to remove the server. I'm not sure what we should do here, and implemented the override anyway. Let me know what you think. :)

---

**Screenshots?**

image

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/invalid-server.html
+++ b/app/renderer/invalid-server.html
@@ -1,0 +1,20 @@
+<!DOCTYPE html>
+<html lang="en" class="responsive desktop">
+    <head>
+        <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+        <meta name="viewport" content="width=device-width">
+        <title>Zulip - Invalid Server</title>
+        <link rel="stylesheet" href="css/network.css" type="text/css" media="screen">
+    </head>
+    <body>
+        <div id="content">
+            <div id="picture"><img src="img/zulip_network.png"></div>
+            <div id="title">Zulip can't connect</div>
+            <div id="description">
+                <div>Looks like this organization isn't a Zulip server.</div>
+                <div>We suggest you remove this organization, contact your server administrator, and re-add the organization with the correct URL.</div>
+            </div>
+        </div>
+    </body>
+    <script>require('./js/shared/preventdrag.js')</script>
+</html>
--- a/app/renderer/js/components/webview.ts
+++ b/app/renderer/js/components/webview.ts
@@ -147,6 +147,10 @@ class WebView extends BaseComponent {
 		});
 	}
 
+	showInvalidServer(): void {
+		this.$el.src = `file://${this.props.rendererDirectory}/invalid-server.html`;
+	}
+
 	getBadgeCount(title: string): number {
 		const messageCountInTitle = (/\((\d+)\)/).exec(title);
 		return messageCountInTitle ? Number(messageCountInTitle[1]) : 0;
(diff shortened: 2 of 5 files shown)


akashnimare:
@kanishk98 is it possible to not allow the server to be loaded? i.e. in screenshot I can see the google.com has been loaded but I think it should not happen?

kanishk98:
Yeah, we can disallow the server from being loaded, but the most obvious way to do that is to wait till the `updateSavedServer()` method resolves. Since that would be bad UX (and would slow down users in almost all cases, where the server URLs are correct after addition), I didn't do that. 

We could choose to show an error screen (similar to our network failure screen) in the webview instead. Users won't get slowed down by that if their URL is correct, and even if Google does load, it'll be replaced fairly quickly. 
Let me know what you think.

akashnimare:
> We could choose to show an error screen (similar to our network failure screen) in the webview instead. Users won't get slowed down by that if their URL is correct, and even if Google does load, it'll be replaced fairly quickly.
Let me know what you think.

Sounds good 👍

kanishk98:
Updated the PR. Here's what the invalid server screen looks like:

image

kanishk98:
@akashnimare whenever you find time, please also let me know what you think of overriding the `removeDomain()` function when the user clicks on Yes. I'm not sure if that's a good idea.

timabbott:
This PR seems somewhat risky; we don't want a transient network failure to result in an error encouraging users to disconnect the organization.  I like the non-modal screen, but I think we should have much clearer text about the range of possibilities, e.g.:

title: "Zulip cannot connect to {{domain}}"

This could be because:
* You're not on the Internet or your proxy is misconfigured
* There is no Zulip organization hosted at this URL
* This Zulip organization is temporarily unavailable
* This Zulip organization has moved or been deleted

[button: retry connection] [button: Go to settings]

kanishk98:
Fair enough, I didn't think about the transient network issues. 
I'll remove the modal entirely, and also update the text on the screen. 

Thanks @timabbott! :)

akashnimare:
@kanishk98 can you please update the PR addressing Tim's comment?

kanishk98:
@akashnimare I've included the fixes included in this PR in #723 instead as we'd discussed earlier (that fixes our reconnection logic too). 
I'll be fixing the `nodeIntegration` issue with that PR tonight, so we should be able to merge both fixes then.

akashnimare:
Okay, in that case, we should close this PR?

kanishk98:
> in that case, we should close this PR?

Yeah, I just updated #723. Please have a look at that instead. I'll close this. :)

PR #818, opened 2019-08-30
system-util: Set User-Agent from main process.

### LINKED ISSUE #817: Set User-Agent when making requests to /server_settings
Our server logs contain a bunch of entries like this:

`/api/v1/server_settings (unauth via Unspecified)`

("Unspecified" means no `User-Agent` header was set in the HTTP request)

Checking what those clients later requested, it's clear that the cause is ZulipElectron not setting that header.  I assume this bug is still in master, but if not, we should fix it... longer-term, we will likely require setting a `User-Agent` in all API requests.

kanishk98:
**What's this PR do?**

* Sets user-agent config item when the app's DOM is ready.
* App sends the right User-Agent to the server-settings API.
* Fixes #817.

Please have a look at #817.

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/index.ts
+++ b/app/main/index.ts
@@ -8,7 +8,7 @@ import path = require('path');
 import fs = require('fs');
 import isDev = require('electron-is-dev');
 import electron = require('electron');
-const { app, ipcMain } = electron;
+const { app, ipcMain, session } = electron;
 
 import AppMenu = require('./menu');
 import BadgeSettings = require('../renderer/js/pages/preference/badge-settings');
@@ -181,6 +181,10 @@ app.on('ready', () => {
 			mainWindow.hide();
 		} else {
 			mainWindow.show();
+		}
+		if (!ConfigUtil.isConfigItemExists('userAgent')) {
+			const userAgent = session.fromPartition('webview:persistsession').getUserAgent();
+			ConfigUtil.setConfigItem('userAgent', userAgent);
 		}
 	});
 
--- a/app/renderer/js/utils/system-util.ts
+++ b/app/renderer/js/utils/system-util.ts
@@ -2,6 +2,7 @@
 import { remote } from 'electron';
 
 import os = require('os');
+import ConfigUtil = require('./config-util');
 
 const { app } = remote;
 let instance: null | SystemUtil = null;
@@ -53,6 +54,9 @@ class SystemUtil {
 	}
 
 	getUserAgent(): string | null {
+		if (!this.userAgent) {
+			this.setUserAgent(ConfigUtil.getConfigItem('userAgent', null));
+		}
 		return this.userAgent;
 	}
 }


app/renderer/js/utils/system-util.ts
```diff
@@ -53,6 +54,9 @@ class SystemUtil {
 	}
 
 	getUserAgent(): string | null {
+		if (!this.userAgent) {
+			this.setUserAgent(ConfigUtil.getConfigItem('user-agent', null));
+		}
 		return this.userAgent;
```
akashnimare: Would something like this make sense 🤔 - 
```js

getUserAgent(): string | null {
		return this.userAgent || this.setUserAgent(ConfigUtil.getConfigItem('user-agent', null));
}
```

akashnimare:
Okay, correct me if I'm wrong. This is what I think -

a) We only set the user-agent when we have a server/org/webview. 
b) We don't have user agent in this cases - When the app loads/check for the server icon/when there is no server added etc.
c) As far as I know, we are sending the user-agent with each and every request.

Your PR takes care of case b), right? You're setting the user-agent in the main process itself so that we always have something in the user-agent.


Can't we do something like this -

```js
	setUserAgent(webViewUserAgent: string): void {
		if (webViewUserAgent) {
			this.userAgent = `ZulipElectron/${app.getVersion()} ${webViewUserAgent}`;
		} else {
			this.userAgent = ConfigUtil.getConfigItem('user-agent', null);
		}
	}

	getUserAgent(): string | null {
		return this.userAgent;
	}

```

kanishk98:
Sure, that works too. I can update the PR right away to your suggestion.

akashnimare:
@kanishk98 The only problem I see is we now have two user-agent. Would it be possible to set the same user-agent (which will atleast contains ZulipElectron) so that webapp knows that the request are coming from the desktop app? @timabbott do we check for string `ZulipElectron` while checking for the user-agent (for desktop app) in the webapp?

timabbott:
We should definitely use the same `ZulipElectron` with version User-Agent string everywhere in the app.  The whole point of this bug report was that there was a case where it was not doing that (leaving it unset).  Having two different values would be even worse than the status quo :). We do indeed parse `User-Agent` in the webapp in some places (e.g. for logging what client is making a request that has errors to help debug why we're getting errors), and want the ability to check it in more in the future (e.g. to detect when we need to do something backwards-compatible).

akashnimare:
@kanishk98 in that case, I think we should just append the ZulipElectron with version number here -
https://github.com/zulip/zulip-desktop/pull/818/files#diff-f2a194c55b6b05a5c7255b3fb971b694R186

akashnimare:
@timabbott this PR should fix the user-agent issue. Please let us know if you encounter something fishy with the desktop app's user-agent.

timabbott:
Awesome, thanks!

app/main/index.ts
```diff
@@ -182,6 +182,10 @@ app.on('ready', () => {
 		} else {
 			mainWindow.show();
 		}
+		if (!ConfigUtil.isConfigItemExists('userAgent')) {
```
manavmehta, line 185: @akashnimare 
The flaw is here I think. On upgradation, UA would _exist_ in settingsJSON and thus will not be updated

PR #822, opened 2019-09-23
certificate: Make certificate location dynamic.

akashnimare:
---

**What's this PR do?**

**Screenshots?**

  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS


app/translations/en-GB.json
```diff
@@ -0,0 +1,118 @@
+{
```
kanishk98, line 1: We should unstage this file, right?

kanishk98 (approved):
This surely solves the problem, but I'm not sure how to generate a self-signed cert that'll help us test this PR. I'd be glad to help out, though - please text me the instructions on czo if required.

app/translations/en-GB.json
akashnimare: Yeah

akashnimare:
I have asked Tim to create an instance for testing this feature.

akashnimare:
@kanishk98 can you test this against - https://abhigyank.zulipdev.org ?

akashnimare:
This works as expected for me. We also need to handle the case where the certs are already in db.

akashnimare:
@muskankhedia can you also test this? I'm bit confused about the case when the user already have the cert installed.

muskankhedia:
Hello @akashnimare, I tried testing the code in my system (currently using OS: Fedora 30) and facing some of the problems with the connected server. Please have a view on the image below regarding details of the error in the console.
Screenshot from 2019-10-13 17-29-00
I don't have any certificates for any of the servers (I have also checked in the `.config/Zulip/certificates` and `.config/Zulip/config/certificates.json` both are empty). I didn't face this error in the console earlier ever. Please have a check on the error.

muskankhedia:
I am also unable to use the server https://abhigyank.zulipdev.org for testing purposes as I am unable to signup into the same. Due to any reason, I am not receiving the confirmation mail after signing up in the server(I have also checked the spam).

akashnimare:
Wait, let me send you the cert first.

app/renderer/js/utils/request-util.ts
```diff
@@ -38,11 +43,13 @@ class RequestUtil {
 		const certificate = CertificateUtil.getCertificate(
 			encodeURIComponent(domain)
 		);
+		const certificateFile = path.join(`${app.getPath('userData')}/certificates`, certificate);
```
kanishk98: I reviewed this again, and to solve the issue with older versions you pointed out to me, I think we should just replace this line with the following:

```ts
                let certificateFile = null;
		if (certificate && certificate.includes('/')) {
			// certificate saved using old app version
			certificateFile = certificate;
		} else if (certificate) {
			certificateFile = path.join(`${app.getPath('userData')}/certificates`, certificate);
		}
```

kanishk98: This works well for me on my end. @muskankhedia it'll be great if you can also try this out. Add the certificate to the app on `master`, checkout to this PR, then disconnect the concerned org and try adding it back. You should not get a prompt asking for you to confirm going ahead with adding the org.

akashnimare:
Made the changes. @kanishk98, @muskankhedia can you quickly cross-check the two cases? Also, let me know if it's GTM.

vaskokj:
Where is the cert supposed to go on the client side?

kanishk98:
Checked the following 3 cases:

1. When the user is on the new PR and then certificate is added. ✔️ 
2. When the user is on `master` with existing certificate and added org and then switches to the new PR. No changes. ✔️ 
3. When the user is on `master` with existing certificate and removed org and then switches to the new PR and tries to re-add the org. Old certificate detected. ✔️

kanishk98:
I think it's GTM.

vaskokj:
@akashnimare @kanishk98 What is the use-case of these changes?

akashnimare:
@vaskokj the current version have hard-coded location of certificate. The app request certificate from -
```
/Users/USERNAME/Library/Application Support/Zulip/certificates
```

So you see the `USERNAME` here was creating the issue since you can't install on one system and copy-paste the config files to another system. This is why we have removed the static location for the certificate. Now the app requests the certificate from [this](https://github.com/zulip/zulip-desktop/pull/822/files#diff-8db7cb157d2ad8cb154d9f20110d7618R52) location.

More info here -
https://chat.zulip.org/#narrow/stream/16-desktop/topic/Enterprise.20Deployment

akashnimare:
@kanishk98 @muskankhedia thanks a lot for testing this.

vaskokj:
@akashnimare 

While I see whats happening, I don't quite see how this helps as the configuration information is still being loaded and managed from a `/Users/USERNAME` location.

Think of an example where you have a single machine that has five users. Two of the users have never logged into the machine which means that the users don't have a $HOME directory. If I install Zulip, I have to force move the certificate and its configuration into each of those users $HOME directory. The other two users that have never logged into the machine wouldn't have a $HOME directory until they do login which means I couldn't add anything to their $HOME directory because it doesn't exist yet. 

OR each user has to run a script to copy and put that information for Zulip into their $HOME directory, which means they would need to know about the script in the first place to run, even though Zulip is already installed. 

A better solution would be to 

1) Easiest but doable, read certs from a centralized location like your "enterprise" config location (e.g. `/etc/zulip-desktop-config`) 

2) (even better) Read from a centralized cert store location for each system such as  `/etc/ssl/certs/ca-certificates.crt` on Ubuntu. Each OS has its own centralized cert store location and code would be needed to handle them explicitly but would be the most robust solution, and generally follows an accepted pattern/solution.

Doing this would resolve anyone from having to "copy-paste" things around on the system.

```
"/etc/ssl/certs/ca-certificates.crt",                // Debian/Ubuntu/Gentoo etc.
"/etc/pki/tls/certs/ca-bundle.crt",                  // Fedora/RHEL 6
"/etc/ssl/ca-bundle.pem",                            // OpenSUSE
"/etc/pki/tls/cacert.pem",                           // OpenELEC
"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", // CentOS/RHEL 7
```

I'm not a web developer but I'll see if I can find some time to write something up.

timabbott:
I agree with @vaskokj that we should support reading from the OS-standard centralized certificate location.  That makes it convenient for a system administrator to just install their CA in the system certificate store and have all apps, including Zulip, be able verify servers using that CA correctly.

akashnimare:
> A better solution would be to read certs from a centralized location your "enterprise" config location (e.g. /etc/zulip-desktop-config) 

@vaskokj I'm not sure if you read this but this is what I said [here](https://chat.zulip.org/#narrow/stream/2-general/topic/desktop.20client.20for.20multiple.20user/near/794051). This will fix the issue you're having, right?

> While I see whats happening, I don't quite see how this helps as the configuration information is still being loaded and managed from a /Users/USERNAME location.

This definitely improves the current certificate location issue. If you have used the app with a certificate you might have noticed this in the certificate.json file -
```json
{
    "https%3A%2F%2Fabhigyank.zulipdev.org": "/Users/yourusername/Library/Application Support/bcd/certificates/my.crt",
}
```

You can see there is a username field here which should not be there in the first place. So we have updated this to -

```

    "https%3A%2F%2Fabhigyank.zulipdev.org": "my.crt"
```

vaskokj:
> > A better solution would be to read certs from a centralized location your "enterprise" config location (e.g. /etc/zulip-desktop-config)
> 
> @vaskokj I'm not sure if you read this but this is what I said [here](https://chat.zulip.org/#narrow/stream/2-general/topic/desktop.20client.20for.20multiple.20user/near/794051). This will fix the issue you're having, right?

If you are referring to the item https://github.com/zulip/zulip-desktop/pull/822 in that comment, then no, I do not feel it will resolve this issue. The current implementation and the changes in 822 are still dependent on information being pulled from the $HOME directory of the user running the app. 

Think of it this way....you image 1000 machines with "standard" set of software on the system. Those 1000 machines you have _no idea_ what user will get those computers. It could be Bob, it could be Sally, it could. You don't know who will get those machines. Someone will walk in, grab a machine and expect it to work. 

In the current solution and in 822, how can I know ahead of time where to copy those files to that users $HOME directory? They don't exist yet. Sure, in 822 I can use the same files and copy them over and over, but it doesn't resolve the problem that when a user pulls a machine off the shelf, signs into it, the very first thing they have to do is go and find some files to copy to their home directory to make zulip work.

Doing what I'm suggesting in my 1) suggestion is to simply make a global configuration file of certs that zulip would pull from (e.g. `/opt/Zulip/certification_files`). Just like in your $HOME directory, but globally to the system. Now with that change, I can simply image the 1000 machines, install Zulip and copy our certification or modify some configuration file to tell it where on the system our companies certificate is. Now when _any_ user pulls a machine off the shelf, they launch Zulip, it reads from a globally configured certification location which was/is already pre-populated for them and BAM Zulip automatically launches with our Zulip servers certificate and our zulip server pops up automatically prompting them to login. There is no fuss, there is no extra work, no copying files around, telling a user how to do something, it "just works". 

The number 2) option I mentioned above would take the solution one step further. What was described in 1) is basically making a Zulip cert store.  The number 2) option is using the available the system wide cert store that every single OS already has. Every OS already maintains a list of approved certificates that tools like wget, firefox, chrome, curl, git interface to and use to check certificates against that system admins can add and remove from as needed. Zulip would just need to interface to it.

Either 1) or 2) would work. 1) would be easier (I haven't look at the code but I would assume), 2) would probably be more difficult because you would have to interface with every single OS's certificate store, but this would be the most optimal, robust, and functionally secure method.  

> 
> > While I see whats happening, I don't quite see how this helps as the configuration information is still being loaded and managed from a /Users/USERNAME location.
> 
> This definitely improves the current certificate location issue. If you have used the app with a certificate you might have noticed this in the certificate.json file -
> 
> ```json
> {
>     "https%3A%2F%2Fabhigyank.zulipdev.org": "/Users/yourusername/Library/Application Support/bcd/certificates/my.crt",
> }
> ```
> 
> You can see there is a username field here which should not be there in the first place. So we have updated this to -
> 
> ```
> 
>     "https%3A%2F%2Fabhigyank.zulipdev.org": "my.crt"
> ```

Doing the solution I described above in option 1) and 2) with the "enterprise" config solution, I feel would make this solution 822 unnecessary in my eyes as then someone wouldn't even need to upload a cert at all (if you implemented 2)) because you could then just add it to the system-cert store. And if you did 1) only _one_ person would have to upload a certificate (or copy/modify some config), then Zulip would pick it up for _every_ user that used Zulip from that point on.

vaskokj:
ok I just tested this (on Ubuntu).

I uploaded the cert from `/etc/ssl/certs/ca-certificates.crt` and Zulip seemed to work properly. 

Which means that instead of reading from a `/home/<USER>/.config/Zulip/certificates/<some cert>` and you just read the certs from these locations you would be 99% of the way there (on Linux at least) on using the system wide cert store. Doing this requires the user to do _nothing_, no uploading certs, storing certs in home directories etc. You are using a system wide common practice at that point. 

You can keep the same functionality you have and allow people to upload certs if you want, but checking the below locations first will remove _any_ need to even upload custom certs if someone adds them to the system wide cert store. This is exactly how git, curl, wget and a pile of other tools read their certs. 
 
```
"/etc/ssl/certs/ca-certificates.crt",                // Debian/Ubuntu/Gentoo etc.
"/etc/pki/tls/certs/ca-bundle.crt",                  // Fedora/RHEL 6
"/etc/ssl/ca-bundle.pem",                            // OpenSUSE
"/etc/pki/tls/cacert.pem",                           // OpenELEC
"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", // CentOS/RHEL 7
```

akashnimare:
@vaskokj putting the certificates to `/etc/zulip-desktop-config/certificates` will work, right?

The structure will look like the following -

```
/etc/zulip-desktop-config/certificates
│   abc.crt
│   zyz.pem

**Comment by vaskokj** [comment]
@akashnimare yes that would suffice for my example for 1). But doing 2) instead looks to be easier than I thought (https://github.com/zulip/zulip-desktop/pull/822#issuecomment-548932688 see what I mentioned here) to implement (at least on linux), less code and would be more robust as you wouldn't have to do any file management. You would simply need to read from those certs to make the http request and that's it.

**Comment by akashnimare** [comment]
> You can keep the same functionality you have and allow people to upload certs if you want, but checking the below locations first will remove any need to even upload custom certs if someone adds them to the system wide cert store. This is exactly how git, curl, wget and a pile of other tools read their certs.

Well, the problem here is we need to specify the certificate (and its location) while requesting to a Zulip server, see [here](https://github.com/zulip/zulip-desktop/blob/17d4d97e2e3ffe5acdbd33a784499f26d63f46a3/app/renderer/js/utils/request-util.ts#L23). Electron doesn't respect the system-wide cert unlike Chrome etc. And this is why we gotta ask the user to upload the cert + the org address in order to access that Zulip server. You can also find some context [here](https://github.com/zulip/zulip-desktop/pull/494).

**Comment by vaskokj** [comment]
> 
> 
> > You can keep the same functionality you have and allow people to upload certs if you want, but checking the below locations first will remove any need to even upload custom certs if someone adds them to the system wide cert store. This is exactly how git, curl, wget and a pile of other tools read their certs.
> 
> Well, the problem here is we need to specify the certificate (and its location) while requesting to a Zulip server, see [here](https://github.com/zulip/zulip-desktop/blob/17d4d97e2e3ffe5acdbd33a784499f26d63f46a3/app/renderer/js/utils/request-util.ts#L23). Electron doesn't respect the system-wide cert unlike Chrome etc. And this is why we gotta ask the user to upload the cert + the org address in order to access that Zulip server. You can also find some context [here](https://github.com/zulip/zulip-desktop/pull/494).

Yeah, I get the problem with Electron. Its actually not an Electron issue but a NodeJS problem. NodeJS in their infinite wisdom hard codes the global certificates into each release. But I digress...

You already have the cert already if you did 1) or 2), its provided by the sysadmin or by the system, respectively. Or are you saying you would need the URL of the zulip server as well to associate it to try with the cert? 

Just use the global sys cert with the request that the user types in. 

Regardless I think you have all of the information you would need to submit the request. You are already doing it, you just are doing it from a $HOME directory instead of globally.

OR 

if you are in an enterprise environment you could just read it from `/etc/zulip-desktop-config` mentioned in this article that you have linked https://github.com/zulip/zulip-desktop/blob/master/docs/Enterprise.md

Example:
```

{
	"presetOrganizations": ["https://chat.zulip.org"],
        "certLocation": "/etc/ssl/certs/ca-certificates.crt"
	"autoUpdate": false
}

```

PR #824, opened 2019-09-25
sidebar: Improve UX for notification settings.

kanishk98:
---

**What's this PR do?**

1. Disables the option for notification settings in the context menu for a server icon if the user is not logged in to that server.
2. If an inactive server's context menu is opened up and its notification settings are requested, we simply switch to that tab first and then show the settings. 

Discussion [here](https://chat.zulip.org/#narrow/stream/16-desktop/topic/disable.20notification.20settings.20option/near/788363).

**Screenshots?**

Disabled context menu item shown below (on the login page).

image

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -731,6 +731,11 @@ class ServerManagerView {
 		this.$dndButton.querySelector('i').textContent = alert ? 'notifications_off' : 'notifications';
 	}
 
+	isLoggedIn(tabIndex: number): boolean {
+		const url = this.tabs[tabIndex].webview.$el.src;
+		return !(url.endsWith('/login/') || this.tabs[tabIndex].webview.loading);
+	}
+
 	addContextMenu($serverImg: HTMLImageElement, index: number): void {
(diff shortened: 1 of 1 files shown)


kanishk98:
Updated PR to also disable the notification settings item if the page is loading.

app/renderer/js/main.ts
```diff
@@ -757,8 +762,11 @@ class ServerManagerView {
 				},
 				{
 					label: 'Notification settings',
+					enabled: this.isLoggedIn(index),
 					click: () => {
-						this.tabs[this.activeTabIndex].webview.showNotificationSettings();
+						// switch to tab whose icon was right-clicked
+						this.activateTab(index);
```
akashnimare, line 768: Do we need this?

kanishk98: Yeah, because I could try clicking on org X's menu while keeping Y active and then Y's notification settings would pop up, which isn't something I really want. 
I think this offers better UX.

PR #828, opened 2019-10-05
Added option to quit on closing the window.

RossBrunton:
This adds an configuration option to quit the app (rather than going to
the tray) when the "close" button is clicked.

---

**What's this PR do?**
Normally, when you click the close button, the app minimises to a tray icon. This behaviour may be undesirable, so this patch adds an option in the settings to cause the close button to actually close the app.

Many apps don't minimise to tray, and IMO it should be the default.

**Screenshots?**

Screenshot

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/index.ts
+++ b/app/main/index.ts
@@ -94,6 +94,9 @@ function createMainWindow(): Electron.BrowserWindow {
 
 	// Keep the app running in background on close event
 	win.on('close', e => {
+		if (ConfigUtil.getConfigItem("quitOnClose")) {
+			app.quit();
+		}
 		if (!isQuitting) {
 			e.preventDefault();
 
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -58,6 +58,7 @@ interface SettingsOptions {
 	};
 	downloadsPath: string;
 	showDownloadFolder: boolean;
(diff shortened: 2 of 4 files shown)


akashnimare:
I think in past we received the opposite feedback which is quitting to tray when a user clicks on `x`. @timabbott is this a reasonable feature?

timabbott:
I think quitting to tray is definitely the right default, in that it's how other similar apps work and quitting completely prevents the user from getting notifications.  But I think it's reasonable to add an option to toggle that behavior for users who want something different.

akashnimare:
For other apps -
* macOS - The app quits to the tray when a user clicks on `x`
* Windows/Linux - It varies so some apps completely quit others quit to tray.

So I think making it configurable makes sense.

muskankhedia:
@akashnimare I tested it in my system(OS: Fedora 30) and it's working completely fine with my system and the quit-on-close option in the settings menu works as expected.

kanishk98 (approved):
Works well for me on Windows 10. :)

akashnimare:
@roberthoenig thanks for working on this :)
@kanishk98, @muskankhedia thanks for testing 🚀

PR #829, opened 2019-10-05
Added option to select download locations.

RossBrunton:
Added an option that, when enabled, will mean any file downloads that
would normally go to ~/Downloads (or wherever), in fact prompt.

---

**What's this PR do?**

Adds an option, when enabled clicking a file that will be downloaded to ~/Downloads (or wherever), it instead displays a "Save" prompt, allowing the user to select where to save it.

I find the default behaviour very annoying. :C

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/index.ts
+++ b/app/main/index.ts
@@ -8,7 +8,7 @@ import path = require('path');
 import fs = require('fs');
 import isDev = require('electron-is-dev');
 import electron = require('electron');
-const { app, ipcMain, session } = electron;
+const { app, ipcMain, session, dialog } = electron;
 
 import AppMenu = require('./menu');
 import BadgeSettings = require('../renderer/js/pages/preference/badge-settings');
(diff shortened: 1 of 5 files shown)


akashnimare:
@RossBrunton hey, thanks for working on this. Can you post a video/gif of the feature you added?

RossBrunton:
Thanks for looking at it, I can't really record video with my system at the moment, but basically when you click a link that would download a file, it shows the system file picker, like:

App

It's configurable in the options as well:

Config

app/main/index.ts
```diff
@@ -294,24 +294,39 @@ app.on('ready', () => {
...
-			};
-
-			// Update the name and path of the file if it already exists
-
-			const updatedFilePath = path.join(downloadPath, formatFile(filePath));
-
-			const setFilePath = fs.existsSync(filePath) ? updatedFilePath : filePath;
+			let setFilePath: string;
+			let shortFileName: string;
+			if (ConfigUtil.getConfigItem("promptDownload", false)) {
```
kanishk98: @akashnimare I just noticed that the `quotes` field for `xo` overrides is set to 0. I'm guessing it's been done so we can use escape strings as well, but it might lead to inconsistency when writing strings. Do you think it's a good idea to set up a post-commit prettier hook?

app/translations/en.json
```diff
@@ -113,5 +113,6 @@
 	"Zoom Out": "Zoom Out",
 	"Zulip Help": "Zulip Help",
 	"keyboard shortcuts": "keyboard shortcuts",
-	"script": "script"
+	"script": "script",
```
kanishk98: @RossBrunton I think you staged this file by accident. Would you mind unstaging it?

app/main/index.ts
```diff
@@ -294,24 +294,39 @@ app.on('ready', () => {
...
+				const showDialogOptions: object = {
+					defaultPath: path.join(downloadPath, item.getFilename())
+				};
+
+				setFilePath = dialog.showSaveDialog(mainWindow, showDialogOptions);
```
kanishk98, line 307: I'm not sure why, but this doesn't seem to run immediately on my Windows laptop. I ended up clicking on the same link 5-6 times before I got the save dialog. 
Would it be a good idea to use `showSaveDialogSync` instead?

kanishk98 (changes requested):
Thanks for working on this, @RossBrunton! I've got a couple of problems though:

This didn't work for me as expected on Windows, and I've included a few suggestions in the comments. 
Also, please fix the merge conflicts by rebasing against master.

Thanks!

app/main/index.ts
RossBrunton: Looks like the version of Electron you are using has saveDialog be the synchronous one. In some later version it seems to have been renamed to saveDialogSync and then a new saveDialog added.

vsvipul (approved):
LGTM! @akashnimare  I'm not sure about the UI of the new setting added though. Any suggestions for that?

akashnimare:
Let me check this again.

akashnimare:
@timabbott do you think we need this option? I doubt it since we already have the following option by which the user can configure the directory where they want to save the downloaded file -

image


Also, I don't think people change this frequently.

RossBrunton:
I'd use it frequently. Either to give a filename to the file, or to download it to an appropriate location. Currently, I have to download the file, go to the Downloads folder, and then move and rename it.

Most applications have an option for manual download locations or prompt by default, while Zulip is the only application I'm aware of which does not support this feature.

timabbott:
@RossBrunton can you point to other similar applications?  It'd be helpful at least for looking at how they name/describe the options.

@akashnimare I think this is a fine option to provide if it's not expensive to maintain; it seems potentially useful for users who care and not a big source of clutter.  We should just figure out how to implement it in a way that's low-cost for us (e.g. by using the same strings Firefox uses to describe it so it's clear what it does).

akashnimare:
Oh wait, I just remember that you can get the download dialog using `option` key + click on the file. This works well in macOS. 

image

@vsvipul can you check if a similar thing exists on Windows. If yes, then we might not need this feature at all (though we need to document it somewhere).

RossBrunton:
The other applications that provide this option are Chrome and Firefox, which present it as follows:

Downloads1
Downloads2

I've tested it under Linux, and CTRL+Clicking on a link just downloads it as normal without prompting.

Out of interest, are there any other applications that behave like Zulip in this manner? I'm honestly not sure how widespread Zulip's behaviour is.

akashnimare:
> I've tested it under Linux, and CTRL+Clicking on a link just downloads it as normal without prompting.

We should go with this option if it's not consistent across all the OSes.

akashnimare:
Works well for me on macOS. @vsvipul GTG?

vsvipul:
@akashnimare  Alt+click asks for download location for me on linux.

RossBrunton:
> @akashnimare Alt+click asks for download location for me on linux.

On my system (an Ubuntu+Mint hybrid), alt+click causes the window to be grabbed for dragging.

akashnimare:
Yeah, `Alt` behaves a bit differently on some distros. But I think a lot of folks might not aware with this little shortcut so it's best to have the setting for this.

vsvipul:
> On my system (an Ubuntu+Mint hybrid), alt+click causes the window to be grabbed for dragging.

Well in that case, I think we can add it. GTG.

akashnimare:
Cool. Merging. Thanks, everyone!

PR #830, opened 2019-10-07
Upgrade all dependencies (except electron); linting

SamuelMarks:
**What's this PR do?**
Upgrades dependencies

Maybe I'll upgrade `electron` in a future PR, but regardless this is a step in the right direction

**Screenshots?**

  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS


akashnimare:
@kanishk98 can you review?

kanishk98 (approved):
The PR works well for me, although I was unable to build an executable on Windows owing to the `electron-notarize` package (I wanted to test the auto-update as well). 

I don't think that's related to the changes in this PR, though. @akashnimare can you try building an executable for Windows on your end and let me know on czo if it works well for you?

akashnimare:
@SamuelMarks can you please explain the changes you made (why it's better than the current one), just leave the comments on the line.

SamuelMarks:
Was having some issues with performance, and upgrading dependencies seemed like a good first step.

A good next step would be upgrading to latest electron, which will likely involve rescaffolding.

app/renderer/js/pages/preference/new-server-form.ts
```diff
@@ -92,9 +92,8 @@ class NewServerForm extends BaseComponent {
 			this.submitFormHandler();
 		});
 		this.$newServerUrl.addEventListener('keypress', event => {
-			const EnterkeyCode = event.keyCode;
 			// Submit form when Enter key is pressed
-			if (EnterkeyCode === 13) {
```
akashnimare, line 97: @SamuelMarks this looks good to me?

package.json
```diff
@@ -146,32 +146,31 @@
...
+    "assert": "2.0.0",
+    "cp-file": "7.0.0",
     "devtron": "1.4.0",
     "electron": "3.1.10",
-    "electron-builder": "20.43.0",
-    "electron-connect": "0.6.2",
-    "electron-debug": "1.4.0",
-    "electron-notarize": "0.2.0",
-    "eslint-config-xo-typescript": "0.14.0",
+    "electron-builder": "21.2.0",
```
akashnimare, line 156: @SamuelMarks please revert this. We don't upgrade electron-builder like this since there is heavy testing involve in this.

timabbott:
Versions of essentially all of these changes were integrated when we did the big dependency upgrade this weekend.

app/renderer/js/pages/preference/new-server-form.ts
SamuelMarks: See the deprecation notice https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

andersk: This was already fixed in 909e0f07e3f7439b0174b9a35bb72a1506bad0e2.

PR #845, opened 2019-11-29
Fix size of macOS dock icon.

bdon:
**What's this PR do?**

Replaces the 512x512 image asset in the ICNS with a new image. The new image is the original 512x512 image scaled to 450x450 and then expanded to 512x512 with transparent padding. 

https://github.com/zulip/zulip-desktop/issues/787
I only replaced the 512x512 image asset which seems to fix the dock, even when the dock is resized. Not sure if the 256x256 and smaller assets should also be scaled or if there's some build script to create the icns.

**Screenshots?**

Before:
icon_before

After:
icon_after

  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [x] macOS


akashnimare:
Adding this in the release. Let's see if it makes any difference.

PR #781, opened 2019-06-28
[WIP] Upgrade Electron version to 5.0.6

kanishk98:
---

**What's this PR do?**

Makes a couple Electron version leaps to 5.0.6.

**Any background context you want to provide?**

Needed to enable `webviewTag` when creating a `BrowserWindow` in the main process.
Still have to assure ourselves that things haven't broken in the upgrade. 

**You have tested this PR on:**
  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

PR #850, opened 2019-12-09
Urgent electron upgrade

kanishk98:
---

**What's this PR do?**

Upgrades the Electron version to 5.0.6 to deal with Google login issues some users have been facing. 

I chose this version because this had been well-tested in an earlier PR (#781). 
@akashnimare I tried testing the auto-update, but ran into the same issue where Windows was unable to locate `Zulip.exe`. Please let me know if there's an alternative solution. 

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/autoupdater.ts
+++ b/app/main/autoupdater.ts
@@ -71,7 +71,8 @@ export function appUpdater(updateFromMenu = false): void {
 				message: messageText,
 				detail: (error).toString() + `\n\nThe latest version of Zulip Desktop is available at -\nhttps://zulipchat.com/apps/.\n
 Current Version: ${app.getVersion()}`
-			}, response => {
+			}).then(res => {
+				const { response } = res;
 				if (response === 0) {
 					shell.openExternal('https://zulipchat.com/apps/');
 				}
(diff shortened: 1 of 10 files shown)


akashnimare:
@kanishk98  Great, let me check. How's the builder + updater behaving on this version?

kanishk98:
It's throwing errors when I run `npm run dist`. I'm still looking into whether this is because of my Windows `npm` setup or something else. 

Please note, though - this still is just an update. I think we'll need to rebase and merge #D831 to fully fix the GitHub and Google login issues. Working on that right now.

akashnimare:
Can't update the error is still there -

image

kanishk98:
Updated to 6.1.4, but looks like the error still persists. Working on 7.1.2 right now.

timabbott:
We should also be testing with the Google issue; that GitHub warning at least looks like one could ignore it if one wanted to.  It's possible that an Electron upgraded is needed for Google but something else or in addition for GitHub.

andersk:
Similar to the Google issue, we should delegate to the real system browser for the GitHub OAuth flow.

akashnimare:
> Similar to the Google issue, we should delegate to the real system browser for the GitHub OAuth flow.

Yeah, that is the plan.

timabbott:
We merged the upgrade to Electron v6 today, which incorporated almost all of these changes.  Thanks for your work on this @kanishk98!

PR #854, opened 2019-12-14
[WIP] Implement social provider logins via OAuth

kanishk98:
---

**What's this PR do?**

Allows users to log in to Google/GitHub via OAuth by delegating the login process to the system browser instead of handling it in the `<webview>`. Done in accordance with Google's recommendations for login security. 

**Why is this still WIP?**

I'm facing some issues with getting back the access token - specifically, the error `setImmediate() is not defined`. Looking into this right now, but I'd love some help too. Currently able to delegate login to the browser and get a callback from there into the app. 

  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/preload.ts
+++ b/app/renderer/js/preload.ts
@@ -11,6 +11,7 @@ import SetupSpellChecker from './spellchecker';
 import isDev = require('electron-is-dev');
 import LinkUtil = require('./utils/link-util');
 import params = require('./utils/params-util');
+import AppAuthUtil = require('./utils/appauth-util');
 
 import NetworkError = require('./pages/network');
 
@@ -79,7 +80,21 @@ process.once('loaded', (): void => {
 // To prevent failing this script on linux we need to load it after the document loaded
 document.addEventListener('DOMContentLoaded', (): void => {
 	if (params.isPageParams()) {
-	// Get the default language of the server
+		const $signInButtons = document.querySelectorAll('.login-social');
+		for (const $signInButton of $signInButtons) {
+			$signInButton.addEventListener('click', event => {
+				event.preventDefault();
+				const appAuthUtil = new AppAuthUtil();
+				appAuthUtil.fetchServiceConfiguration().then(res => {
+					return appAuthUtil.makeAuthorizationRequest();
+				});
+				appAuthUtil.authStateEmitter.on('on_token_response', event => {
+					console.log(event);
+				});
+			});
+		}
+
+		// Get the default language of the server
 		const serverLanguage = page_params.default_language; // eslint-disable-line no-undef
 		if (serverLanguage) {
 			// Init spellchecker
--- a/app/renderer/js/preload.ts
+++ b/app/renderer/js/preload.ts
@@ -148,7 +148,8 @@ window.addEventListener('load', (event: any): void => {
 
 // electron's globalShortcut can cause unexpected results
 // so adding the reload shortcut in the old-school way
-// Zoom from numpad keys is not supported by electron, so adding it through listeners.
+// electron does not support adding multiple accelerators for a menu item
+// so adding numpad shortcuts here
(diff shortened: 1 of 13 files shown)


app/renderer/js/preload.ts
```diff
@@ -79,7 +80,21 @@ process.once('loaded', (): void => {
 // To prevent failing this script on linux we need to load it after the document loaded
 document.addEventListener('DOMContentLoaded', (): void => {
 	if (params.isPageParams()) {
-	// Get the default language of the server
+		const $signInButtons = document.querySelectorAll('.login-social');
```
timabbott, line 83: If we go with this approach, it's likely that we'll want to agree on a standard API for this discovery, and add a comment in the webapp codebase forbidding changing the name of the CSS class, so this doesn't do something crazy.  

(I think we also want `.signup-social` for the registration page?)

kanishk98: Yeah, we should add that comment too. 
I actually dislike this approach myself, but I figured this would be a quick fix since listening for URL changes causes 2 tabs to open up (since both URLs that Google redirects us through are under the same domain too), and that messes up the local webserver listening for an auth confirmation. 

I'll have to check out the registration page thing, didn't think about that. Thanks!

mateuszmandera: Perhaps this could take adventage of ``external_authentication_methods`` in ``/server_settings`` (https://chat.zulip.org/api/server-settings) like the mobile app to get the urls by which to locate the buttons, instead of hardcoding selection by ``.login-social``? Especially since we probably want to change that name, as we've recently merged changes enabling us to render those button for backends that we won't want to call "social". RemoteUser SSO is already included in that.

mateuszmandera: Besides RemoteUser SSO, also SAML is a social backend, but it's not OAuth-based, do these changes account for that?

kanishk98: > Perhaps this could take adventage of external_authentication_methods in /server_settings

That's a good idea, I'll have a look at the mobile app's code and change things here. 

> also SAML is a social backend, but it's not OAuth-based, do these changes account for that?

No, I didn't consider SAML when working on this. The problem that we set out to address was mostly with Google logins (since Google has completely blocked all in-app login attempts), so I just jumped to implement an OAuth mechanism to replace our `<webview>` stuff.

mateuszmandera: Ah, while it seems like a reasonable thing to do in the future, we should be fine without handling SAML for now. The reason for my question was to make sure the app will know not to apply the OAuth logic to SAML (since it will find SAML in the list ``$signInButtons = document.querySelectorAll('.login-social');``).

kanishk98: Yeah, that's a good point, thanks! :)
I think your suggestion of using the `/server_settings` API is great. Once we figure out how to solve this pesky 400 error on the app's end, I'll change the way we intercept requests. That way, OAuth logins won't interfere with SAML ones.

app/renderer/js/preload.ts
```diff
@@ -133,7 +148,8 @@ window.addEventListener('load', (event: any): void => {
 
 // electron's globalShortcut can cause unexpected results
 // so adding the reload shortcut in the old-school way
-// Zoom from numpad keys is not supported by electron, so adding it through listeners.
+// electron does not support adding multiple accelerators for a menu item
```
vsvipul, line 151: Why is this change a part of this ?

vsvipul:
Why is the electron version update required for this ?

vsvipul:
This PR fails to work for me. @akashnimare  Does it work for you ?

akashnimare:
Yeah, getting some typescript errors.

kanishk98:
Closing this PR in favour of #857.

PR #857, opened 2020-01-06
[WIP] Session ID-based login for OAuth providers

kanishk98:
---

**What's this PR do?**

Generates an OTP, similar to the mobile app flow, and then XORs the same with the encrypted API key returned by the server. Picks up this URL using the `zulip://` scheme. 

I'm working on writing an API endpoint for generating the session ID for a logged in user (will get back to czo about this tomorrow morning). 

@akashnimare @vsvipul please have a look at this whenever you find time. The patched version of the server you'll need to test this app against can be found [here](https://github.com/kanishk98/zulip/tree/desktop-oauth). 

  - [x] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/index.ts
+++ b/app/main/index.ts
@@ -38,13 +38,17 @@ const mainURL = 'file://' + path.join(__dirname, '../renderer', 'main.html');
 
 const singleInstanceLock = app.requestSingleInstanceLock();
 if (singleInstanceLock) {
-	app.on('second-instance', () => {
+	app.on('second-instance', (event, argv) => {
 		if (mainWindow) {
 			if (mainWindow.isMinimized()) {
 				mainWindow.restore();
 			}
 
 			mainWindow.show();
+			if (process.platform !== 'darwin') {
+				const deepLinkingUrl = argv.slice(1);
+				handleDeepLink(deepLinkingUrl[(isDev && process.platform === 'win32') ? 1 : 0]);
+			}
 		}
 	});
 } else {
@@ -65,6 +69,10 @@ const toggleApp = (): any => {
 		mainWindow.hide();
 	}
 };
+function handleDeepLink(url: string): void {
+	mainWindow.webContents.focus();
+	mainWindow.webContents.send('deep-linking-url', url);
+}
 
 function createMainWindow(): Electron.BrowserWindow {
 	// Load the previous state with fallback to defaults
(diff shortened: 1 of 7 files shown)


app/renderer/js/utils/auth-util.ts
```diff
@@ -0,0 +1,29 @@
...
+		}
+		ConfigUtil.setConfigItem('desktopOtp', otp);
+		shell.openExternal(`${domain}accounts/login/?desktop_flow_otp=${otp}`);
+	};
+
+	xorStrings = (a: string, b: string): string => {
+		let res = '';
+		let i = a.length;
+		let j = b.length;
+		while (i-- > 0 && j-- > 0) {
```
vsvipul, line 22: "> 0" is redundant here

app/main/index.ts
```diff
@@ -38,13 +38,17 @@ const mainURL = 'file://' + path.join(__dirname, '../renderer', 'main.html');
...
+	app.on('second-instance', (event, argv) => {
 		if (mainWindow) {
 			if (mainWindow.isMinimized()) {
 				mainWindow.restore();
 			}
 
 			mainWindow.show();
+			if (process.platform !== 'darwin') {
+				const deepLinkingUrl = argv.slice(1);
+				handleDeepLink(deepLinkingUrl[(isDev && process.platform === 'win32') ? 1 : 0]);
```
vsvipul, line 50: Might me good to mention that this function is for linux and windows (same as you have mentioned for macos below).

akashnimare:
@kanishk98 what's the expected behaviour here?

vsvipul:
@akashnimare I tested this PR after getting the keys from github. Login works fine for me, but everything works in the app. I think we still need to shift the workflow to browser for linux using zulip:// scheme. Is this the case @kanishk98  ? 
In that case I am working on top of this to integrate that as well.

kanishk98:
Yeah, the workflow does shift to the browser - because this is a WIP version, I haven't written the interception of the login process into the app (see [this comment](https://github.com/zulip/zulip-desktop/pull/854#discussion_r357955978) for reasons about why that's a little more complex). 

For the login link to the browser, could you try deleting and re-adding an organisation? That's what I tested with on my end and it seemed to work well. You can also right-click on the server tab in the sidebar and click on 'Login with OAuth' (horrible UX, I know, but it helps us test things quickly). 

@akashnimare I'm trying to make sure that we get a valid API key back from the Zulip server. If that's the case, we can use `zulip-js` to make a request to the Zulip API for a session ID and then log in the user.

akashnimare:
@kanishk98 thanks for working on this. Closing this since we have cherry-picked this PR's commit in #863.

PR #858, opened 2020-01-07
Decode server name in Window menu.

vsvipul:
**What's this PR do?**
We are decoding the name in the webview but are currently not doing so in the Window menu. I hope this is what we want @akashnimare . 

**Screenshots?**
Earlier
Screenshot from 2020-01-08 00-19-33

Now
Screenshot from 2020-01-08 00-21-05

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS


akashnimare:
We might have missed this. Thanks, @vsvipul, merged.

PR #859, opened 2020-01-10
startup: Use inbuilt electron API for autostartup.

### LINKED ISSUE #851: Zulip not starting up in Windows
---

- **Operating System**:
  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS
Simply install latesst version v4.0.0, ensure in settings startup at login enabled  and restart
The zulip app is not started on startup (not on main desktop nor in system tray)

vsvipul:
**What's this PR do?**
Switch from using external auto-launch module to
inbuilt setLoginItemSettings for windows and macos,
as some users reported issues on windows.
Fixes #851.

@akashnimare @kanishk98  Need to test this on windows and macos. Can you do it ?

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/main/startup.ts
+++ b/app/main/startup.ts
@@ -11,21 +11,23 @@ export const setAutoLaunch = (AutoLaunchValue: boolean): void => {
 		return;
 	}
 
-	// On Mac, work around a bug in auto-launch where it opens a Terminal window
-	// See https://github.com/Teamwork/node-auto-launch/issues/28#issuecomment-222194437
-
-	const appPath = process.platform === 'darwin' ? app.getPath('exe').replace(/\.app\/Content.*/, '.app') : undefined; // Use the default
-
-	const ZulipAutoLauncher = new AutoLaunch({
-		name: 'Zulip',
-		path: appPath,
-		isHidden: false
-	});
 	const autoLaunchOption = ConfigUtil.getConfigItem('startAtLogin', AutoLaunchValue);
 
-	if (autoLaunchOption) {
-		ZulipAutoLauncher.enable();
+	// setLoginItemSettings doesn't support linux
+	if (process.platform === 'linux') {
+		const ZulipAutoLauncher = new AutoLaunch({
+			name: 'Zulip',
+			isHidden: false
+		});
+		if (autoLaunchOption) {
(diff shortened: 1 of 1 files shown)


app/main/startup.ts
```diff
@@ -11,21 +11,23 @@ export const setAutoLaunch = (AutoLaunchValue: boolean): void => {
...
-		isHidden: false
-	});
 	const autoLaunchOption = ConfigUtil.getConfigItem('startAtLogin', AutoLaunchValue);
 
-	if (autoLaunchOption) {
-		ZulipAutoLauncher.enable();
+	// setLoginItemSettings doesn't support linux
+	if (process.platform === 'linux') {
+		const ZulipAutoLauncher = new AutoLaunch({
+			name: 'Zulip',
```
akashnimare, line 19: @vsvipul you removed the path arg here.

akashnimare (changes requested):

akashnimare:
Left a comment. Otherwise all good.

akashnimare:
@vsvipul I can merge this real quick if you can make those changes.

app/main/startup.ts
vsvipul: @akashnimare  We were changing the path only for mac previously and using the default for linux. As we now use AutoLaunch for only linux, this is no longer required. See https://www.npmjs.com/package/auto-launch .
```
For NW.js and Electron apps, you don't have to specify the path. We guess based on process.execPath.
```

vsvipul:
@akashnimare I have addressed the issue. Before merging, please test this on mac and windows too as I haven't done so.

akashnimare:
Works well on macOS.

app/main/startup.ts
akashnimare: Got it. But do we need to update the auto-launch version?

vsvipul: We are already on the latest version i.e. 5.0.5 .

parthi2929:
so current change is failing in windows? @akashnimare

akashnimare:
Tested on Windows, works as expected. Merging. @parthi2929 this will be added in the next release.

PR #863, opened 2020-01-16
auth: Move social login process to browser.

vsvipul:
Moves the login to browser since there
was no way to verify the authencity of the
auth process for a custom server and to
prevent phishing attacks.

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/utils/auth-util.ts
+++ b/app/renderer/js/utils/auth-util.ts
@@ -1,0 +1,36 @@
+import { remote } from 'electron';
+
+import cryptoRandomString = require('crypto-random-string');
+import ConfigUtil = require('./config-util');
+
+const { shell } = remote;
+
+class AuthUtil {
+	openInBrowser = (link: string) => {
+		const otp = cryptoRandomString({length: 64});
+		ConfigUtil.setConfigItem('desktopOtp', otp);
+		shell.openExternal(`${link}?desktop_flow_otp=${otp}`);
+	};
+
+	xorStrings = (a: string, b: string): string => {
+		if (a.length === b.length) {
+			return a
+				.split('')
+				.map((char, i) => (parseInt(a[i], 16) ^ parseInt(b[i], 16)).toString(16))
+				.join('')
+				.toUpperCase();
+		} else {
+			return '';
+		}
+	};
+
+	hexToAscii = (hex: string) => {
+		let ascii = '';
+		for (let i = 0; i < hex.length; i += 2) {
+			ascii += String.fromCharCode(parseInt(hex.substr(i, 2), 16));
+		}
+		return ascii;
+	};
+}
+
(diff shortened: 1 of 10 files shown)


app/renderer/js/utils/auth-util.ts
```diff
@@ -0,0 +1,39 @@
+import { remote } from 'electron';
+
+import randomstring = require('randomstring');
+import ConfigUtil = require('./config-util');
+
+const { shell } = remote;
+
+class AuthUtil {
+	openInBrowser = (link: string) => {
+		const otp = randomstring.generate({
```
timabbott: We should be sure that we're using a cryptographically secure random string here.  We should look at whether e.g.  https://www.npmjs.com/package/crypto-random-string does anything different from the `randomstring` module.  Though probably in any case, I'd prefer to use something that explicitly documents itself as for cryptographic use.

vsvipul: This module does the same thing that ```randomstring``` does. Will still change it.

akashnimare:
@vsvipul you need to drop the last commit I think?

vsvipul:
@akashnimare  @timabbott  I have updated this PR to use browser login only when we have authMethods in page_params. Please check if I have done it correctly or some change is required in that case.

Also, added an option to toggle login from browser (default- false), in case some user faces problem while logging in from browser, he always has the default electron login option as a backup.

Once the signup part of zulip/zulip is done, i think we can merge this after testing.

timabbott:
Where does that option exist?  I'm not sure it really makes sense to provide it given that for most social auth provides, direct login won't work anymore anyway.

vsvipul:
@timabbott  It exists in app settings under 'Advanced Settings'. Can you cite the source where it says that direct login will not be supported by electron webview ? If it is actually tha case, i'll remove the setting.

akashnimare:
I don't remember the app but I have seen folks supporting both the options i.e. login via web browser and in-built login using a new electron window.

timabbott:
@vsvipul both Google and GitHub have moved to block direct login via Electron apps; that's why we're doing this transition.  Since those are the primary social auth methods we support, it seems like we should just go through the browser everywhere.  But I suppose if it's an easy advanced setting to add, we can include it.

vsvipul:
@akashnimare  Since webapp side is complete, can you test and merge this ?

akashnimare:
Awesome, will do.

akashnimare:
Merged. Thanks everyone ❤️

andersk:
This failed to work in the AppImage for Linux. Therefore, the security team made a decision to revert this temporarily to avoid blocking the v5.0.0 security release. We’ll be discussing the right way to move forward on this in the coming days.

PR #864, opened 2020-01-29
deps: Update electron, builder, updater and spellchecker.

vsvipul:
Update electron version to v6 and builder,
updater and spellchecker to latest versions.

Still need to test working of autoupdate. I have fixed other breaking changes. Please check if something is left.

  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS


akashnimare:
@vsvipul can focus on this once we merge the oAuth PR + push a new release?

timabbott:
@andersk and I cleaned this up and merged it as 107e5229141879b69225153dc9e6d15f54c8c7bf.

Huge thanks @vsvipul for your work on this.

The one thing I asked about is that I tried removing that `initSettings` change and the app and settings pages seemed to work fine, so I merged this without that change.  What was the failure mode you saw without it?  It's possible we still need to do that.

vsvipul:
@timabbott That change was required because without it, I was getting an error in the developer console. Apparently, we need to have the webview loaded before calling any of it's events, so there was a race condition there. So, I tweaked the code to load the settings view when the app starts itself. This prevented that race condition.

As mentioned in the comment - // This is required to attach webview to DOM so that dom-ready event is emitted before any setting method is called.

timabbott:
@vsvipul what's the precise sequence of steps you needed to do in order to see that error, and what's the error?  I'd like to see that to figure out if there's another solution.

(Maybe it makes sense to move this discussion to a new PR containing just that `init_settings` change that we can discuss in review?)

PR #868, opened 2020-02-09
TS TODOs : makes reconnect-util generic, provides type annotations

manavmehta:
**What's this PR do?**
What does this PR do:
    Worked on TypeScript TODOs
	Can pass class into reconnect-util from main.ts (generic class)
	CurrentWindow now has a .send() property
	We now use ProxyRule for the proxy property
	Type annotated some variables

  - [✓ ] macOS (Catalina 10.15.2)

Zulip App v 4.0.3

### DIFF
--- a/app/renderer/js/components/handle-external-link.ts
+++ b/app/renderer/js/components/handle-external-link.ts
@@ -3,13 +3,13 @@ import { ipcRenderer, remote } from 'electron';
 import LinkUtil = require('../utils/link-util');
 import DomainUtil = require('../utils/domain-util');
 import ConfigUtil = require('../utils/config-util');
+import type WebView from './webview';
 
 const { shell, app } = remote;
 
 const dingSound = new Audio('../resources/sounds/ding.ogg');
 
-// TODO: TypeScript - Figure out a way to pass correct type here.
-function handleExternalLink(this: any, event: any): void {
+function handleExternalLink(this: WebView, event: Electron.NewWindowEvent): void {
 	const { url } = event;
 	const domainPrefix = DomainUtil.getDomain(this.props.index).url;
 	const downloadPath = ConfigUtil.getConfigItem('downloadsPath', `${app.getPath('downloads')}`);
--- a/app/renderer/js/notification/helpers.ts
+++ b/app/renderer/js/notification/helpers.ts
@@ -80,9 +80,7 @@ const webContentsId = webContents.id;
 // this function will focus the server that sent
 // the notification. Main function implemented in main.js
 export function focusCurrentServer(): void {
-	// TODO: TypeScript: currentWindow of type BrowserWindow doesn't
-	// have a .send() property per typescript.
-	(currentWindow as any).send('focus-webview-with-id', webContentsId);
+	currentWindow.webContents.send('focus-webview-with-id', webContentsId);
 }
 // this function parses the reply from to notification
 // making it easier to reply from notification eg
--- a/app/renderer/js/pages/preference/general-section.ts
+++ b/app/renderer/js/pages/preference/general-section.ts
@@ -290,9 +290,7 @@ class GeneralSection extends BaseSection {
 				const newValue = !ConfigUtil.getConfigItem('silent', true);
 				ConfigUtil.setConfigItem('silent', newValue);
 				this.updateSilentOption();
-				// TODO: TypeScript: currentWindow of type BrowserWindow doesn't
-				// have a .send() property per typescript.
-				(currentBrowserWindow as any).send('toggle-silent', newValue);
+				currentBrowserWindow.webContents.send('toggle-silent', newValue);
 			}
 		});
 	}
--- a/app/renderer/js/utils/request-util.ts
+++ b/app/renderer/js/utils/request-util.ts
@@ -15,11 +15,9 @@ const logger = new Logger({
 	timestamp: true
 });
 
-// TODO: TypeScript - Use ProxyRule for the proxy property
-// we can do this now since we use export = ProxyUtil syntax
 interface RequestUtilResponse {
 	ca: string;
-	proxy: string | void | object;
+	proxy: string | void | ProxyUtil.ProxyRule;
 	ecdhCurve: 'auto';
 	headers: { 'User-Agent': string };
 	rejectUnauthorized: boolean;
--- a/app/renderer/js/utils/proxy-util.ts
+++ b/app/renderer/js/utils/proxy-util.ts
@@ -4,7 +4,7 @@ import url = require('url');
 
 import ConfigUtil = require('./config-util');
 
-interface ProxyRule {
+export interface ProxyRule {
 	hostname?: string;
 	port?: number;
 }


app/renderer/js/components/handle-external-link.ts
```diff
@@ -75,7 +76,7 @@ function handleExternalLink(this: any, event: any): void {
 						body: 'Download failed'
 					});
 				} else {
-					this.$el.downloadURL(url);
+					this.downloadURL(url);
```
vsvipul: Why was this changed? This will also change the functioning, i guess.

app/renderer/js/components/handle-external-link.ts
```diff
@@ -1,15 +1,16 @@
-import { ipcRenderer, remote } from 'electron';
+import { ipcRenderer, remote, NewWindowEvent } from 'electron';
```
vsvipul: Use Electron.NewWindowEvent instead of importing.

app/renderer/js/notification/helpers.ts
```diff
@@ -82,7 +82,7 @@ const webContentsId = webContents.id;
 export function focusCurrentServer(): void {
 	// TODO: TypeScript: currentWindow of type BrowserWindow doesn't
 	// have a .send() property per typescript.
-	(currentWindow as any).send('focus-webview-with-id', webContentsId);
+	currentWindow.webContents.send('focus-webview-with-id', webContentsId);
```
vsvipul, line 83: Why webcontents is added ? This is wrong.

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -288,7 +288,7 @@ class GeneralSection extends BaseSection {
 				this.updateSilentOption();
 				// TODO: TypeScript: currentWindow of type BrowserWindow doesn't
 				// have a .send() property per typescript.
-				(currentBrowserWindow as any).send('toggle-silent', newValue);
+				currentBrowserWindow.webContents.send('toggle-silent', newValue);
```
vsvipul, line 293: Same here.

app/renderer/js/utils/link-util.ts
```diff
@@ -22,12 +22,12 @@ class LinkUtil {
 	}
 
 	isInternal(currentUrl: string, newUrl: string): IsInternalResponse {
-		const currentDomain = wurl('hostname', currentUrl);
-		const newDomain = wurl('hostname', newUrl);
+		const currentDomain: string = wurl('hostname', currentUrl);
```
vsvipul: Explicitly mentioning type is not required anywhere. Same for all 4 changes below.

vsvipul:
Left some comments. Also changes to package-lock.json are not required to be added to the PR anywhere. Remove those. Also, TODO messages also need to be removed when fixing that todo. I would suggest fixing all TODOs in a single commit, as all of them are minor changes. Also better commit messages and PR names and description will help.

app/renderer/js/utils/link-util.ts
manavmehta: A doubt here actually! When (in VS Code), I hover to these vars (currentDomain and newDomain only), they say :any. So isn't an explicit mention required to avoid the type 'any' ?

app/renderer/js/notification/helpers.ts
manavmehta: BrowserWindow as per electronjs docs, doesn't have any .send() property. It is a property used as BrowserWindow.webContents.send().
That's why I did this.

vsvipul: Test the functionality. If it works this way, then it is fine.

app/renderer/js/utils/link-util.ts
vsvipul: No, not required.

app/renderer/js/components/handle-external-link.ts
manavmehta: Done

app/renderer/js/notification/helpers.ts
manavmehta: Tested. Works Fine.

manavmehta:
Hey @vsvipul I have fixed it according to your review. can you please re-review and merge it ?
cc: @akashnimare

abhigyank:
The completed ToDos look fine to me for this PR. Commit message can be improved while merging.

manavmehta:
@abhigyank I'll keep that in mind and henceforth provide better commit messages

vsvipul (approved):
LGTM!

app/renderer/js/components/handle-external-link.ts
```diff
@@ -9,7 +9,7 @@ const { shell, app } = remote;
 const dingSound = new Audio('../resources/sounds/ding.ogg');
 
 // TODO: TypeScript - Figure out a way to pass correct type here.
-function handleExternalLink(this: any, event: any): void {
+function handleExternalLink(this: any, event: Electron.NewWindowEvent): void {
```
andersk: Should be `this: WebView` (with `import type WebView from './webview'`).

app/renderer/js/utils/reconnect-util.ts
```diff
@@ -11,11 +11,8 @@ const logger = new Logger({
 	timestamp: true
 });
 
-class ReconnectUtil {
-	// TODO: TypeScript - Figure out how to annotate webview
-	// it should be WebView; maybe make it a generic so we can
-	// pass the class from main.ts
-	webview: any;
+class ReconnectUtil<ClassName> {
+	webview: ClassName ;
```
andersk: This generic doesn’t make sense because we do `webview.props.url`. TypeScript doesn’t catch that only because you forgot to change the `webview: any` parameter of the constructor. We should just use `webview: WebView` in both places.

app/renderer/js/utils/request-util.ts
```diff
@@ -17,11 +17,14 @@ const logger = new Logger({
 
 let instance: null | RequestUtil = null;
 
-// TODO: TypeScript - Use ProxyRule for the proxy property
-// we can do this now since we use export = ProxyUtil syntax
+interface ProxyRule {
+	hostname?: string;
+	port?: number;
+}
```
andersk: This is already defined in `proxy-util.ts`; the comment is suggesting we should import it, not redefine it.

andersk:
Updated as suggested and merged. Thanks @manavmehta!

manavmehta:
Hey @andersk I had these changes in mind but I've been encountering some warnings so avoided them purposefully. I've commented all the warnings above replying to each of your comment.
Are these to be ignored ?

app/renderer/js/components/handle-external-link.ts
manavmehta: Cannot find namespace 'WebView'   .ts(2503)

app/renderer/js/utils/reconnect-util.ts
manavmehta: Did I get the TODO wrong as you're correct that `webview.props.url` even after changing the constructor to `webview: ClassName`

P.S. It was asked to make it generic

andersk: You read the TODO correctly, but the TODO did not make sense, and we should do things that make sense. :slightly_smiling_face:

app/renderer/js/components/handle-external-link.ts
andersk: It works fine on `master`. Did you forget to run `npm install` to follow the upgrades we did to TypeScript and other dependencies?

manavmehta: solved

Issue #878, opened 2020-03-01
Enable promise/prefer-await-to-then linter rule


timabbott:
We should convert all of our asynchronous code using `.then()` to use `async/await` instead.

We can find the code to change easily by removing the `package.json` configuration we have to disable the  `promise/prefer-await-to-then linter rule` linter rule.

iyashwantsaini:
@zulipbot claim

zulipbot:
Hello @meyash, it looks like you've currently claimed 1 issue in this repository. We encourage new contributors to focus their efforts on at most 1 issue at a time, so please complete your work on your other claimed issues before trying to claim this issue again.

We look forward to your valuable contributions!

adityamudgil2505:
@zulipbot claim

zulipbot:
Welcome to Zulip, @adityamudgil2505! We just sent you an invite to collaborate on this repository at https://github.com/zulip/zulip-desktop/invitations. Please accept this invite in order to claim this issue and begin a fun, rewarding experience contributing to Zulip!

Here's some tips to get you off to a good start:
* Join me on the [Zulip developers' server](https://chat.zulip.org), to get help, chat about this issue, and meet the other developers.
* [Unwatch this repository](https://help.github.com/articles/unwatching-repositories/), so that you don't get 100 emails a day.

As you work on this issue, you'll also want to refer to the [Zulip code contribution guide](https://zulip.readthedocs.io/en/latest/contributing/index.html), as well as the rest of the developer documentation on that site.

See you on the other side (that is, the pull request side)!

adityamudgil2505:
@timabbott Do I need to solve this error or ignore it while testing?
`Unexpected negated condition.  no-negated-condition`

PR #881, opened 2020-03-02
Solve #878 Converting promise to async-await

adityamudgil2505:
Convert all of our asynchronous code using .then() to use async/await instead. Hence solved issue #878 

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [x] macOS

### DIFF
--- a/app/renderer/js/pages/preference/new-server-form.ts
+++ b/app/renderer/js/pages/preference/new-server-form.ts
@@ -58,16 +58,18 @@ class NewServerForm extends BaseComponent {
 		this.$newServerUrl = this.$newServerForm.querySelectorAll('input.setting-input-value')[0] as HTMLInputElement;
 	}
 
-	submitFormHandler(): void {
+	async submitFormHandler(): Promise<void> {
 		this.$saveServerButton.innerHTML = 'Connecting...';
-		DomainUtil.checkDomain(this.$newServerUrl.value).then(serverConf => {
-			DomainUtil.addDomain(serverConf).then(() => {
-				this.props.onChange(this.props.index);
-			});
-		}, errorMessage => {
+		let serverConf;
+		try {
+			serverConf = await DomainUtil.checkDomain(this.$newServerUrl.value);
+		} catch (errorMessage) {
 			this.$saveServerButton.innerHTML = 'Connect';
 			alert(errorMessage);
-		});
+			return;
+		}
+		await DomainUtil.addDomain(serverConf);
+		this.props.onChange(this.props.index);
 	}
 
 	openCreateNewOrgExternalLink(): void {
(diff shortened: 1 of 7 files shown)


timabbott:
@adityamudgil2505 how have you tested your changes?  Ideally, we'd manually test each code path to confirm we didn't accidentally break it in the conversion.

Also, we should use re-enable the linter rule in `package.json` as part of this series.

andersk:
You don’t need to `return Promise.resolve(foo)` from an `async` function; you can just `return foo` (or omit the `return` statement for `Promise<void>`).

adityamudgil2505:
@timabbott @akashnimare Please review

app/renderer/js/utils/proxy-util.ts
```diff
@@ -121,7 +121,8 @@ class ProxyUtil {
 			return socksString;
 		})();
 
-		Promise.all([httpProxy, httpsProxy, ftpProxy, socksProxy]).then(values => {
+		(async () => {
```
andersk: Instead of adding an `async` lambda here, we should make the entire `resolveSystemProxy` function `async`.

app/renderer/js/pages/preference/new-server-form.ts
```diff
@@ -58,16 +58,16 @@ class NewServerForm extends BaseComponent {
...
-		DomainUtil.checkDomain(this.$newServerUrl.value).then(serverConf => {
-			DomainUtil.addDomain(serverConf).then(() => {
-				this.props.onChange(this.props.index);
-			});
-		}, errorMessage => {
+		try {
+			const serverConf = await DomainUtil.checkDomain(this.$newServerUrl.value);
+			await DomainUtil.addDomain(serverConf);
+			this.props.onChange(this.props.index);
+		}	catch (errorMessage) {
```
andersk: This previously caught errors from `checkDomain` only, which probably makes more sense.

app/renderer/js/tray.ts
```diff
@@ -51,7 +51,7 @@ const config = {
 	thick: process.platform === 'win32'
 };
 
-const renderCanvas = function (arg: number): Promise<HTMLCanvasElement> {
+const renderCanvas = async function (arg: number): Promise<HTMLCanvasElement> {
```
andersk: This function doesn’t actually do anything asynchronous. We should maybe just migrate it to `function (arg: number): HTMLCanvasElement`.

adityamudgil2505: If we don't use async, eslint will give us an error.
`Functions that return promises must be async.  @typescript-eslint/promise-function-async`.

andersk: Hence the suggestion to not return a `Promise`. This can just return `HTMLCanvasElement`—or I guess it’s actually `HTMLCanvasElement | undefined`.

andersk:
You seem to have committed an entire `node_modules` folder there. I’ve fixed that for you and made some other minor changes.

PR #887, opened 2020-03-03
dependencies: Use one package.json structure

andersk:
The two `package.json` structure is [no longer needed](https://www.electron.build/tutorials/two-package-structure).


abhigyank:
Are we using the latest electron-builder for this?

package.json
```diff
@@ -50,9 +50,7 @@
       "**/*.node"
     ],
     "files": [
-      "**/*",
-      "!docs${/*}",
-      "!node_modules/@paulcbetts/cld/deps/cld${/*}"
+      "app/**/*"
     ],
```
akashnimare: @andersk we added this because the spellchecker module is generating a lot of unnecessary binary files which we don't need in the app. I don't know the current status but this was the case then. The app size will increase if we remove this.

akashnimare (changes requested):

andersk:
> Are we using the latest electron-builder for this?

We are using the latest everything, as of yesterday. (Note though that `electron-builder` 8 was released three years ago.)

package.json
andersk: We haven’t had `@paulcbetts/cld` since 107e5229141879b69225153dc9e6d15f54c8c7bf. On further investigation, it was renamed to `cld`. Fixed in master.

PR #888, opened 2020-03-03
sentry-util: Hard-code the Sentry DSN.

andersk:
Commit 088ddf9c62d1a181b88da25c45c1d996250bfaba (#755) does not work, because neither the `.env` file nor the environment variables it provides are available to normal users at runtime.  This silently broke Sentry data collection.  When we upgraded `@sentry/electron` in commit 107e5229141879b69225153dc9e6d15f54c8c7bf, the silent failure became an error that prevented the app from starting.

The Sentry DSN is not a secret, so we should just commit it to the repository.


timabbott:
lgtm; I'll let @akashnimare merge this one after taking a look.

akashnimare:
Looks good. Merged.

abhigyank:
The reason I believe the sentry dsn should not be public is because it could be then be used in any application and bogus error reports would be reported. Further, someone with malicious intent could bombard our sentry with not actual errors which would be a problem.

andersk:
@abhigyank It has always been public:

```console
$ wget https://github.com/zulip/zulip-desktop/releases/download/v3.0.0/Zulip-3.0.0-x86_64.AppImage
$ chmod +x Zulip-3.0.0-x86_64.AppImage
$ ./Zulip-3.0.0-x86_64.AppImage --appimage-extract
$ npx asar extract ./squashfs-root/resources/app.asar app
$ grep dsn app/renderer/js/utils/sentry-util.js
            dsn: 'https://628dc2f2864243a08ead72e63f94c7b1@sentry.io/204668',
```

It’s fundamentally not possible to hide a secret in a publicly distributed app. Bogus reports can be filtered out in Sentry itself.

abhigyank:
In that case, this seems fine.

PR #890, opened 2020-03-04
macOS: Fix undo redo not working on macOS.

### LINKED ISSUE #866: Undo not working in compose box on macOS
EDIT: Observed only on macOS so far. Seems not to affect Linux; untested on Windows.

I (on Zulip desktop 4.0.0) and another member of the Recurse Center noticed that undo isn't working in the compose box on macOS, either from the Edit menu or by using command-Z. Immediately after typing some text, if you try to undo, your input doesn't change at all.

I'm not sure if this is a bug, or a feature that hasn't been introduced yet.

akashnimare:
The default API provided by Electron doesn't work
as expected. More info here -
https://github.com/electron/electron/issues/15728

Fixes: #866.

---

**What's this PR do?**

**Screenshots?**

  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [x] macOS

### DIFF
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -737,7 +737,7 @@ class ServerManagerView {
 		return !(url.endsWith('/login/') || this.tabs[tabIndex].webview.loading);
 	}
 
-	getActiveWebview(): any {
+	getActiveWebview(): Electron.WebviewTag {
 		const selector = 'webview:not(.disabled)';
 		const webview: Electron.WebviewTag = document.querySelector(selector);
 		return webview;
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -739,6 +739,14 @@ class ServerManagerView {
 		return !(url.endsWith('/login/') || this.tabs[tabIndex].webview.loading);
 	}
 
+	getActiveWebview(): any {
+		const selector = 'webview:not([class*=disabled])';
+		const webview = document.querySelector(selector);
+		if (webview) {
+			return webview;
+		}
+	}
+
 	addContextMenu($serverImg: HTMLImageElement, index: number): void {
 		$serverImg.addEventListener('contextmenu', e => {
 			e.preventDefault();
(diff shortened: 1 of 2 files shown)


vsvipul:
LGTM! Can't test this though as I don't have a mac.

andersk:
electron/electron#15728 was reported with Electron 3. Did you confirm that this is still an issue with Electron 8?

akashnimare:
> [electron/electron#15728](https://github.com/electron/electron/issues/15728) was reported with Electron 3. Did you confirm that this is still an issue with Electron 8?

@andersk yeah, it's still broken. Let's merge?

app/renderer/js/main.ts
```diff
@@ -739,6 +739,14 @@ class ServerManagerView {
 		return !(url.endsWith('/login/') || this.tabs[tabIndex].webview.loading);
 	}
 
+	getActiveWebview(): any {
```
andersk: `any` → `Electron.WebviewTag`

app/renderer/js/main.ts
```diff
@@ -739,6 +739,14 @@ class ServerManagerView {
...
+		const selector = 'webview:not([class*=disabled])';
+		const webview = document.querySelector(selector);
+		if (webview) {
```
andersk: Why the `if`?

app/renderer/js/main.ts
```diff
@@ -739,6 +739,14 @@ class ServerManagerView {
 		return !(url.endsWith('/login/') || this.tabs[tabIndex].webview.loading);
 	}
 
+	getActiveWebview(): any {
+		const selector = 'webview:not([class*=disabled])';
```
andersk: `'webview:not(.disabled)`?

andersk (commented):
I guess so. We should also get the Electron issue reopened or open a new one.

app/renderer/js/main.ts
akashnimare: This is just to be sure that we get the webview. Can also remove this.

app/renderer/js/main.ts
akashnimare: Yup, let me update it!

akashnimare:
@andersk made some changes.

adityamudgil2505:
I think the undo-redo issue is being solved. If there is any remaining work @akashnimare please share it with me.

akashnimare:
@adityamudgil2505 can you get the Electron issue (https://github.com/electron/electron/issues/15728) reopened or maybe open a new one since it still doesn't work on v8?

PR #897, opened 2020-03-11
[WIP] Fix #650 Broken back button 

### LINKED ISSUE #650: Broken back and forward when exiting settings UI
Looks like `Alt + Left` and `Alt + Forward` don't work on master. This seems to have been introduced with updating to Electron v3.

adityamudgil2505:
**What's this PR do?**
Solve the issue of the broken back button #650 . When someone goes to a setting option, then want to return back to the previous tab of the organization, then the user can do using the back button.

This is being solved by adding a new variable that will keep track of the last tab(organization window) before going to the setting and when the user click on the back button then he will move to tab whose value is being stored in the variable.

**GIF**

Working

  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [x] macOS

### DIFF
--- a/app/renderer/js/components/webview.ts
+++ b/app/renderer/js/components/webview.ts
@@ -25,6 +25,7 @@ export default class WebView extends BaseComponent {
 	$webviewsContainer: DOMTokenList;
 	$el: Electron.WebviewTag;
 	domReady?: Promise<void>;
+	prevTabIndex: number;
 
 	// This is required because in main.js we access WebView.method as
 	// webview[method].
@@ -260,16 +261,18 @@ export default class WebView extends BaseComponent {
 		this.$el.loadURL(url);
 	}
 
-	back(): void {
+	back(): boolean {
 		if (this.$el.canGoBack()) {
 			this.$el.goBack();
 			this.focus();
+			return true;
 		}
+		return false;
 	}
 
 	canGoBackButton(): void {
 		const $backButton = document.querySelector('#actions-container #back-action');
-		if (this.$el.canGoBack()) {
+		if (this.$el.canGoBack() || this.prevTabIndex >= 0) {
 			$backButton.classList.remove('disable');
 		} else {
 			$backButton.classList.add('disable');
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -320,7 +320,7 @@ class ServerManagerView {
 			}
 			// checkDomain() and webview.load() for lastActiveTab before the others
 			DomainUtil.updateSavedServer(servers[lastActiveTab].url, lastActiveTab);
-			this.activateTab(lastActiveTab);
+			this.activateTab(lastActiveTab, true);
 			for (const [i, server] of servers.entries()) {
 				// after the lastActiveTab is activated, we load the others in the background
 				// without activating them, to prevent flashing of server icons
@@ -413,7 +413,21 @@ class ServerManagerView {
 			this.openSettings('General');
 		});
 		this.$backButton.addEventListener('click', () => {
-			this.tabs[this.activeTabIndex].webview.back();
+			if (this.tabs.length === 1 && this.tabs[this.activeTabIndex].props.name === 'Settings') {
+				return;
+			}
+			const canMoveBack = this.tabs[this.activeTabIndex].webview.back();
+			if (canMoveBack && this.tabs[this.activeTabIndex].props.name === 'Settings') {
+				setTimeout(async _ => {
+					const nav = this.currentSettingWindowName();
+					await this.tabs[this.functionalTabs.Settings].webview.send('switch-settings-nav', nav);
+				}, 50);
+			}
+			if (!canMoveBack && this.tabs[this.activeTabIndex].webview.prevTabIndex >= 0) {
+				const newTab = this.tabs[this.activeTabIndex].webview.prevTabIndex;
+				this.tabs[this.activeTabIndex].webview.prevTabIndex = -1;
+				this.activateTab(newTab, true, true);
+			}
 		});
 
 		this.sidebarHoverEvent(this.$addServerButton, this.$addServerTooltip, true);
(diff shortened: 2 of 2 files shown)


abhigyank:
@adityamudgil2505 Does this add working for back cases with in the settings tab?

adityamudgil2505:
@abhigyank No, it will get back to organisation.

abhigyank:
I am not sure switching back to the organization after navigating a bit in the settings is a behavior that the user would expected. If the back is enabled in settings I suppose the user would then expect it to work with the settings window? I am not sure how the UX would be based on this change. Maybe someone else suggest if this'd be a good behavior. cc @rishig

adityamudgil2505:
@abhigyank As both the ways have their own advantage and the way to solve this issue using both methods is similar.

app/renderer/js/components/tab.ts
```diff
@@ -28,7 +28,12 @@ export default class Tab extends BaseComponent {
 	}
 
 	activate(): void {
-		this.$el.classList.add('active');
```
andersk: This removal breaks highlighting of the active tab.

app/renderer/js/components/tab.ts
```diff
@@ -28,7 +28,12 @@ export default class Tab extends BaseComponent {
...
+		if (this.webview.$el) {
+			if (this.webview.$el.classList.contains('disabled')) {
+				this.webview.$el.classList.remove('disabled');
+			}
+			this.webview.$el.classList.add('active');
+		}
```
andersk: This should already happen inside `this.webview.load()`, right?

app/renderer/js/main.ts
```diff
@@ -410,10 +412,16 @@ class ServerManagerView {
...
 			this.openSettings('General');
 		});
 		this.$backButton.addEventListener('click', () => {
-			this.tabs[this.activeTabIndex].webview.back();
+			if (this.tabs[this.activeTabIndex].webview.props.name === 'Settings') {
+				this.tabs[this.activeTabIndex].deactivate();
+				this.activateTab(this.previousTabIndex);
+			} else {
+				this.tabs[this.activeTabIndex].webview.back();
+			}
```
andersk: This logic has some problems. What if there is no previous tab? What if the previous tab is also the Settings tab? Empirically, if I open the app with a freshly deleted configuration and the first thing I click on is the Back button, I end up at a white screen from which neither the Settings button nor the Add organization button do anything.

adityamudgil2505: @andersk I need to fix the issue from the beginning so that it will take care of all the test cases. Thanks for pointing me out about this case.

app/renderer/js/components/tab.ts
adityamudgil2505: Yes. I have to use `this.webview.load()` instead of rewriting a new function.

app/renderer/js/components/tab.ts
adityamudgil2505: It breaks, but later it will handle the highlighting of the active tab. But I need to go through the code from beginning to cover all the test cases.

adityamudgil2505:
@andersk @abhigyank I updated the code. Now when the user opens the application after installing, then the back button will be disabled. After adding one or more organizations, it will work as what back button has to do. We can also use the back button in settings. The code is very messy, as there are many corner cases.
Back
Idea:
I created new data member in webview which will keep track of previous webview's index because we can move back inside webview without any problem but we can't move to different webview. 
Now we have two types of movement
1. When going in the forward direction, when we just update the previous index of the new webview with the previous one.
2. When going in a backward direction, first we need to update the previous webview index of the current webview tab to -1 to avoid infinite loop and then move.

To handle this, I modify activateTab to add one more parameter which will tell whether we are going in forward direction or backward direction.
This will handle almost every case but it doesn't handle when we install the application for the first time, so I disable the back button there. We can enable the functionality of the back button in this case also but it requires some hard code.

app/renderer/js/components/webview.ts
```diff
@@ -25,6 +25,7 @@ export default class WebView extends BaseComponent {
 	$webviewsContainer: DOMTokenList;
 	$el: Electron.WebviewTag;
 	domReady?: Promise<void>;
+	previoueTabIndex: number;
```
manavmehta: fix typo
`previoueTabIndex` -> `previousTabIndex` or `prevTabIndex`

app/renderer/js/components/webview.ts
```diff
@@ -260,16 +261,18 @@ export default class WebView extends BaseComponent {
...
 			this.focus();
+			return true;
 		}
+		return false;
 	}
 
 	canGoBackButton(): void {
 		const $backButton = document.querySelector('#actions-container #back-action');
-		if (this.$el.canGoBack()) {
+		if (this.$el.canGoBack() || this.previoueTabIndex >= 0) {
```
manavmehta: same typo here

app/renderer/js/main.ts
```diff
@@ -740,6 +757,12 @@ class ServerManagerView {
 		return webview;
 	}
 
+	getCurrentSettingWindowName(): string {
+		const selector = 'webview.active';
+		const webview: Electron.WebviewTag = document.querySelector(selector);
+		return webview.getAttribute('src').split('#')[1];
+	}
+
```
manavmehta: Just out of curiosity, can't we rename this to something like `getCurrentWindowName` or in-fact `currentWindowName` as this can further be used in fetching a focussed 'non-setting' window ?
@akashnimare @timabbott

manavmehta (commented):
@adityamudgil2505 can you please fix the typo and reconsider the name of the `getCurrentSettingWindowName() `  function as it can be used elsewhere too, afaik.
Otherwise LGTM

app/renderer/js/main.ts
adityamudgil2505: @manavmehta  Actually this function is specifically to get current Settings Window name, so function name should include setting in its name. Although, I will rename it to currentSettingWindowName.

app/renderer/js/components/webview.ts
adityamudgil2505: Yes, its by mistake. I will change it.

manavmehta:
@abhigyank Is this g2g or some work is left?
cc @adityamudgil2505

PR #901, opened 2020-03-19
[WIP] Build app installer via GitHub actions.

### LINKED ISSUE #406: Build artifacts using CI
macOS/Linux - Travis
Windows - AppVeyor

adityamudgil2505:
**What's this PR do?**
This will allow artifacts to upload to the GitHub package registry as draft release. This solves the Issue #406 .

Earlier we are using Travis CI for macOS and Linux, and appveyor for windows testing. Now we are shifting this to GitHub actions which will also allow artifacts to upload to the GitHub package registry as a draft release.
We need to upgrade electron-builder version from 22.3.2 to 22.4.1 so that our windows pipeline will work properly.

  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [x] macOS

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -1,0 +1,58 @@
+name: Zulip Desktop CI
+
+on: [push]
+
+jobs:
+  test:
+    name: Test on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: 
+        - ubuntu-latest
+        - windows-latest
+#         - macOS-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Setting up Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12.x'
+
+    - name: Setup npm cache
+      id: npm-cache
+      run: |
+         echo "::set-output name=dir::$(npm config get cache)"
+         
+    - uses: actions/cache@v1
+      with:
+        path: ${{ steps.npm-cache.outputs.dir }}
+        key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-node-
+          
+    - name: Installing npm modules
+      run: npm install
+
+    - name: Running Tests
+      run: |
+        npm run test
+        npm run build
+
+    - name: Building app installers
+      uses: samuelmeuli/action-electron-builder@v1.3.0
+      with:
+        github_token: ${{ secrets.github_token }}
+        mac_certs: ${{ secrets.mac_certs }}
+        mac_certs_password: ${{ secrets.mac_certs_password }}
+
+    - name: Listing content in ./dist
+      run: ls ./dist
+
+    - name: Uploading artifacts
+      uses: actions/upload-artifact@v1.0.0
+      with:
+        name: ${{matrix.os}}
+        path: ./dist
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
     "mas": "tsc && electron-builder --mac mas",
     "travis": "cd ./scripts && ./travis-build-test.sh",
     "build-locales": "node tools/locale-helper",
-    "build": "tsc --publish never"
+    "build": "tsc",
+    "publish": "never"
   },
   "pre-commit": [
     "test"
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
     "dist": "tsc && electron-builder",
     "mas": "tsc && electron-builder --mac mas",
     "travis": "cd ./scripts && ./travis-build-test.sh",
-    "build-locales": "node tools/locale-helper"
+    "build-locales": "node tools/locale-helper",
+    "build": "tsc --publish never"
   },
   "pre-commit": [
     "test"
--- a/package-lock.json
+++ b/package-lock.json
@@ -5550,25 +5550,29 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "bundled": true,
+          "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
+          "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
           "dev": true,
           "optional": true
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "bundled": true,
+          "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
+          "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
           "dev": true,
           "optional": true
         },
         "aproba": {
           "version": "1.2.0",
-          "bundled": true,
+          "resolved": "https://registry.npmjs.org/aproba/-/aproba-1.2.0.tgz",
+          "integrity": "sha512-Y9J6ZjXtoYh8RnXVCMOU/ttDmk1aBjunq9vO0ta5x85WDQiQfUF9sIPBITdbiiIVcBo03Hi3jMxigBtsddlXRw==",
           "dev": true,
           "optional": true
         },
         "are-we-there-yet": {
           "version": "1.1.5",
-          "bundled": true,
+          "resolved": "https://registry.npmjs.org/are-we-there-yet/-/are-we-there-yet-1.1.5.tgz",
+          "integrity": "sha512-5hYdAkZlcG8tOLujVDTgCT+uPX0VnpAH28gWsLfzpXYm7wP6mp5Q/gYyR7YQ0cKVJcXJnl3j2kpBan13PtQf6w==",
           "dev": true,
           "optional": true,
           "requires": {
@@ -5578,13 +5582,15 @@
         },
         "balanced-match": {
           "version": "1.0.0",
-          "bundled": true,
+          "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.0.tgz",
+          "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c=",
           "dev": true,
           "optional": true
         },
         "brace-expansion": {
           "version": "1.1.11",
-          "bundled": true,
+          "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
+          "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
           "dev": true,
           "optional": true,
           "requires": {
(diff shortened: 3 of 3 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,52 @@
...
+
+jobs:
+  test:
+    name: Test on node ${{ matrix.node_version }} and ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        node_version:
+        - 10
+        - 12
```
priyank-p: We should only test latest node version. Electron internally uses its own bundled node version so testing with different node version have no effect.

adityamudgil2505:
@priyank-p  @akashnimare Linux app is successfully built and the artifact is uploaded in both node versions.[Check this link to see the logs and to download the uploaded artifact]( https://github.com/adityamudgil2505/zulip-desktop/pull/3/checks?sha=f59563d7c633259e8b57ab46d128f8bbc9c4e413)

But in case of Windows, the build was failed due to the following error [Error link](https://github.com/adityamudgil2505/zulip-desktop/pull/3/checks?check_run_id=520513844#step:8:101)
and I found that it's a electron-builder version 22.3.2 problem from this https://github.com/electron-userland/electron-builder/pull/4620#issuecomment-592118743

And for mac will successfully build only if we add the environment variable 
```
mac_certs: ${{ secrets.mac_certs }}
mac_certs_password: ${{ secrets.mac_certs_password }}
```
you can find the working of this in [this link](https://github.com/adityamudgil2505/zulip-desktop/pull/3/checks?check_run_id=520427728#step:8:27)

I have one query regarding Linux build, the artifact upload is of size around 400MB and it contains following files after downloading, do we need all of these to build for Linux or only a few of them because its time taking to download all the files.
Screenshot 2020-03-20 at 7 40 13 PM

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,52 @@
+name: CI/CD
+
+on:
+  push:
+    branches: 
+    - master
+  pull_request:
+    branches: 
+    - master
```
priyank-p: For now, let just make this `on: [push]` so it runs on all push. We sometimes utilize development branches and would want the CI to run.

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,76 @@
...
+      if: matrix.os != 'windows-latest'
+      run: |
+          mkdir artifacts
+          mv "dist/(*.exe,*.deb,*.AppImage,*.dmg)" artifacts || true
+
+    - name: Cleanup artifacts Win
+      if: matrix.os == 'windows-latest'
+      run: |
+          mkdir artifacts
+          mv dist/*.exe artifacts
```
priyank-p: This cleanup blocks are unnecessary! There is no reason for the artifacts to be there when the build runs unless there are being cached and brought back which does not seem to be the case.

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,76 @@
...
+
+    - name: Deploying to GPR
```
priyank-p: Name this step, `Uploading artifacts` or similar. "Deploying" and "GPR" and unrelated words to describe the current step.

Also, what is GPR? GitHub Package Repository? If we are not interacting with it at all here.

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,76 @@
...
+      uses: actions/upload-artifact@v1.0.0
+      with:
+        name: Zulip - ${{matrix.os}} -${{matrix.node_version}}
```
priyank-p: Drop the `${matrix.node_version}` since that is not useful. Also, can you post the screenshot of how it looks in github so we can figure out what exactly to name this?

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,76 @@
...
+        - windows-latest
+        - macOS-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Setting up Node.js ${{ matrix.node_version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node_version }}
```
priyank-p: Just let the action set up the latest version of the node here.

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,76 @@
...
+
+    - name: Get npm cache directory
```
priyank-p: Name this step: `Setup npm cache` or similar.

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,76 @@
...
+      id: npm-cache
```
priyank-p, line 25: What does `id` do here? Just asking because I am unsure of its purpose.

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,76 @@
...
+        restore-keys: |
+          ${{ runner.os }}-node-
+          
+    - name: Installing npm modules
+      run: npm install
+
+    - name: Running Tests
+      run: npm run test
+
+    - name: Electron Builder Action
```
priyank-p: Name this: `Building app installers` or similar.

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,76 @@
...
+      uses: samuelmeuli/action-electron-builder@v1.3.0
+      with:
+        github_token: ${{ secrets.github_token }}
+        mac_certs: ${{ secrets.mac_certs }}
+        mac_certs_password: ${{ secrets.mac_certs_password }}
+
+    - name: See files in dist folder
+      run: ls ./dist
```
priyank-p, line 52: Here is one more thing you can do to make the artifacts convenient. Rename the output directory to have the same name as it is on GitHub releases. For example, when building for macOS, the electron builder may spit out a dmg named `Zulip-macOS-4.0.3.dmg` when on releases page it is `Zulip-4.0.3.dmg`.  Let me know if it unclear.

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,76 @@
+name: CI/CD
```
priyank-p: Let's call it `Zulip Desktop CI` or similar?

priyank-p (commented):
@adityamudgil2505 I posted a bunch of comments.

priyank-p:
@adityamudgil2505 For Linux (and for other platform), we do need all the files in dist. 

Regarding the windows issue, you can try and downgrade the electron-build and see if it works. If it works. then add a commit changing the version and mention this issue in the commit message. 

As for the macOS, once we have everything else working, I think we will push this to a branch in this repository and add the secrets to test if it works. If it works we will push it to master.

Finally, thanks for working on this issue.

.github/workflows/ci_cd.yml
priyank-p: I saw how it looks in the GitHub UI. We should definitely not include the node version in its name. Likewise, `Zulip -` is also unnecessary. Just having an OS name will suffice here.

.github/workflows/ci_cd.yml
adityamudgil2505: According to the documentation, the use of id is unnecessary for us [link for the purpose of id](https://github.com/actions/cache#skipping-steps-based-on-cache-hit). But if we don't use id then  CI will fail at step "Run actions/cache@v1" and also the documentation given uses id [link](https://github.com/actions/cache/blob/master/examples.md#node---npm)

.github/workflows/ci_cd.yml
adityamudgil2505: I think these things are done by modifying package.json. I explored the way in which GitHub action can access package.json file content, but can't find any way to do this. Currently we get  ubuntu-latest > Zulip-4.1.0-beta-amd64.deb this type of name.
Also, we can name the artifact folder according to the version of Zulip by adding one more environment variable.

.github/workflows/ci_cd.yml
adityamudgil2505: Yes, I changed the name. I am new to the action feature of GitHub, and I thought it will upload the artifact to GitHub Package Repository (GRP) but it has a different way to do this. Currently, we can use this artifact from the checks view of the pull request.

.github/workflows/ci_cd.yml
adityamudgil2505: If we enter the following:
```
- name: Uploading artifacts
      uses: actions/upload-artifact@v1.0.0
      with:
        name: ${{matrix.os}}
```
as output we get this 
Screenshot

.github/workflows/ci_cd.yml
adityamudgil2505: Zulip Desktop CI is perfect for this.

adityamudgil2505:
@priyank-p I made all the changes. For node-version, there is no way for the latest version in the documentation. We need to set it manually https://github.com/actions/setup-node/issues/61. But I found an open merge request https://github.com/actions/setup-node/pull/104 which solves this issue. 
After updating the version of Electron-builder from 22.3.2 to 22.4.1 and now the windows pipeline is running fine. 
[Link of Linux, Windows CI](https://github.com/adityamudgil2505/zulip-desktop/pull/3/checks?sha=860b01023a1940d41f9e80d5519f631dcf8feee3)

.github/workflows/ci_cd.yml
priyank-p: For this case, just using a node script that renames the file is enough but we can revisit this later.

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,56 @@
+name: Zulip Desktop CI
+
+on: [push]
+
+jobs:
+  test:
+    name: Test on ${{ matrix.os }}
```
priyank-p, line 7: You can use `${{ runner.os }}` here to get a readable name of the platform i.e. Linux instead of ubuntu-latest. Also, let's drop the `Test on` part there.

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,56 @@
...
+        - windows-latest
+#         - macOS-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Setting up Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: '*'
```
priyank-p: Here, let's do `12.x` here. The use of `*` sounds like it will use beta version of node which we should avoid.

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,56 @@
...
+        mac_certs_password: ${{ secrets.mac_certs_password }}
+
+    - name: See files in dist folder
+      run: ls ./dist
+
+    - name: Uploading artifacts
+      uses: actions/upload-artifact@v1.0.0
+      with:
+        name: ${{matrix.os}}
+        path: ./dist
```
priyank-p: Add an EOL (new line) here.

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,56 @@
...
+      run: npm run test
+
+    - name: Building app installers
+      uses: samuelmeuli/action-electron-builder@v1.3.0
+      with:
+        github_token: ${{ secrets.github_token }}
+        mac_certs: ${{ secrets.mac_certs }}
+        mac_certs_password: ${{ secrets.mac_certs_password }}
+
+    - name: See files in dist folder
```
priyank-p: Minor nit: This can be `Listing content in ./dist`

.github/workflows/ci_cd.yml
```diff
@@ -0,0 +1,56 @@
...
+          ${{ runner.os }}-node-
+          
+    - name: Installing npm modules
+      run: npm install
+
+    - name: Running Tests
+      run: npm run test
+
+    - name: Building app installers
+      uses: samuelmeuli/action-electron-builder@v1.3.0
```
priyank-p, line 45: Two of the major things I missed in the previous review:
  - We need to add a `build` npm script that runs `tsc`. We need to do this in order for this action to build typescript files before creating installers. Currently, the installers created do not run because of this.
  - In `package.json` we need to add `publish` property under `build` and set it to `never` so the release is not automatically pushed to GitHub. See https://www.electron.build/configuration/publish.

priyank-p (commented):
@adityamudgil2505 This is looking great! I added some more comments. Also, please make sure the changes to electron-builder version is separated into its own separate preparatory commit. And other changes are squashed into their own coherent commits.

Checkout [Zulip commit guidelines](https://zulip.readthedocs.io/en/latest/contributing/version-control.html) for detailed information on how to create coherent commits.

.github/workflows/ci_cd.yml
priyank-p: Let's use `${{ runner.os }}` instead then. The current output is good too because we just need to know what platform an installer belongs to.

.github/workflows/ci_cd.yml
adityamudgil2505: Using `${{ runner.os }}` instead of `${{ matrix.os }}` will give error
 `Unrecognized named-value: 'runner'` and the pipeline doesn't run.

.github/workflows/ci_cd.yml
adityamudgil2505: Yes, 12.x will use the recommended version.

adityamudgil2505:
@priyank-p Please review.

.github/workflows/ci_cd.yml
priyank-p: Hmm, It is documented in the [GitHub Actions docs](https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#runner-context) so I do not know why it won't work? Perhaps, it only work if it is used for steps and env declartions?

.github/workflows/ci_cd.yml
priyank-p: @adityamudgil2505 I don't think you addressed these two issues yet in your latest push.

adityamudgil2505: @priyank-p I added a new script in package.json file.
`"build": "tsc --publish never"` 
and run this script in .yml file
```
- name: Running Tests
       run: |
         npm run test
         npm run build
```

package.json
```diff
@@ -37,7 +37,8 @@
     "dist": "tsc && electron-builder",
     "mas": "tsc && electron-builder --mac mas",
     "travis": "cd ./scripts && ./travis-build-test.sh",
-    "build-locales": "node tools/locale-helper"
+    "build-locales": "node tools/locale-helper",
+    "build": "tsc --publish never"
```
priyank-p: This should just be `tsc`. The `--publish never` _belongs_ to electron-builder cli hence the suggestion the add a `"publish": "never"` under the `"build"` section in package.json.

adityamudgil2505: @priyank-p I modified it.

PR #902, opened 2020-03-21
Add feature to set application language.

### LINKED ISSUE #855: Change language desktop app
Hello.

I am using the Windows 10 operating system and the latest current version of Zulip. The fact is that I can’t find a language change in the application, I suspect that the application is taking information from the operating system. I could use the default language, but it is not readable. Please tell me how to change the language in the desktop version without changing the language in the operating system.
---

- **Operating System**:
  - [x] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

image

manavmehta:
**What's this PR do?**
Fixes: #855 
Adds feature that lets user to change Zulip desktop client language without requiring to change the system language as was the case earlier

Earlier, Zulip client used to fetch system locale thus setting default language to the client. Now we enable user to set custom language to the client.

**Screenshots?**
Screenshot 2020-03-21 at 12 04 20 PM

[✓ ] macOS Catalina 10.15.2

### DIFF
--- a/app/renderer/css/preference.css
+++ b/app/renderer/css/preference.css
@@ -724,3 +724,13 @@ i.open-network-button {
         margin-top: 10px;
     }
 }
+
+.lang-menu {
+    font-size: 13px;
+    font-weight: bold;
+    background: rgba(78, 191, 172, 1.000);
+    width: 100px;
+    height: 38px;
+    color: rgba(255, 255, 255, 1.000);
+    border-color: rgba(0, 0, 0, 0);
+}
--- a/app/renderer/js/pages/preference/general-section.ts
+++ b/app/renderer/js/pages/preference/general-section.ts
@@ -92,10 +92,17 @@ export default class GeneralSection extends BaseSection {
 				</div>
 				<div class="title">${t.__('Advanced')}</div>
 				<div class="settings-card">
+
 					<div class="setting-row" id="enable-error-reporting">
 						<div class="setting-description">${t.__('Enable error reporting (requires restart)')}</div>
 						<div class="setting-control"></div>
 					</div>
+					
+					<div class="setting-row" id="app-language">
+						<div class="setting-description">${t.__('App language (requires restart)')}</div>
+						<div  id="lang-div" class="lang-div"></div>
+					</div>
+
 					<div class="setting-row" id="add-custom-css">
 						<div class="setting-description">
 							${t.__('Add custom CSS')}
@@ -135,7 +142,7 @@ export default class GeneralSection extends BaseSection {
 						<button class="reset-data-button red w-150">${t.__('Reset App Data')}</button>
 					</div>
 				</div>
-            </div>
+			</div>
 		`;
 	}
 
@@ -159,6 +166,7 @@ export default class GeneralSection extends BaseSection {
 		this.updateQuitOnCloseOption();
 		this.updatePromptDownloadOption();
 		this.enableErrorReporting();
+		this.setLocale();
 
 		// Platform specific settings
 
@@ -393,6 +401,29 @@ export default class GeneralSection extends BaseSection {
 		const resetDataButton = document.querySelector('#resetdata-option .reset-data-button');
 		resetDataButton.addEventListener('click', () => {
 			this.clearAppDataDialog();
+		});
+	}
+
+	setLocale(): void {
+		const langDiv: HTMLSelectElement = document.querySelector('.lang-div');
+		// This path is for the JSON file that stores key: value pairs for supported locales
+		const path = __dirname.replace('renderer/js/pages/preference', 'translations/supported-locales.json');
+		const data = fs.readFileSync(path, {encoding: 'utf8'});
+		const langs = JSON.parse(data);
+		const langList = this.generateSelectTemplate(langs, 'lang-menu');
+		langDiv.innerHTML += langList;
+		// langMenu is the select-option dropdown menu formed after executing the previous command
+		const langMenu: HTMLSelectElement = document.querySelector('.lang-menu');
+
+		// The next three lines set the selected language visible on the dropdown button
+		let langIndex = ConfigUtil.getConfigItem('languageIndex');
+		langIndex = (langIndex === null) ? 4 : langIndex;
+		// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+		(langMenu.options[langIndex] as HTMLOptionElement).setAttribute('selected', '');
+
+		langMenu.addEventListener('change', (event: Event) => {
+			ConfigUtil.setConfigItem('appLanguage', (event.target as HTMLSelectElement).value);
+			ConfigUtil.setConfigItem('languageIndex', (event.target as HTMLSelectElement).selectedIndex);
 		});
 	}
 
--- a/app/renderer/js/utils/translation-util.ts
+++ b/app/renderer/js/utils/translation-util.ts
@@ -1,20 +1,15 @@
 import path from 'path';
-import electron from 'electron';
 import i18n from 'i18n';
-
-let app: Electron.App = null;
-
-/* To make the util runnable in both main and renderer process */
-if (process.type === 'renderer') {
-	app = electron.remote.app;
-} else {
-	app = electron.app;
-}
+import * as ConfigUtil from './config-util';
 
 i18n.configure({
 	directory: path.join(__dirname, '../../../translations/')
 });
 
+/* fetches the current appLocale from settings.json */
+const appLocale = ConfigUtil.getConfigItem('appLanguage');
+
+/* if no locale present in the json, en is set default */
 export function __(phrase: string): string {
-	return i18n.__({ phrase, locale: app.getLocale() ? app.getLocale() : 'en' });
+	return i18n.__({ phrase, locale: appLocale ? appLocale : 'en' });
 }
--- a/app/renderer/js/pages/preference/base-section.ts
+++ b/app/renderer/js/pages/preference/base-section.ts
@@ -40,6 +40,18 @@ export default class BaseSection extends BaseComponent {
 		}
 	}
 
+	/* a method that in future can be used to create dropdown menus using <select> <option> tags.
+		it needs an object which has ``key: value`` pairs and will return a string that can be appended to HTML
+	*/
+	generateSelectTemplate(options: {[key: string]: string}, className?: string, idName?: string): string {
+		let select = `<select class='${className}' id='${idName}'>\n`;
+		Object.keys(options).forEach(key => {
+			select += `<option value="${key}">${options[key]}</option>\n`;
+		});
+		select += '</select>';
+		return select;
+	}
+
 	reloadApp(): void {
 		ipcRenderer.send('forward-message', 'reload-viewer');
 	}
--- a/app/translations/supported-locales.json
+++ b/app/translations/supported-locales.json
@@ -1,0 +1,23 @@
+{
+    "bg": "Bulgarian",
+    "cs": "Czech",
+    "de": "Deutsch",
+    "en-GB": "English (UK)",
+    "en-US": "English (US)",
+    "en": "English",
+    "es": "Español",
+    "fr": "French",
+    "hi": "Hindi",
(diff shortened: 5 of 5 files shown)


manavmehta:
@akashnimare @timabbott @abhigyank Please review

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -418,6 +447,30 @@ export default class GeneralSection extends BaseSection {
...
+		/*
+			The next three lines set the selected language visible on the dropdown button always
+			eg if user selects Español, then he shouldn't see Bulgarian when dropdown isn't opened
+			as it is the first option in the menu
+		*/
+		let localeIndex = ConfigUtil.getConfigItem('localeIndex');
+		localeIndex = (localeIndex === null) ? 5 : localeIndex;
+		// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+		(langList.options[localeIndex] as any).setAttribute('selected', '');
+		langMenu.addEventListener('change', () => {
```
abhigyank: Pass `event` here as parameter, then there is no need of the `opt` and the for-loop below.
```ts
langMenu.addEventListener('change', (event: Event) => {
```

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -418,6 +447,30 @@ export default class GeneralSection extends BaseSection {
...
+			let opt;
+			for (opt of langList.options) {
+				if (opt.selected) {
+					break;
+				}
+			}
+			ConfigUtil.setConfigItem('appLocale', opt.value);
```
abhigyank: use `event.target.value` here and `event.target.selectedIndex` below instead.

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -418,6 +447,30 @@ export default class GeneralSection extends BaseSection {
 		});
 	}
 
+	setLocale(): void {
+		const langMenu = document.querySelector('#app-language .lang-menu');
+		const langList: any = document.querySelector('.lang-menu');
```
abhigyank: `langMenu` and `langList` both are the same I think. Can do with one of them.

app/renderer/css/preference.css
```diff
@@ -724,3 +724,13 @@ i.open-network-button {
         margin-top: 10px;
     }
 }
+
+select {
```
abhigyank: Don't use `select` here, use the class name instead.

app/renderer/css/preference.css
```diff
@@ -724,3 +724,13 @@ i.open-network-button {
...
+    font-family: Verdana;
```
abhigyank: No need of this. Default (Arial) will be inherited.

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -92,6 +92,34 @@ export default class GeneralSection extends BaseSection {
 				</div>
 				<div class="title">${t.__('Advanced')}</div>
 				<div class="settings-card">
+
+					<div class="setting-row" id="app-language">
+						<div class="setting-description">${t.__('App Laguage (requires refresh)')}</div>
```
abhigyank: @akashnimare We should go or `requires restart` here? Since the menu bar needs to be rebuild.

abhigyank (changes requested):

app/renderer/js/pages/preference/general-section.ts
abhigyank: Can use as following for TS errors - 
```ts
(event.target as HTMLSelectElement).value;
```

app/renderer/js/pages/preference/general-section.ts
abhigyank: Also instead of `any`, use `HTMLSelectElement` here.

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -418,6 +447,30 @@ export default class GeneralSection extends BaseSection {
...
+		const langList: any = document.querySelector('.lang-menu');
+		/*
+			The next three lines set the selected language visible on the dropdown button always
+			eg if user selects Español, then he shouldn't see Bulgarian when dropdown isn't opened
+			as it is the first option in the menu
+		*/
+		let localeIndex = ConfigUtil.getConfigItem('localeIndex');
+		localeIndex = (localeIndex === null) ? 5 : localeIndex;
+		// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+		(langList.options[localeIndex] as any).setAttribute('selected', '');
```
abhigyank: use `HTMLOptionElement` instead of `any`.

app/renderer/js/pages/preference/general-section.ts
manavmehta: I was also ambiguous so kept refresh for the time being, now changed to restart
If you/ @akashnimare  think otherwise, I will revert it

manavmehta:
@abhigyank thanks, refactored according to your review.

abhigyank:
LGTM!
Just fix the commit messages. Checkout Zulip [commit guidelines](https://zulip.readthedocs.io/en/latest/contributing/version-control.html) for detailed information on how to create coherent commits.

app/renderer/js/pages/preference/general-section.ts
akashnimare: Yeah, gotta add restart here.

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -92,6 +92,34 @@ export default class GeneralSection extends BaseSection {
 				</div>
 				<div class="title">${t.__('Advanced')}</div>
 				<div class="settings-card">
+
+					<div class="setting-row" id="app-language">
+						<div class="setting-description">${t.__('App Laguage (requires restart)')}</div>
+						<select id="lang-menu" class="lang-menu">
+							<option value="bg">Bulgarian</option>
```
akashnimare: I think it makes sense to store the lang names in some array and call it using a for loop or something. You can make a component for the same.

manavmehta: Changes made! Please have a look.

akashnimare: This is better.

akashnimare:
A couple of doubts -
* Electron also started supporting spellchecker in v8, can you check what's the status of that?
* I see only a couple of options in the supported languages. Are we just supporting the web app languages or what?

abhigyank:
@akashnimare This PR for the app language ( to be used in menus and settings mostly), not related to spellchecker.
I suppose the list of languages are the ones for which have translation available on the desktop app. @manavmehta can you confirm this?

manavmehta:
@abhigyank that's right
@akashnimare, As @abhigyank mentioned we use translations/'lang'.json for this
Initially we used to fetch system language which would be done manually here for end-user's convenience. Spellchecker comes into play in chats which was not the purpose hereby.
We may work on that separately forth.
2. We use all of the languages for which the translations are available in the desktop client. Further addition of supported languages in app can be appropriately done.

akashnimare:
A couple of changes -

* The border looks weird on focus, can you update it?


image

* Would it make sense to move this setting to just above the add custom CSS?

manavmehta:
@akashnimare updated the border color to transparent

> Would it make sense to move this setting to just above the add custom CSS?
>

Yeah. I thought the user would use it often so kept it above but now it looks much coherent.

Have a look!

new

akashnimare:
@manavmehta can you fix the merge conflicts?

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -418,6 +426,50 @@ export default class GeneralSection extends BaseSection {
...
+			ta: 'Tamil',
+			tr: 'Turkish'
+		};
+		const localesList = this.generateSelectTemplate(locales, 'lang-menu');
+		langDiv.innerHTML += localesList;
+		/* langMenu is the select-option dropdown menu formed after executing the previous command */
+		const langMenu: HTMLSelectElement = document.querySelector('.lang-menu');
+		/*
+			The next three lines set the selected language visible on the dropdown button always
+			eg if user selects Español, then he shouldn't see Bulgarian when dropdown isn't opened
```
akashnimare: @manavmehta need to rephrase this and don't make it gender-specific.

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -418,6 +426,50 @@ export default class GeneralSection extends BaseSection {
 		});
 	}
 
+	setLocale(): void {
+		const langDiv: HTMLSelectElement = document.querySelector('.lang-div');
+		const locales: {[key: string]: string} = {
+			bg: 'Bulgarian',
```
akashnimare: @manavmehta I was thinking about how we can sync this up with the translations we have in `/translations`. Maybe it's worth putting this in lang.json or something and then import it. cc @abhigyank

abhigyank: Hmm yes, then the languages would be dynamic. But the issue would be that `/translations` only has the `key` in the file name, not sure how we can get the language string from there if we are doing some sort of sync.

manavmehta: @akashnimare @abhigyank checkout [1219be8](https://github.com/zulip/zulip-desktop/pull/902/commits/1219be8ac7e4b023c3e8d0dd94d6edfcb4c34c15)
I hope this is what you're looking for @akashnimare ?

Also any refinements?

app/renderer/js/utils/translation-util.ts
```diff
@@ -1,20 +1,15 @@
...
 	directory: path.join(__dirname, '../../../translations/')
 });
 
+/* fetches the current appLocale from settings.json */
+const appLocale = ConfigUtil.getConfigItem('appLocale');
+
+/* if no locale present in the json, en is set default */
 export function __(phrase: string): string {
-	return i18n.__({ phrase, locale: app.getLocale() ? app.getLocale() : 'en' });
+	return i18n.__({ phrase, locale: appLocale ? appLocale : 'en' });
```
akashnimare, line 14: @manavmehta can you set "en-US" as the default locale here?

akashnimare: Also, let's rename "appLocale" to "appLanguage"?

akashnimare:
Looks good. Left a few comments. @abhigyank GTG?

app/renderer/js/utils/translation-util.ts
manavmehta: You want this change in translation-util or everywhere I used the word locale ?

akashnimare: We want to put "appLanguage" in the settings.json so make the changes where required.

abhigyank:
LGTM.

akashnimare:
Merged. Thanks, @manavmehta.

akashnimare:
@manavmehta as a follow-up can you take care of the translation thing we talked about?

manavmehta:
@akashnimare On it

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -396,6 +404,29 @@ export default class GeneralSection extends BaseSection {
...
+		const data = fs.readFileSync(path, {encoding: 'utf8'});
+		const langs = JSON.parse(data);
+		const langList = this.generateSelectTemplate(langs, 'lang-menu');
+		langDiv.innerHTML += langList;
+		// langMenu is the select-option dropdown menu formed after executing the previous command
+		const langMenu: HTMLSelectElement = document.querySelector('.lang-menu');
+
+		// The next three lines set the selected language visible on the dropdown button
+		let langIndex = ConfigUtil.getConfigItem('languageIndex');
+		langIndex = (langIndex === null) ? 4 : langIndex;
```
andersk, line 420: Storing this as an index is extremely fragile—it means we can never reorder, insert, or delete languages without disturbing other languages.

#939

manavmehta: Agreed. Thanks @andersk :)

PR #913, opened 2020-04-01
downloadFiles: Fix issue of showing two Save As dialog box.

abhigyank:
Currently, there are two dialog boxes shown while downloading files (in Ubuntu). One by default behavior of electron and other by the dialog box for save as feature.
This PR fixes this issue by using electron's save as dialog box.

### DIFF
--- a/app/main/index.ts
+++ b/app/main/index.ts
@@ -5,7 +5,7 @@ import {setAutoLaunch} from './startup';
 import windowStateKeeper from 'electron-window-state';
 import path from 'path';
 import fs from 'fs';
-import electron, {app, ipcMain, session, dialog} from 'electron';
+import electron, {app, ipcMain, session} from 'electron';
 
 import * as AppMenu from './menu';
 import * as BadgeSettings from '../renderer/js/pages/preference/badge-settings';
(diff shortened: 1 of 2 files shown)


abhigyank:
@andersk Can you review and merge this before the new release? 
The current behavior results in broken download notifications.

manavmehta:
@abhigyank LGTM
1. there is no double dialogue issue on macOS and this PR doesn't break the initial behavior
2. it certainly does fix the double download dialogue on Ubuntu

Only one thing I'm concerned about is, _Ideally_, the download button in the lightbox should not ask the download location (show 'save as' dialogue) when the `Ask where to save files before downloading` option is disabled. But in my case, on macOS and Ubuntu, it still shows the dialogue asking the save location.

Should we configure it to behave that way?

@akashnimare @timabbott ??


I can post a few lines of behavior analysis on both systems if you like :)

timabbott:
@abhigyank can you post screenshots showing before/after?

abhigyank:
Turn the "Ask where to save file before Downloading" on in the settings.
Before - Two asks to choose place to download, and no complete notification. 
before

After Just one ask and complete notification - 
after

*Sorry for a video like this, for some reason my screen recorder was unable to record both save dialog boxes together. Hope this is still viable.

akashnimare:
What happens when you click on an already downloaded file?

abhigyank:
> What happens when you click on an already downloaded file?

The first pop-up shows a warning that the file already exists and you want to replace it. Second pop-up does nothing, as the download doesn't happen in the second popup folder.

In the PR, the only pop up shows the warning.

akashnimare:
@abhigyank please fix the merge conflicts and we can merge?

abhigyank:
@akashnimare Updated. Review once please!

akashnimare:
Works as expected on Ubuntu + macOS.

akashnimare:
Merged. Thanks, @abhigyank.

PR #922, opened 2020-04-07
Ubuntu: Fix for the issue when tray disappears on Locking/Unlocking Screen

### LINKED ISSUE #910: Tray Icon disappears on Screen Lock & Unlock
- **Operating System**:
  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

I'm on Ubuntu Focal (GNOME Shell 3.36.0). 

When running the desktop app on the latest master, the tray icon seems to disappear when I lock my screen and unlock it. Hard reloading the app seems to restore it. CZO discussion [here](https://chat.zulip.org/#narrow/stream/16-desktop/topic/Tray.20icon).

manavmehta:
**What's this PR do?**
Fixes: #910
Implement monitorTray() which when screen is locked and then unlocked, destroys and recreates the tray
Picks up from @dotenorio's comment on the thread
This is more of a workaround as we are relying on electron/chromium for this being an [upstream issue](https://github.com/ubuntu/gnome-shell-extension-appindicator/issues/75) of GNOME

  - [X] Linux/Ubuntu 18.04

### DIFF
--- a/app/renderer/js/tray.ts
+++ b/app/renderer/js/tray.ts
@@ -2,6 +2,8 @@ import { ipcRenderer, remote, WebviewTag, NativeImage } from 'electron';
 
 import path from 'path';
 import * as ConfigUtil from './utils/config-util';
+//	eslint-disable-next-line @typescript-eslint/no-var-requires
+const { exec } = require('child_process');
 
 const { Tray, Menu, nativeImage, BrowserWindow, nativeTheme } = remote;
 
@@ -183,6 +185,34 @@ ipcRenderer.on('tray', (_event: Event, arg: number): void => {
 	}
 });
 
+const monitorTray = (): void => {
+	if (process.platform !== 'linux') {
+		return;
+	}
+
+	/* Here we check if the screen has been locked and unlocked.
+		If the monitor returns SCREEN_UNLOCKED, then that has to be our case
+		And we destroy and recreate the tray
+	*/
+	const monitor = exec(`
+	  dbus-monitor --session "type='signal',interface='org.gnome.ScreenSaver'" |
+	  while read x; do
+		case "$x" in 
+		  *"boolean true"*) echo SCREEN_LOCKED;;
+		  *"boolean false"*) echo SCREEN_UNLOCKED;;  
+		esac
+	  done
+	`);
+
+	monitor.stdout.on('data', (data: any) => {
+		const out = data.toString().trim();
+		if (out === 'SCREEN_UNLOCKED') {
+			tray.destroy();
+			createTray();
+		}
+	});
+};
+
 function toggleTray(): void {
 	let state;
 	if (tray) {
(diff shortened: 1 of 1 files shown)


manavmehta:
@punchagan can you confirm if it works for you?

punchagan:
Works for me on Ubuntu Focal

akashnimare:
@abhigyank can you take a look? Also, curious on how other apps are handling this issue.

abhigyank:
I don't have the same issue as in #910 as it is for @punchagan and others.
I don't have the tray icon at all - https://github.com/electron/electron/issues/21445 so I can't verify this. But I have seen people saying this works in some threads.

app/renderer/js/tray.ts
```diff
@@ -183,6 +185,34 @@ ipcRenderer.on('tray', (_event: Event, arg: number): void => {
...
 
+const monitorTray = (): void => {
+	if (process.platform !== 'linux') {
+		return;
+	}
+
+	/* Here we check if the screen has been locked and unlocked.
+		If the monitor returns SCREEN_UNLOCKED, then that has to be our case
+		And we destroy and recreate the tray
+	*/
```
timabbott, line 196: Can you please add comments here linking to the upstream issue and mentioning that we can delete this workaround once upstream resolves the issue?  That will help maintainers a lot in the future.

app/renderer/js/tray.ts
```diff
@@ -183,6 +185,34 @@ ipcRenderer.on('tray', (_event: Event, arg: number): void => {
...
+	const monitor = exec(`
+	  dbus-monitor --session "type='signal',interface='org.gnome.ScreenSaver'" |
+	  while read x; do
+		case "$x" in 
+		  *"boolean true"*) echo SCREEN_LOCKED;;
+		  *"boolean false"*) echo SCREEN_UNLOCKED;;  
+		esac
+	  done
+	`);
```
andersk, line 205: * We should write the logic in JS instead of copying weird shell script snippets from StackOverflow.
* Instead of looking for `org.gnome.ScreenSaver` signals, I think we want to look for `org.freedesktop.DBus.NameAcquired("org.kde.StatusNotifierWatcher")`. (AppIndicator is based on KStatusNotifierItem.)
* Instead of forking a `dbus-monitor`, can we use a JS D-Bus library? Maybe [`dbus-native`](https://www.npmjs.com/package/dbus-native)?

app/renderer/js/tray.ts
manavmehta: Sure thing. Will be adding in the next commit

manavmehta:
@akashnimare @abhigyank Should I close this? As the issue is solved upstream in v5.3.0

akashnimare:
Can close this once @punchagan confirm the fix.

Issue #231, opened 2017-07-21
bug: app Window opens beyond screen boundaries


penCsharpener:
Had this twice now that Zulip opened on a PC but was not visible. Figured out that window positions where saved that cause the app to start beyond the screen resolution. 

- OS: `Windows10`
- reproduce: close Zulip, in `%appdata%\Zulip\config.json` make x more negative than width is positive

for non techy users this could be a killer because even if your uninstall Zulip the appdata folder remains and so does the unvalid config file.

fix: prevent Zulip desktop app to save a value for `x` that is larger negative than width positiv. Ideally you want to be able to see at least 100 pixels of the window in order to drag it back to the screen. 

![config.json](https://user-images.githubusercontent.com/26190934/28464877-29eec88c-6e27-11e7-8a7d-0be54be4b474.png)

akashnimare:
I think to handle these types of cases we should not allow negative values of x and y.

akashnimare:
@vbNETonIce can you pull master and check if it fixes this issue?

vbNETonIce:
will do but I won't be able tuntil Monday. I'll let you know then.

penCsharpener:
Yes, it's fixed but I realised something else now. i've got three screen, the centre one is the main display. The left screen has x: -... where Zulip app runs. This is more convenience really. But now zulip would always start up on my centre rather than left screen. 
Is it possible for you to check the maximum screen estate and base the maximum negative x on that. 
That would be low priority though.

akashnimare:
> Is it possible for you to check the maximum screen estate and base the maximum negative x on that. That would be low priority though.

Complicated but possible.

akashnimare:
Can you pull [restore-window](https://github.com/zulip/zulip-electron/tree/restore-window) and see if it works?

penCsharpener:
just did, when I move the app to the left screen and close it, it will still open on the centre screen (which is main the screen). It's like before.

penCsharpener:
ok, sorry... didn't checkout the correct branch. App now restores the last position even on the left screen. But the config.json always shows `x=0`. Is that intended? Do you save the window position somewhere else now?

akashnimare:
> Do you save the window position somewhere else now?

Yeah, it's in `window-state.json`. Can you cross check all the possible cases? I have a doubt that it might not work as intended in certain cases.

penCsharpener:
[...]

Issue #28, opened 2016-07-27
Window size and position between launches


GervaisdeM:
Electron always opens centered on screen and with what must be a default window size.

I like to make my window smaller with "Display user list on left side in narrow windows." and push it into the upper right of my screen. It does not save my settings between launches.

akashnimare:
@GervaisdeM We are not saving the last window settings/state anywhere. I think we should save it somewhere. I'm working on this.

akashnimare:
@GervaisdeM I have added this feature. Let me know if it's working properly. It might not work as expected on OSX.

GervaisdeM:
@akashnimare 
👍 
Works great!

akashnimare:
Yayy :tada:

akashnimare:
@GervaisdeM just wanted to cross-check if you faced this issue recently?

PR #924, opened 2020-04-07
Handle Reset options gracefully.

### LINKED ISSUE #903: “Factory Reset” and “Reset App Data” use two completely separate code paths
The **Tools → Factory Reset** menu option and the **Settings → General → Reset App Data** button use two completely separate code paths that do slightly different things. We should remove one of them or unify the two to use the same code.

manavmehta:
**What's this PR do?**
Fixes: #903
- Transport `Factory Reset` from menu bar to Settings page
- Remove `Reset App Data` option as it does not do something much useful

**Screenshots?**
Reset
P.S. The `?` is added in the messageBox heading, just after screenshot

  - [X] macOS Catalina 10.15.2
  - [X] Linux/Ubuntu 18.04

### DIFF
--- a/app/renderer/js/pages/preference/general-section.ts
+++ b/app/renderer/js/pages/preference/general-section.ts
@@ -138,12 +138,12 @@ export default class GeneralSection extends BaseSection {
 						<div class="setting-control"></div>
 					</div>
 				</div>
-				<div class="title">${t.__('Reset Application Data')}</div>
+				<div class="title">${t.__('Factory Reset Data')}</div>
                 <div class="settings-card">
-					<div class="setting-row" id="resetdata-option">
-						<div class="setting-description">${t.__('This will delete all application data including all added accounts and preferences')}
+					<div class="setting-row" id="factory-reset-option">
+						<div class="setting-description">${t.__('Reset the application, thus deleting all the connected organizations, accounts, and certificates.')}
 						</div>
-						<button class="reset-data-button red w-150">${t.__('Reset App Data')}</button>
+						<button class="factory-reset-button red w-150">${t.__('Factory Reset')}</button>
 					</div>
 				</div>
 			</div>
@@ -159,7 +159,7 @@ export default class GeneralSection extends BaseSection {
 		this.betaUpdateOption();
 		this.updateSidebarOption();
 		this.updateStartAtLoginOption();
-		this.updateResetDataOption();
+		this.factoryReset();
 		this.showDesktopNotification();
 		this.enableSpellchecker();
 		this.minimizeOnStart();
@@ -372,23 +372,6 @@ export default class GeneralSection extends BaseSection {
 		});
 	}
 
-	async clearAppDataDialog(): Promise<void> {
-		const clearAppDataMessage = 'By clicking proceed you will be removing all added accounts and preferences from Zulip. When the application restarts, it will be as if you are starting Zulip for the first time.';
-		const getAppPath = path.join(app.getPath('appData'), app.name);
-
-		const {response} = await dialog.showMessageBox({
-			type: 'warning',
-			buttons: ['YES', 'NO'],
-			defaultId: 0,
-			message: 'Are you sure',
-			detail: clearAppDataMessage
-		});
-		if (response === 0) {
-			await fs.remove(getAppPath);
-			ipcRenderer.send('forward-message', 'hard-reload');
-		}
-	}
-
 	async customCssDialog(): Promise<void> {
 		const showDialogOptions: OpenDialogOptions = {
 			title: 'Select file',
@@ -401,13 +384,6 @@ export default class GeneralSection extends BaseSection {
 			ConfigUtil.setConfigItem('customCSS', filePaths[0]);
 			ipcRenderer.send('forward-message', 'hard-reload');
 		}
-	}
-
-	updateResetDataOption(): void {
-		const resetDataButton = document.querySelector('#resetdata-option .reset-data-button');
-		resetDataButton.addEventListener('click', async () => {
-			await this.clearAppDataDialog();
-		});
 	}
 
 	setLocale(): void {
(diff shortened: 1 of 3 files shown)


app/renderer/js/pages/preference/general-section.ts
```diff
@@ -130,11 +138,20 @@ export default class GeneralSection extends BaseSection {
 				<div class="title">${t.__('Reset Application Data')}</div>
                 <div class="settings-card">
 					<div class="setting-row" id="resetdata-option">
-						<div class="setting-description">${t.__('This will delete all application data including all added accounts and preferences')}
+						<div class="setting-description">${t.__('This will reset all application settings to default preserving added organisations and accounts')}
```
timabbott: These strings aren't great; we use American English and the "This will" is weird.  How about:

"Reset all desktop application settings to defaults, preserving connected organizations and accounts"

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -130,11 +138,20 @@ export default class GeneralSection extends BaseSection {
...
 						</div>
 						<button class="reset-data-button red w-150">${t.__('Reset App Data')}</button>
 					</div>
 				</div>
+				<div class="title">${t.__('Factory Reset')}</div>
+                <div class="settings-card">
+					<div class="setting-row" id="factory-reset-option">
+						<div class="setting-description">${t.__('This will delete all connected organisations, accounts, certificates and preferences')}
```
timabbott: "Reset all desktop application settings to defaults, deleting connected organizations and accounts"

app/renderer/js/pages/preference/general-section.ts
manavmehta: Sure thing. Please have a look :)

manavmehta:
@akashnimare Can you please have a look?

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -389,11 +390,78 @@ export default class GeneralSection extends BaseSection {
...
+		resetAppButton.addEventListener('click', () => {
+			this.resetAppSettings();
+		});
+	}
+
+	async resetAppSettings(): Promise<void> {
+		const resetMessage = 'Selecting YES will reset all the settings to default without removing any connected organizations and preferences from Zulip.';
+		// We save the Application's settings/configurations in following files
+		// Preferences file is able to reset zoom levels but requires restart, not hard reload
+		const settingFiles = ['config/settings.json', 'config/window-state.json', 'Partitions/webviewsession/Preferences'];
```
manavmehta: @timabbott @akashnimare any suggestions on what else can be done with `Reset App Data`?

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -389,11 +390,78 @@ export default class GeneralSection extends BaseSection {
...
+	/* Reset App Data */
+	resetAppData(): void {
+		const resetAppButton = document.querySelector('#resetdata-option .reset-data-button');
+		resetAppButton.addEventListener('click', () => {
+			this.resetAppSettings();
+		});
+	}
+
+	async resetAppSettings(): Promise<void> {
+		const resetMessage = 'Selecting YES will reset all the settings to default without removing any connected organizations and preferences from Zulip.';
```
abhigyank: We are removing preferences here. So should no write that preferences will not be removed.

manavmehta: My bad. That should be organisations and accounts

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -389,11 +390,78 @@ export default class GeneralSection extends BaseSection {
...
+	factoryReset(): void {
+		const resetDataButton = document.querySelector('#factory-reset-option .factory-reset-button');
 		resetDataButton.addEventListener('click', () => {
-			this.clearAppDataDialog();
+			this.factoryResetSettings();
+		});
+	}
+
+	async factoryResetSettings(): Promise<void> {
+		const clearAppDataMessage = 'Selecting YES will remove all organisations, certificates, accounts and preferences from Zulip.\nWhen the application restarts, it will be as if you are starting Zulip for the first time.';
```
timabbott: Please spell "organizations"; we use American english spelling.

app/renderer/js/pages/preference/general-section.ts
timabbott: Can you show what this looks like? I don't understand why you have "Selecting YES will ..." in the label for this setting.  We don't have similar extra verbosity in other settings, and I just specifically suggested removing such extra verbosity yesterday.

manavmehta: Earlier messages were: 
_By proceeding you will be removing all connected organizations and preferences from Zulip._
and
_By clicking proceed you will be removing all added accounts and preferences from Zulip. When the application restarts, it will be as if you are starting Zulip for the first time._

Should we keep a simple "Do you want to proceed?" or "Are you sure you want to proceed?"

manavmehta: Screenshot 2020-04-09 at 1 44 27 AM
This is how it, at present in this PR, looks like

I totally agree on the part: 
> I just specifically suggested removing such extra verbosity yesterday.
>
Just left this as we already displayed these messages

app/renderer/js/pages/preference/general-section.ts
manavmehta: Will push in next commit :)

app/renderer/js/pages/preference/general-section.ts
manavmehta: @timabbott ping

manavmehta:
@akashnimare @timabbott please have a look if it seems concise

.
**Reset App Data**
reset-data

.
**Factory Reset**
factory-reset

timabbott:
The messages in the popups messages are still kinda confusing.  E.g. "you are starting Zulip for the first time" makes it sound like it might delete your message history.  The first part but ending with "... as if you just downloaded the Zulip desktop app for the first time" would be clear on that front.

And for the first one, I think "Reset all the Zulip desktop app's configuration settings to the defaults." would be clear.

Otherwise lgtm.

akashnimare:
I think "Factory Reset" setting should go somewhere in the Menu. Two settings for wiping out the app data doesn't look good UX wise and it may confuse users. 

On second thought we should rethink our decision of adding the "Reset app data" in the setting. As far as I remember the main reason behind adding this was to handle the case where users weren't able to restore the window position of the app and I think we have not received any issue related to the window restoring feature. So I think the best way would be -

* Remove the "Reset app data" setting completely
* Keep the "Factory reset" in the setting option

open for discussion.

cc - @timabbott @abhigyank

manavmehta:
@akashnimare as you said, **One option indeed looks better**
Regarding
> handle the case where users weren't able to restore the window position of the app
>we have not received any issue related to the window restoring feature
>
Issue #28 is solved but I can;t check for #231 as I don;t have an external display. If it's solved and `window-state` does the necessary, then we might keep only `Factory Reset`
If #231 still persists, we may solve that separately and then only merge this.
@timabbott @abhigyank if it sounds fair ?

akashnimare:
Folks have not reported the window position error in the last two years so we can assume it works well.

manavmehta:
@akashnimare Should we wait for their reply on the thread and then go on with

- Removing Reset App Data
- Keep Factory Reset as the only reset

akashnimare:
Fix the merge conflicts. I think we can remove the one which deletes the app data. It doesn't seem to serve much of a purpose. So for now, let's keep the factory one. We can revisit this later based on more feedbacks etc.

manavmehta:
For the record, screwed up big time on this one.
I'll be closing this PR and opening a new cleaner one.

akashnimare:
LGTM. Merging.

akashnimare:
@manavmehta as a follow-up, can you send a PR updating the translating file since some strings are being updated in this PR.

manavmehta:
@akashnimare Sure. The context-menu would be having new strings too. So I'll look into them also at once :)

PR #929, opened 2020-04-12
Generate User-Agent when app starts, thus avoiding storing it on disk

### LINKED ISSUE #921: User-Agent is incorrectly cached in settings.json, with stale value remembered on upgrades
We seem to be saving and restoring (part of) the User-Agent in `settings.json`, leading to self-contradictory user-agents like

`ZulipElectron/5.0.0 Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_3) AppleWebKit/537.36 (KHTML, like Gecko) Zulip/4.0.3 Chrome/66.0.3359.181 Electron/3.1.10 Safari/537.36`

(Note the `ZulipElectron/5.0.0` and `Zulip/4.0.3`.)

This setting seems to have been added in #818. There should be no reason to persist the User-Agent to disk at all. Why are we doing that?

Cc @kanishk98

manavmehta:
Fixes: #921 

**What's this PR do?**
Fetch and send UA from main process and use ipc call to catch it in renderer process: system-util.ts and use it elsewhere in the app as required, avoiding storing it on disk.

Please refer #921 

  - [ ] Linux/Ubuntu
  - [X] macOS Catalina 10.15.2

### DIFF
--- a/app/main/index.ts
+++ b/app/main/index.ts
@@ -158,6 +158,9 @@ app.on('activate', () => {
 });
 
 app.on('ready', () => {
+	const ses = session.fromPartition('persist:webviewsession');
+	ses.setUserAgent(`ZulipElectron/${app.getVersion()} ${ses.getUserAgent()}`);
+
 	AppMenu.setMenu({
 		tabs: []
 	});
@@ -190,10 +193,10 @@ app.on('ready', () => {
 		} else {
 			mainWindow.show();
 		}
-		if (!ConfigUtil.isConfigItemExists('userAgent')) {
-			const userAgent = session.fromPartition('webview:persistsession').getUserAgent();
-			ConfigUtil.setConfigItem('userAgent', userAgent);
-		}
+	});
+
+	ipcMain.on('fetch-user-agent', event => {
+		event.returnValue = session.fromPartition('persist:webviewsession').getUserAgent();
 	});
 
 	page.once('did-frame-finish-load', () => {
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -140,6 +140,7 @@ class ServerManagerView {
 		await this.loadProxy();
 		this.initDefaultSettings();
 		this.initSidebar();
+		this.removeUAfromDisk();
 		if (EnterpriseUtil.configFile) {
 			this.initPresetOrgs();
 		}
@@ -237,6 +238,12 @@ class ServerManagerView {
 	initSidebar(): void {
 		const showSidebar = ConfigUtil.getConfigItem('showSidebar', true);
 		this.toggleSidebar(showSidebar);
+	}
+
+	// Remove the stale UA string from the disk if the app is not freshly
+	// installed.  This should be removed in a further release.
+	removeUAfromDisk(): void {
+		ConfigUtil.removeConfigItem('userAgent');
 	}
 
 	async queueDomain(domain: any): Promise<boolean> {
--- a/app/renderer/js/utils/system-util.ts
+++ b/app/renderer/js/utils/system-util.ts
@@ -1,9 +1,6 @@
-import { remote } from 'electron';
+import { ipcRenderer } from 'electron';
 
 import os from 'os';
-import * as ConfigUtil from './config-util';
-
-const { app } = remote;
 
 export const connectivityERR: string[] = [
 	'ERR_INTERNET_DISCONNECTED',
@@ -14,7 +11,7 @@ export const connectivityERR: string[] = [
 	'ERR_NETWORK_CHANGED'
 ];
 
-let userAgent: string | null = null;
+const userAgent = ipcRenderer.sendSync('fetch-user-agent');
 
 export function getOS(): string {
 	const platform = os.platform();
@@ -32,14 +29,6 @@ export function getOS(): string {
 		return '';
 	}
 }
-
-export function setUserAgent(webViewUserAgent: string): void {
-	userAgent = `ZulipElectron/${app.getVersion()} ${webViewUserAgent}`;
-}
-
-export function getUserAgent(): string | null {
-	if (!userAgent) {
-		setUserAgent(ConfigUtil.getConfigItem('userAgent', null));
-	}
+export function getUserAgent(): string {
 	return userAgent;
 }
--- a/app/renderer/js/components/webview.ts
+++ b/app/renderer/js/components/webview.ts
@@ -149,12 +149,6 @@ export default class WebView extends BaseComponent {
 			if (!isSettingPage) {
 				this.props.switchLoading(true, this.props.url);
 			}
-			let userAgent = SystemUtil.getUserAgent();
-			if (!userAgent) {
-				SystemUtil.setUserAgent(this.$el.getUserAgent());
-				userAgent = SystemUtil.getUserAgent();
-			}
-			this.$el.setUserAgent(userAgent);
 		});
 
 		this.$el.addEventListener('did-stop-loading', () => {


manavmehta:
@andersk @timabbott Can you please have a look? If this feels right, we can refine it :)

timabbott:
This certainly looks plausible.  It's probably a good idea to test this against a Zulip development server configured with a Python `print()` statement to print the User-Agent it received in both `GET /server_settings` and `GET /events` to confirm it's sending the right data in both contexts.

app/renderer/js/utils/system-util.ts
```diff
@@ -16,6 +14,19 @@ export const connectivityERR: string[] = [
...
+ipcRenderer.on('user-agent', (event, arg) => {
+	userAgent = arg;
+	if (!userAgent) {
+		console.log('User-Agent found null');
+	}
+});
+
+// This is to remove the UA setting from the disk if the app is not freshly installed
+// We should remove this in a further release
+if (isConfigItemExists('userAgent')) {
```
akashnimare: @manavmehta we should put this where we initialize other settings?

app/renderer/js/utils/system-util.ts
```diff
@@ -32,14 +43,9 @@ export function getOS(): string {
 		return '';
 	}
 }
-
-export function setUserAgent(webViewUserAgent: string): void {
```
akashnimare, line 36: We need to use this with the session's UA. We are using `ZulipElectron` to detect the desktop app. Check out -
https://github.com/zulip/zulip/search?q=ZulipElectron&unscoped_q=ZulipElectron

cc @timabbott

akashnimare (changes requested):

app/renderer/js/utils/system-util.ts
manavmehta: @akashnimare Maybe we should refine `test_insecure_desktop_app(self)` itself? (wrt desktop app only)
As looking at L101, isn't this exactly what @andersk pointed out? As this forces us to append `ZulipElectron/${app.getVersion()}` to the UA.

https://github.com/zulip/zulip/blob/17723b16177663a558c03ef6a41a926163ce4241/zerver/tests/test_compatibility.py#L101

akashnimare:
@manavmehta can you test this against a zulip test server and note down the UA in the following cases -
a) When user add a new server
b) API requests from the renderer process

manavmehta:
null_UA_on_adding_server
The UA is correct in /server_settings and /events
But it is sent as null (or the default value) while adding a server, thereafter, it is correct in renderer API calls.

Reason: The new-server-form calls domain-util which calls system-utils afresh.

Workaround as suggested by @abhigyank  ebb1238
Change the default value of UA
This would work because:
1. This is the only part we use to recognize Desktop app
2. A user adds server least often.

@timabbott @andersk any fixes in your mind or should we go with this?
(documenting this a TODO so that someone could come up with a fix)

andersk:
A more robust approach would be for the renderer process to ask the main process for the user-agent with `ipcRenderer.invoke` (caching the resulting `Promise`) or `ipcRenderer.sendSync` (caching the resulting string).

manavmehta:
@andersk That was a great suggestion.
It is ready for review and merge from my side.
cc: @akashnimare 

P.S. If you want, I can send you the log where I printed a bunch of UAs received by the test server

app/renderer/js/utils/system-util.ts
```diff
@@ -32,14 +34,9 @@ export function getOS(): string {
...
-	userAgent = `ZulipElectron/${app.getVersion()} ${webViewUserAgent}`;
-}
-
 export function getUserAgent(): string | null {
 	if (!userAgent) {
-		setUserAgent(ConfigUtil.getConfigItem('userAgent', null));
+		console.log('userAgent found null');
```
kanishk98: Hmm, I don't think using a `console.log()` statement is a great idea - we're already logging the UA sent on the server side, right?
Maybe we can use a Sentry call here instead if we do client-side logs.

app/renderer/js/main.ts
```diff
@@ -239,6 +240,14 @@ class ServerManagerView {
 		this.toggleSidebar(showSidebar);
 	}
 
+	/*  Removes the stale UA string from the disk if the app is not freshly installed
+		We should remove this in a further release	*/
+	removeUAfromDisk(): void {
+		if (ConfigUtil.isConfigItemExists('userAgent')) {
```
kanishk98: Do we need this conditional? I believe the db handles that for us, doesn't it?

app/main/index.ts
```diff
@@ -185,15 +185,18 @@ app.on('ready', () => {
...
 		}
-		if (!ConfigUtil.isConfigItemExists('userAgent')) {
-			const userAgent = session.fromPartition('webview:persistsession').getUserAgent();
-			ConfigUtil.setConfigItem('userAgent', userAgent);
-		}
+	});
+
+	ipcMain.handle('fetch-user-agent', async () => {
+		const userAgent = session.fromPartition('webview:persistsession').getUserAgent();
+		return userAgent;
```
kanishk98: Sorry, another nitpick, would you mind just returning the result of the expression directly?
I don't see the need to declare a `const` here.

kanishk98 (commented):
Alright, this looks good to me, except for a few comments I've left that you may want to consider. 
Great work, @manavmehta!

app/main/index.ts
manavmehta: Sure

app/renderer/js/main.ts
manavmehta: Probably it does. I was a little unsure about that. But you confirmed it for me. Thanks!

app/renderer/js/utils/system-util.ts
manavmehta: @kanishk98 I've changed the process from console.log() to logger.log()
DO let me know if there is a need to make a Sentry call :)
@akashnimare what would you suggest?

manavmehta:
@kanishk98 Thanks. Will push the changes ASAP as the logging part is confirmed :)

andersk:
The goal is to avoid the race condition where we see `ZulipElectron/5.0.0 ` for some time at startup instead of the full `ZulipElectron/5.0.0 Mozilla/5.0 blah/blah blah` string. Your code still doesn’t do that.

Either use `ipcRenderer.invoke` this way:

```ts
const userAgentPromise: Promise<str> = ipcRenderer.invoke('fetch-user-agent');

export async function getUserAgent(): string {
    return `ZulipElectron/${app.getVersion()} ${await userAgentPromise}`;
}
```

(which requires adjusting its callers to be `async` as well), or use `ipcRenderer.sendSync`.

manavmehta:
@andersk I tested 37c420a against Zulip test server and logging both. Couldn't find a single instance of the race condition.

But since theoretically, the condition is possible, I think we may go with `ipcRenderer.sendSync`
Tested it and it works fine. Plus, since it fetches UA synchronously, I don't think there's a possibility for the race condition.

Though it adds a delay, which is probably lesser than a millisecond and better than the race condition anyway.
Thanks :)

app/main/index.ts
```diff
@@ -185,15 +185,17 @@ app.on('ready', () => {
 	const page = mainWindow.webContents;
 
 	page.on('dom-ready', () => {
+		// On dom-ready, fetch user agent and send to renderer processes.
+		// As of now, only system-util.ts catches it.
```
andersk: This comment is wrong now since we don’t wait for `dom-ready` (and I don’t think there’s a reason to wait).

app/renderer/js/utils/system-util.ts
```diff
@@ -32,14 +36,9 @@ export function getOS(): string {
...
-
-export function setUserAgent(webViewUserAgent: string): void {
-	userAgent = `ZulipElectron/${app.getVersion()} ${webViewUserAgent}`;
-}
-
 export function getUserAgent(): string | null {
 	if (!userAgent) {
-		setUserAgent(ConfigUtil.getConfigItem('userAgent', null));
+		logger.log('userAgent found null');
 	}
```
andersk: Remove this `if` branch, the `null` from the return type, and the logger. No reason to have a bunch of code for an impossible case.

manavmehta: My bad. Should've removed this in prev commit.
Refactored

andersk:
I edited this to use `ses.setUserAgent` to set the User-Agent once for the entire session rather than individually for each webview. I also squashed all the commits together.

PR #861, opened 2020-01-13
macOS: Add colorless tray icon for macOS.

akashnimare:
---

**What's this PR do?**
Fixes: #825. 

**Any background context you want to provide?**

**Screenshots?**

**You have tested this PR on:**
  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [ ] macOS

PR #894, opened 2020-03-07
macOS: Update dock icon.

adityamudgil2505:
Added inner padding (transparent) of 30px to the icon having size 512x512.

We are getting too large size because other application logos are giving inner padding of 30px whereas Zulip application icon is giving 0px inner padding.
Detailed explanation in term of image

After giving inner padding of 30px, now our application logo is looking perfect.

Working image

**Note:** The actual size of the icon after giving padding will remain the same that is 512x512.
Solve Issue #787 

**You have tested this PR on:**
  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [x] macOS

PR #941, opened 2020-04-24
macOS: Use electron API to get dark tray icon for the light theme.

manavmehta:
Context: #861 

**What's this PR do?**
We made a change to make the tray icon look flushed and coherent for the dark theme
A corresponding dark icon would do the necessary for the light theme
Electron makes it possible by using [Template Image](https://www.electronjs.org/docs/api/native-image#template-image)

**Screenshots?**
Presently we have this
old light theme
dark theme

This commit makes this change
new light theme
dark theme

  - [X] macOS Catalina 10.15.2


manavmehta:
@akashnimare if you could test it on your mac

akashnimare:
@manavmehta -

https://cl.ly/4d37215646da

Check if electron handles this by default otherwise we need to add IPC when user updates the theme.

manavmehta:
@akashnimare please have a look.
Electron is a beauty 😄

akashnimare:
@manavmehta please fix merge conflicts.

manavmehta:
@akashnimare fixed them.
Also was [this comment](https://github.com/zulip/zulip-desktop/pull/894#issuecomment-622369518) intended to be here?

akashnimare:
Did you update the tray image? I think it was updated in #894.

manavmehta:
@akashnimare Afaik, #894 only updates the dock icon, not the tray icon
What this PR does: electron uses Template Image which consists of a black and an alpha channel and can adjust its appearance based on the theme

Quoting from [electron docs](https://www.electronjs.org/docs/api/native-image#template-image) : 
>The most common case is to use template images for a menu bar icon, so it can adapt to both light and dark menu bars.
>
So, we don't need two images and any ipc calls to change the image, electron does this by itself
I just renamed the tray image to take advantage of that and we won't need the green icon - coherence

akashnimare:
@manavmehta fix the merge conflicts and we can merge this. As a follow-up, we should update our tray image because the white padding is clearly visible on macOS.

manavmehta:
@akashnimare resolved.
Also, would it be good to open an issue for this? (as there is less chance of missing out on an issue if it's listed)
>  we should update our tray image because the white padding is clearly visible on macOS.
>

app/renderer/js/tray.ts
```diff
@@ -3,14 +3,13 @@ import {ipcRenderer, remote, WebviewTag, NativeImage} from 'electron';
 import path from 'path';
 import * as ConfigUtil from './utils/config-util';
 
-const {Tray, Menu, nativeImage, BrowserWindow, nativeTheme} = remote;
+const {Tray, Menu, nativeImage, BrowserWindow} = remote;
 
 let tray: Electron.Tray;
 
 // Get the theme on macOS
-const theme = nativeTheme.shouldUseDarkColors ? 'dark' : 'light';
```
andersk: The comment no longer applies and should be deleted too.

app/renderer/js/tray.ts
```diff
@@ -3,14 +3,13 @@ import {ipcRenderer, remote, WebviewTag, NativeImage} from 'electron';
...
 
-const ICON_DIR = process.platform === 'darwin' ? `../../resources/tray/${theme}` : '../../resources/tray';
+const ICON_DIR = process.platform === 'darwin' ? '../../resources/tray/macOS' : '../../resources/tray';
```
andersk: There’s no longer a reason to have the macOS icon in a different directory, right? This could just always be `../../resources/tray`.

manavmehta: I had the thought but preserved the extra directory so that the other contributors could comprehend what is Template Image and where it comes into play. (though we do have osx in the name)
If you feel I might shift it one dir up?

andersk: Yeah, I don’t see the extra directory doing anything to improve comprehensibility. Use the same directory to simplify the code. Feel free to rename the files to `tray-macOS-Template.png`, etc. or something.

kanishk98:
>  As a follow-up, we should update our tray image because the white padding is clearly visible on macOS.

Hmm, just going through this issue and I'm not sure I follow this. @akashnimare could you clarify what you mean by white padding?

akashnimare:
LGTM on macOS.

akashnimare:
Merged. Thanks, everyone!

PR #943, opened 2020-04-25
electron-bridge: Add decrypt_clipboard helper (for browser-based social login)

andersk:
This one helper allows us to implement browser-based social login entirely on the server side. See zulip/zulip#14747.

### DIFF
--- a/app/renderer/js/electron-bridge.ts
+++ b/app/renderer/js/electron-bridge.ts
@@ -2,6 +2,7 @@ import {ipcRenderer} from 'electron';
 
 import {EventEmitter} from 'events';
 
+import {ClipboardDecrypter} from './clipboard-decrypter';
 import {NotificationData, newNotification} from './notification';
 
 type ListenerType = ((...args: any[]) => void);
@@ -46,6 +47,9 @@ class ElectronBridge extends EventEmitter {
 	set_send_notification_reply_message_supported = (value: boolean): void => {
 		this.send_notification_reply_message_supported = value;
 	};
+
+	decrypt_clipboard = (version: number): ClipboardDecrypter =>
+		new ClipboardDecrypter(version);
 }
 
 const electron_bridge = new ElectronBridge();
--- a/app/renderer/js/clipboard-decrypter.ts
+++ b/app/renderer/js/clipboard-decrypter.ts
@@ -1,0 +1,82 @@
+import {clipboard} from 'electron';
+import crypto from 'crypto';
+
+// This helper is exposed via electron_bridge for use in the social
+// login flow.
+//
+// It consists of a key and a promised token.  The in-app page sends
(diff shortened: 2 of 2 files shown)


app/renderer/js/electron-bridge.ts
```diff
@@ -46,6 +47,28 @@ class ElectronBridge extends EventEmitter {
...
+				try {
+					const ciphertext = Buffer.from(clipboard.readText(), 'hex');
+					const decipher = crypto.createDecipheriv(
+						'aes-256-gcm',
+						key,
+						ciphertext.slice(0, 12),
+						{ authTagLength: 16 }
+					);
+					decipher.setAuthTag(ciphertext.slice(-16));
+					return decipher.update(ciphertext.slice(12, -16), undefined, 'utf8') + decipher.final('utf8');
```
timabbott: I would probably extract variables for the `ciphertext.slice()` sections all the start for readability, e.g. `const iv = ciphertext.slice(0, 12);`, etc.

app/renderer/js/electron-bridge.ts
```diff
@@ -46,6 +47,28 @@ class ElectronBridge extends EventEmitter {
 	set_send_notification_reply_message_supported = (value: boolean): void => {
 		this.send_notification_reply_message_supported = value;
 	};
+
+	decrypt_clipboard = (): { key: string; check(): string | null } => {
+		const key = crypto.randomBytes(32);
```
timabbott: I think probably it's worth a block comment here explaining the interface of this, namely that it provides a randomly generated key as well as a function to decrypt content encrypted using that key; and the encryption algorithm used by the server needs to be the same as what is used here.  

I sorta wonder if we should have this present a third parameter, `algorithm_version`, that is an int that gets passed on to the server; this would be setup for if we decide to change the algorithm in the future, the server could know how the app is expecting us to encrypt.  I suppose we could always add this later with "didn't send" meaning v0, but it'd be pretty cheap to just have an `algorithm_version: 1` here so that the server-side code can assume that field is always present.

(Even if it's ignored for now, the server-side JS could always change to use it)

app/renderer/js/electron-bridge.ts
andersk: Done.

app/renderer/js/electron-bridge.ts
andersk: Done.

andersk:
Also moved this helper to a new file, and changed the polling interface to a promise interface in case we want to do something more clever than `setInterval(…, 1000)` in the future.

timabbott:
This lgtm.

andersk:
Updated to poll the clipboard only when the app has focus.

timabbott:
The polling logic lgtm; feel free to merge.

PR #944, opened 2020-04-26
Upgrade xo to 0.30, and reenable most disabled lints

andersk:


timabbott:
This PR lgtm.  A lot of these are great cleanups.

PR #950, opened 2020-05-02
help: Update Help Center URL.

### LINKED ISSUE #948: Help center link incorrect in 5.1.0
The URL associated with the Zulip "Help Center" seems to have lost some information between 5.0.0 and 5.1.0 on macOS.
---
- **Operating System**:
  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [X] macOS
- **Clear steps to reproduce the issue**: Click the "Help Center" link in the "Help" menu. 
- **Relevant error messages and/or screenshots**: Incorrectly sends browser to non-existent "/help" instead of "https://zulipchat.com/help/" as in previous versions.

manavmehta:
Please refer: #948 
Earlier the menu option used to redirect to *.zulipchat.com/help which is already reachable from the webapp.
This will enable the desktop app to redirect only to zulipchat.com/help instead of the help page of the corresponding organization.

  - [X] macOS Catalina 10.15.2

Tested on 4 added organizations, About Zulip page, Add New Server page
Fixes: #948

### DIFF
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -861,7 +861,7 @@ class ServerManagerView {
 
 		ipcRenderer.on('open-help', async () => {
 			// Open help page of current active server
-			await LinkUtil.openBrowser(new URL('/help', this.getCurrentActiveServer()));
+			await LinkUtil.openBrowser(new URL('https://zulipchat.com/help/'));
 		});
 
 		ipcRenderer.on('reload-viewer', this.reloadView.bind(this, this.tabs[this.activeTabIndex].props.index));


manavmehta:
@timabbott though I've hardcoded the URL as you said in the thread, can you please mention why is it not a good approach to let it redirect to (orgname).zulipchat.com/help/ ?
P.S. We may fix only the corner case mentioned in the thread - About page.
cc: @akashnimare

akashnimare:
As suggested in that thread https://zulipchat.com/help is good enough for that case.

PR #957, opened 2020-05-12
transifex: Setup Transifex for better translation process.

manavmehta:
**What's this PR do?**

1. add .tx/config to configure communication with Transifex CLI
2. add how-to-doc for translating zulip-desktop
3. add tx-pull executable script for pulling translations from Transifex
4. update supported-locales.json
5. remove now-inessential scripts and translations

and other necessary changes

  - [X] macOS
  - [ ] Ubuntu 20.04


docs/howto/translations.md
```diff
@@ -0,0 +1,95 @@
...
+Sometimes when downloading translated strings we get a file for a new
+language.  This happens when we've opened up a new language for people
+to contribute translations into in the Zulip project on Transifex,
+which we do when someone expresses interest in contributing them.
+
+Each messages file in `translations/static/` should be reflected in
+three boring, more-or-less mechanical lists:
+ * `translations/flow-typed/translations.js`
+ * `translations/i18n/locale.js`
+ * `translations/i18n/messages.js`
```
manavmehta: These might or might not be required as-it-is and are prone to changes

timabbott:
The slug in `.tx/config` should be: `zulip.desktopjson`

Otherwise, the Transifex integration wfm.

timabbott:
I fixed that slug, ran `tools/tx-pull`, and pushed back here following that change.

manavmehta:
@akashnimare FYI.
May we look into removing the inessential modules [here](https://github.com/zulip/zulip-desktop/tree/master/tools/locale-helper)?

akashnimare:
@manavmehta yes, please remove that since we won't be using that module.

manavmehta:
@akashnimare Removed the locale-helper from 8e04920 which is now not needed.
I've updated the `supported-locales` module :)
I've used native scripts for language names as we use them on the webapp. If you feel we could go for English names only.
cc: @timabbott

akashnimare:
`tx-pull` works fine. @andersk can you give a quick review so that we can merge it?

akashnimare:
This is GTM after we pull a few strings which are there on Trasifex.

akashnimare:
Merging. Will update the strings on tx.

PR #928, opened 2020-04-09
spellchecker: Use Electron 8 built-in spellchecker

andersk:

PR #967, opened 2020-05-23
spellchecker: Use Electron 8 built-in spellchecker. 

### LINKED ISSUE #504: Please add usage of default context menu on MacOS
The desktop app currently only shows a basic context menu if you right-click a word:

screen shot 2018-06-21 at 14 56 11

It would be very helpful if the full MacOS context menu would be supported like in this screenshot:

screen shot 2018-06-21 at 14 58 09

Especially the "Services" section is important as this will allow you to use selections with AppleScript services e.g. for automatically looking up selected numbers in Bugzilla, Jira or other tools of choice.

---
- **Operating System**:
  - [ ] Windows
  - [ ] Linux/Ubutnu
  - [X] macOS

manavmehta:
**What's this PR do?**
#928 takes away with it the context menu as so this PR implements a custom context menu
Inherits commit from #928 as it was needed
**Screenshots?**

  - [X] macOS Catalina 10.15.2
  - [x] Linux/Ubuntu
  - [ ] Windows

Also fixes: #504


akashnimare:
Just had a quick run - this works well.

package.json
```diff
@@ -148,15 +147,14 @@
     "backoff": "^2.5.0",
     "electron-is-dev": "^1.2.0",
     "electron-log": "^4.1.2",
-    "electron-spellchecker": "^2.2.1",
     "electron-updater": "^4.3.1",
     "electron-window-state": "^5.0.3",
     "escape-html": "^1.0.3",
     "fs-extra": "^9.0.0",
     "i18n": "^0.9.1",
+    "nan": "^2.14.0",
```
abhigyank, line 162: Any need of this?

manavmehta: Inherited from #928 I'll check the package once

app/renderer/js/components/context-menu.ts
```diff
@@ -0,0 +1,112 @@
...
+		});
+	};
+
+	const ctx = getContextMenu();
+	const menu = Menu.buildFromTemplate(ctx);
+
+	if (props.misspelledWord) {
+		if (props.dictionarySuggestions.length > 0) {
+			Object.values(props.dictionarySuggestions).forEach(value => {
+				menu.insert(0, makeSuggestion(value as string));
```
manavmehta: bug: Most relevant suggestion is added at the bottom

app/renderer/js/components/context-menu.ts
```diff
@@ -0,0 +1,112 @@
...
+
+	const getContextMenu = (): Electron.MenuItemConstructorOptions[] => {
+		return [{
+			label: t.__('Add to Dictionary'),
+			visible: props.isEditable && isText && props.misspelledWord,
+			click(_item) {
+				webContents.session.addWordToSpellCheckerDictionary(props.misspelledWord);
+			}
+		}, {
+			type: 'separator'
```
manavmehta: bug: Separator visible on ubuntu even if word is not misspelled.
`visible` attribute doesn't seem to work here too.

manavmehta (commented):
2 bugs to keep updated

package.json
andersk: It’s required to work around CharlieHess/node-mac-notifier#19.

app/renderer/js/components/context-menu.ts
abhigyank: Can we first store everything from `props.dictionarySuggestions` to an array, reverse that and then insert into `menu`? That may solve this bug.

manavmehta: Yes. That was my first choice. But I was going to push this to save some minor space  
```suggestion
			let suggestionIndex = 0;
			Object.values(props.dictionarySuggestions).forEach(value => {
				menu.insert(suggestionIndex++, makeSuggestion(value as string));
```

app/renderer/js/components/context-menu.ts
```diff
@@ -0,0 +1,102 @@
...
+import * as t from '../utils/translation-util';
+const {clipboard, Menu, MenuItem} = remote;
+
+export const contextMenu = (webview: WebView, event: any) => {
+	const webContents = webview.$el.getWebContents();
+	const props = event.params;
+	const isText = Boolean(props.selectionText.length);
+	const isLink = Boolean(props.linkURL);
+
+	const getContextMenu = (): Electron.MenuItemConstructorOptions[] => {
```
andersk: There’s no reason for this to be a function.

app/renderer/js/components/context-menu.ts
andersk: Instead of confusing everything with repeated `insert` calls (which also waste quadratic time), just use `map` to build an array and use a single `concat` to prepend it to the template before calling `Menu.buildFromTemplate`.

manavmehta: Thanks @andersk :)
Please have a look now.

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -493,4 +499,59 @@ export default class GeneralSection extends BaseSection {
...
+			const availableLanguages = session.fromPartition('persist:webviewsession').availableSpellCheckerLanguages;
+			const languagePairs: Map<string, string> = new Map();
+			availableLanguages.forEach((l: string) => {
+				if (ISO6391.getName(l) !== '') {
+					languagePairs.set(ISO6391.getName(l), l);
+				}
+			});
+
+			// Manually set names for languages not available in ISO6391
+			languagePairs.set('English (AU)', 'en-AU'); languagePairs.set('English (CA)', 'en-CA'); languagePairs.set('English (GB)', 'en-GB'); languagePairs.set('English (US)', 'en-US'); languagePairs.set('Spanish (Latin America)', 'es-419'); languagePairs.set('Spanish (Argentina)', 'es-AR'); languagePairs.set('Spanish (Mexico)', 'es-MX'); languagePairs.set('Spanish (US)', 'es-US'); languagePairs.set('Portuguese (Brazil)', 'pt-BR'); languagePairs.set('Portuguese (Portugal)', 'pt-PT'); languagePairs.set('Serbo-Croatian', 'sh'); // eslint-disable-line max-statements-per-line
```
manavmehta: This had to be done manually for the same reason.

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -493,4 +499,59 @@ export default class GeneralSection extends BaseSection {
...
+			const configuredLanguages: string[] = [];
+			ConfigUtil.getConfigItem('spellcheck-languages').forEach((code: any) => {
+				[...languagePairs].forEach((pair): string | void => {
+					if (pair[1] === code) {
+						configuredLanguages.push(pair[0]);
+					}
+				});
+			});
+
+			tagify.addTags(configuredLanguages);
```
manavmehta, line 548: Again a consequence of missing language names in the module

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -493,4 +499,59 @@ export default class GeneralSection extends BaseSection {
...
+			<div class="setting-description">${t.__('SpellChecker Languages')}</div>
+			<input name='spellcheck' placeholder='Enter Languages'>`;
+
+			const availableLanguages = session.fromPartition('persist:webviewsession').availableSpellCheckerLanguages;
+			const languagePairs: Map<string, string> = new Map();
+			availableLanguages.forEach((l: string) => {
+				if (ISO6391.getName(l) !== '') {
+					languagePairs.set(ISO6391.getName(l), l);
+				}
+			});
```
manavmehta: This map could've been avoided.

manavmehta (commented):
The code would've been much shorter if the package [ISO-639-1](https://www.npmjs.com/package/iso-639-1) had corresponding names for some of the locale codes from electron.

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -493,4 +499,59 @@ export default class GeneralSection extends BaseSection {
...
+		// The elctron API is a no-op on macOS and macOS default spellchecker is used.
+		if (process.platform === 'darwin') {
+			const note = document.querySelector('#note');
+			note.innerHTML += 'On macOS, the OS spellchecker is used.<br>Change the language from System Preferences -> Keyboard -> Text -> Spelling.';
+		} else {
+			const note = document.querySelector('#note');
+			note.innerHTML += 'You can select a maximum of 3 languages for spellchecking.';
+			const spellDiv = document.querySelector('#spellcheck-langs');
+			spellDiv.innerHTML += `
+			<div class="setting-description">${t.__('SpellChecker Languages')}</div>
```
abhigyank:
```suggestion
				<div class="setting-description">${t.__('SpellChecker Languages')}</div>
```

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -493,4 +499,59 @@ export default class GeneralSection extends BaseSection {
...
+			languagePairs.set('English (AU)', 'en-AU'); languagePairs.set('English (CA)', 'en-CA'); languagePairs.set('English (GB)', 'en-GB'); languagePairs.set('English (US)', 'en-US'); languagePairs.set('Spanish (Latin America)', 'es-419'); languagePairs.set('Spanish (Argentina)', 'es-AR'); languagePairs.set('Spanish (Mexico)', 'es-MX'); languagePairs.set('Spanish (US)', 'es-US'); languagePairs.set('Portuguese (Brazil)', 'pt-BR'); languagePairs.set('Portuguese (Portugal)', 'pt-PT'); languagePairs.set('Serbo-Croatian', 'sh'); // eslint-disable-line max-statements-per-line
+			const tagField = document.querySelector('input[name=spellcheck]');
+			const tagify = new Tagify(tagField, {
+				whitelist: [...languagePairs.keys()],
+				enforceWhitelist: true,
+				maxTags: 3,
+				dropdown: {
+					enabled: 0,
+					maxItems: 100, // If chromium supports more and more languages in future
+					closeOnSelect: false,
```
abhigyank, line 553: I think we should have this as true.

manavmehta: Seems a little inconvenient to click on the input field to select more languages, so I preferred `false`. If you think it should be `true`, we'll go with that.
cc @akashnimare

abhigyank:
Functionality is working for me. Almost mergable I suppose.

app/renderer/js/pages/preference/general-section.ts
andersk: Why’s this crammed onto one line?

app/renderer/css/tagify.css
```diff
@@ -0,0 +1,2 @@
...
...9;transform:translateY(1px);overflow:hidden}.tagify__dropdown[placement=top]{margin-top:0;transform:translateY(-2px)}.tagify__dropdown[placement=top] .tagify__dropdown__wrapper{border-top-width:1px;border-bottom-width:0}.tagify__dropdown--text{box-shadow:0 0 0 3px rgba(var(--tagify-dd-color-primary),.1);font-size:.9em}.tagify__dropdown--text .tagify__dropdown__wrapper{border-width:1px}.tagify__dropdown__wrapper{max-height:300px;overflow:hidden;background:#fff;background:var(--tagify-dd-bg-color);border:1px solid #3595f6;border-color:var(--tagify-dd-color-primary);border-top-width:0;box-shadow:0 2px 4px -2px rgba(0,0,0,.2);transition:.25s cubic-bezier(0,1,.5,1)}.tagify__dropdown__wrapper:hover{overflow:auto}.tagify__dropdown--initial .tagify__dropdown__wrapper{max-height:20px;transform:translateY(-1em)}.tagify__dropdown--initial[placement=top] .tagify__dropdown__wrapper{transform:translateY(2em)}.tagify__dropdown__item{box-sizing:inherit;padding:.3em .5em;margin:1px;cursor:pointer;border-radius:2px;position:relative;outline:0}.tagify__dropdown__item--active{background:#3595f6;background:var(--tagify-dd-color-primary);color:#fff}.tagify__dropdown__item:active{filter:brightness(105%)}
```
andersk: I assume this is copied and pasted from the `@yaireo/tagify` package. We should never copy and paste dependencies; we should load them from `node_modules`. One way to do that would be to add the stylesheet in JavaScript from `require.resolve("@yaireo/tagify/dist/tagify.css")`. A more general solution would be to build the app with webpack and html-webpack-plugin, or some other bundler.

app/renderer/js/components/context-menu.ts
```diff
@@ -0,0 +1,105 @@
+import {remote} from 'electron';
+import WebView from './webview';
+import * as t from '../utils/translation-util';
+const {clipboard, Menu} = remote;
+
+export const contextMenu = (webview: WebView, event: any) => {
```
andersk: Avoid `any` if at all possible. `event` has type `Event`. There is no `params` member of `Event`, but the `context-menu` event handler is passed `params: ContextMenuParams` as a second parameter. (Reference: https://github.com/electron-userland/electron-builder/blob/master/typings/electron.d.ts.)

This also makes the below `selectionText as string` casts unnecessary.

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -493,4 +499,59 @@ export default class GeneralSection extends BaseSection {
 			await this.factoryResetSettings();
 		});
 	}
+
+	initSpellChecker(): void {
+		// The elctron API is a no-op on macOS and macOS default spellchecker is used.
+		if (process.platform === 'darwin') {
+			const note = document.querySelector('#note');
+			note.innerHTML += 'On macOS, the OS spellchecker is used.<br>Change the language from System Preferences -> Keyboard -> Text -> Spelling.';
```
andersk: This is invalid HTML (due to the `->`s). We should always try to avoid HTML generation in favor of DOM manipulation:

```ts
note.append('On macOS, the OS spellchecker is used.');
note.append(document.createElement('br'));
note.append('Change the language from System Preferences -> Keyboard -> Text -> Spelling.');
```

Additionally, the UI strings should be internationalized.

(And although it doesn’t change the fundamental point, consider true Unicode arrows: →.)

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -493,4 +499,59 @@ export default class GeneralSection extends BaseSection {
...
+			});
+
+			const configuredLanguages: string[] = [];
+			ConfigUtil.getConfigItem('spellcheck-languages').forEach((code: any) => {
+				[...languagePairs].forEach((pair): string | void => {
+					if (pair[1] === code) {
+						configuredLanguages.push(pair[0]);
+					}
+				});
+			});
```
andersk: This is a quadratic loop that could be avoided with better data structures.

`string | void` doesn’t make sense.

Generally loops around `push` should be avoided in favor of `map`.

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -493,4 +499,59 @@ export default class GeneralSection extends BaseSection {
...
+			// Manually set names for languages not available in ISO6391
+			languagePairs.set('English (AU)', 'en-AU'); languagePairs.set('English (CA)', 'en-CA'); languagePairs.set('English (GB)', 'en-GB'); languagePairs.set('English (US)', 'en-US'); languagePairs.set('Spanish (Latin America)', 'es-419'); languagePairs.set('Spanish (Argentina)', 'es-AR'); languagePairs.set('Spanish (Mexico)', 'es-MX'); languagePairs.set('Spanish (US)', 'es-US'); languagePairs.set('Portuguese (Brazil)', 'pt-BR'); languagePairs.set('Portuguese (Portugal)', 'pt-PT'); languagePairs.set('Serbo-Croatian', 'sh'); // eslint-disable-line max-statements-per-line
+			const tagField = document.querySelector('input[name=spellcheck]');
+			const tagify = new Tagify(tagField, {
+				whitelist: [...languagePairs.keys()],
+				enforceWhitelist: true,
+				maxTags: 3,
+				dropdown: {
+					enabled: 0,
+					maxItems: 100, // If chromium supports more and more languages in future
```
andersk: …then we just want to ignore all but 100 of them?

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -493,4 +499,59 @@ export default class GeneralSection extends BaseSection {
...
+				[...languagePairs].forEach((pair): string | void => {
+					if (pair[1] === code) {
+						configuredLanguages.push(pair[0]);
+					}
+				});
+			});
+
+			tagify.addTags(configuredLanguages);
+
+			tagField.addEventListener('change', (event: any) => {
```
andersk: Avoid `any`.

typings.d.ts
```diff
@@ -16,6 +16,11 @@ declare module '@electron-elements/send-feedback' {
 
 declare module 'node-mac-notifier';
 
+declare module '@yaireo/tagify' {
+	const tagify: any;
```
andersk: Avoid `any`.

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -493,4 +499,59 @@ export default class GeneralSection extends BaseSection {
...
+				if (ISO6391.getName(l) !== '') {
+					languagePairs.set(ISO6391.getName(l), l);
+				}
+			});
+
+			// Manually set names for languages not available in ISO6391
+			languagePairs.set('English (AU)', 'en-AU'); languagePairs.set('English (CA)', 'en-CA'); languagePairs.set('English (GB)', 'en-GB'); languagePairs.set('English (US)', 'en-US'); languagePairs.set('Spanish (Latin America)', 'es-419'); languagePairs.set('Spanish (Argentina)', 'es-AR'); languagePairs.set('Spanish (Mexico)', 'es-MX'); languagePairs.set('Spanish (US)', 'es-US'); languagePairs.set('Portuguese (Brazil)', 'pt-BR'); languagePairs.set('Portuguese (Portugal)', 'pt-PT'); languagePairs.set('Serbo-Croatian', 'sh'); // eslint-disable-line max-statements-per-line
+			const tagField = document.querySelector('input[name=spellcheck]');
+			const tagify = new Tagify(tagField, {
+				whitelist: [...languagePairs.keys()],
```
andersk, line 547: These ought to be sorted.

app/renderer/css/tagify.css
abhigyank: > One way to do that would be to add the stylesheet in JavaScript from `require.resolve("@yaireo/tagify/dist/tagify.css")`.

By this are you suggesting to add something like this in `preference.html` - 
```js
	let tagifyCSS = document.createElement("link");
	tagifyCSS.rel = "stylesheet";
	tagifyCSS.type = "text/css";
	tagifyCSS.href = require.resolve("@yaireo/tagify/dist/tagify.css");
	document.querySelector("head").appendChild(tagifyCSS);
```

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -493,4 +499,59 @@ export default class GeneralSection extends BaseSection {
...
+		// The elctron API is a no-op on macOS and macOS default spellchecker is used.
+		if (process.platform === 'darwin') {
+			const note = document.querySelector('#note');
+			note.innerHTML += 'On macOS, the OS spellchecker is used.<br>Change the language from System Preferences -> Keyboard -> Text -> Spelling.';
+		} else {
+			const note = document.querySelector('#note');
+			note.innerHTML += 'You can select a maximum of 3 languages for spellchecking.';
+			const spellDiv = document.querySelector('#spellcheck-langs');
+			spellDiv.innerHTML += `
+				<div class="setting-description">${t.__('SpellChecker Languages')}</div>
```
andersk: Should be “Spell checker languages”.

app/renderer/js/pages/preference/general-section.ts
manavmehta: There were 12 of them which would take up space and allow less code on screen thus reducing readability. I've spread it now.

app/renderer/js/pages/preference/general-section.ts
manavmehta: Presently 50 locales are supported, so made it 100.
Now changed to `Infinity`

app/renderer/css/tagify.css
manavmehta: @abhigyank the snippet can be shortened by assigning id to existing tagify \<link\>
```ts
document.querySelector('#tagifyCSS').href = require.resolve('@yaireo/tagify/dist/tagify.css');
```

typings.d.ts
manavmehta: Can we only declare the module? Thought it won't be able to use Intellisense, but we only need one method - `tagify.addTags()`.

app/renderer/preference.html
```diff
@@ -5,6 +5,7 @@
         <meta name="viewport" content="width=device-width">
         <title>Zulip - Settings</title>
         <link rel="stylesheet" href="css/preference.css" type="text/css" media="screen">
+        <link id = "tagify-css" rel="stylesheet" type="text/css" media="screen">
```
andersk:
```suggestion
        <link id="tagify-css" rel="stylesheet" href="data:text/css,">
```

(Spaces around `=` are unconventional in HTML, `href` is required, `type` is not, and restricting to `screen` might hide important clues from speech synthesizers.)

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -493,4 +499,66 @@ export default class GeneralSection extends BaseSection {
...
+				maxTags: 3,
+				dropdown: {
+					enabled: 0,
+					maxItems: Infinity,
+					closeOnSelect: false,
+					highlightFirst: true
+				}
+			});
+
+			const configuredLanguages: string[] = ConfigUtil.getConfigItem('spellcheck-languages').map((code: string) => [...languagePairs].filter(pair => (pair[1] === code))[0][0]);
```
andersk: That’s still quadratic.

```suggestion
			const configuredLanguages: string[] =
				ConfigUtil.getConfigItem('spellcheck-languages')
					.map((code: string) => ISO6391.getName(code));
```

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -493,4 +499,66 @@ export default class GeneralSection extends BaseSection {
...
+			note.append(t.__('You can select a maximum of 3 languages for spellchecking.'));
+			const spellDiv = document.querySelector('#spellcheck-langs');
+			spellDiv.innerHTML += `
+				<div class="setting-description">${t.__('Spellchecker Languages')}</div>
+				<input name='spellcheck' placeholder='Enter Languages'>`;
+
+			const availableLanguages = session.fromPartition('persist:webviewsession').availableSpellCheckerLanguages;
+			let languagePairs: Map<string, string> = new Map();
+			availableLanguages.forEach((l: string) => {
+				if (ISO6391.getName(l) !== '') {
```
andersk:
```suggestion
				if (ISO6391.validate(l)) {
```

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -493,4 +499,66 @@ export default class GeneralSection extends BaseSection {
...
+			languagePairs.set('English (US)', 'en-US');
+			languagePairs.set('Spanish (Latin America)', 'es-419');
+			languagePairs.set('Spanish (Argentina)', 'es-AR');
+			languagePairs.set('Spanish (Mexico)', 'es-MX');
+			languagePairs.set('Spanish (US)', 'es-US');
+			languagePairs.set('Portuguese (Brazil)', 'pt-BR');
+			languagePairs.set('Portuguese (Portugal)', 'pt-PT');
+			languagePairs.set('Serbo-Croatian', 'sh');
+
+			languagePairs = new Map([...languagePairs].sort(undefined));
```
andersk: Default-sorting an array of arrays is weird in JS; sort the keys instead with `[...languagePairs.keys()].sort()`.

app/renderer/js/pages/preference/general-section.ts
abhigyank: We won't be able to do this because ISO6391 doesn't have all the languages , so we have to map from languagePairs.
Even though, this looks quadratic, `ConfigUtil.getConfigItem('spellcheck-languages')` would at max have 3 values, so in reality the algorithm is actually linear on the size of languagePairs, which I think is okay.

manavmehta: @andersk 
```ts 
.map((code: string) => ISO6391.getName(code)); 
``` 
You're correct. This IS the convention. But lack of all language pairs makes it unusable here.
DO you have some other way in mind to make it linear ?

app/renderer/js/pages/preference/general-section.ts
manavmehta: Passed the comparator

app/renderer/js/components/context-menu.ts
manavmehta: @andersk Sorry but I'm having a little hard time getting to this. Can you please give some suggestion?

* The listener here is from `WebContents` class. Though we can definitely pass the webContents of the webview as that's what we're operating on, but the listener won't call our context-menu component.
* How is it obtaining the `params` object which is being passed as `ContextMenuParams`?
* Also keeping in mind the ContextMenuParams has everything we need but the `dictionarySuggestions[]`

Update: Solved

app/renderer/js/pages/preference/general-section.ts
```diff
@@ -93,7 +95,10 @@ export default class GeneralSection extends BaseSection {
 						<div class="setting-description">${t.__('Enable spellchecker (requires restart)')}</div>
 						<div class="setting-control"></div>
 					</div>
+					<div class="setting-row" id="spellcheck-langs"></div>
```
abhigyank: You can add this - 
` "display:${process.platform === 'darwin' ? '' : 'none'}" `, this will take care of one of your ToDo below.

akashnimare:
Merged. Thanks, everyone for working on this 🚀

akashnimare:
This works well. There are a couple of minor things which we'll be adding as a follow-up.

PR #971, opened 2020-05-31
mac: Support pkg installer on macOS.

akashnimare:
Add support for pkg on macOS.

To-dos -
- [x] Check installer
- [x] Test auto-updates

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,11 @@
     "copyright": "©2019 Kandra Labs, Inc.",
     "mac": {
       "category": "public.app-category.productivity",
+      "target": [
+        "dmg",
+        "pkg",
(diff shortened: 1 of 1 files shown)


andersk:
What happened to `package-lock.json`?

akashnimare:
Yeah, was wondering the same. Maybe I was on the other PR where I did an npm install and then switched to master and checked out to a new branch and did npm install again.

andersk:
Hmm, my `npm` does that too now. I guess we should commit the `package-lock.json` change—but separately, so it doesn’t get mixed in with other changes.

akashnimare:
Installer works fine. Just need to test auto-updates.

PR #974, opened 2020-06-09
translations: Prevent i18n to automatically add strings.

manavmehta:
**What's this PR do?**

Now that we have moved to Transifex, auto updation of translation strings is not required.
We would not want any of these changes to be getting committed.
We only need to update the resource file (en.json) and push it to Transifex and pulling the rest <lang>.json
The PR prevents auto updation of the strings not present
  - [X] macOS Catalina 10.15.4
  - [ ] Linux/Ubuntu
  - [ ] Windows


manavmehta:
The resource file will be committed right after testing.

PR #984, opened 2020-06-22
Electron v9 upgrade

manavmehta:
**What's this PR do?**
Upgrade to electron v9 (9.1.0)

  - [X] macOS Catalina 10.15.5
  - [ ] Windows
  - [ ] Linux/Ubuntu

**Structured Clone Algorithm Checklist (Check if working)**
- [ ] Sidebar
- [ ] DevTools for Active Tab
- [ ] Menubar menu items

### DIFF
--- a/app/renderer/js/utils/link-util.ts
+++ b/app/renderer/js/utils/link-util.ts
@@ -4,6 +4,13 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 
+import Logger from './logger-util';
+
+const logger = new Logger({
+	file: 'link-util.log',
+	timestamp: true
+});
+
 export function isUploadsUrl(server: string, url: URL): boolean {
 	return url.origin === server && url.pathname.startsWith('/user_uploads/');
 }
@@ -36,7 +43,11 @@ export async function openBrowser(url: URL): Promise<void> {
     </body>
 </html>
 `);
-		await shell.openPath(file);
+		const shellResponse = await shell.openPath(file);
+		if (shellResponse.length > 0) {
+			logger.log(shellResponse);
+		}
+
 		setTimeout(() => {
 			fs.unlinkSync(file);
 			fs.rmdirSync(dir);
--- a/app/renderer/js/utils/link-util.ts
+++ b/app/renderer/js/utils/link-util.ts
@@ -36,7 +36,7 @@ export async function openBrowser(url: URL): Promise<void> {
     </body>
 </html>
 `);
-		shell.openItem(file);
+		await shell.openPath(file);
 		setTimeout(() => {
 			fs.unlinkSync(file);
 			fs.rmdirSync(dir);
--- a/package-lock.json
+++ b/package-lock.json
@@ -3177,9 +3177,9 @@
       }
     },
     "electron": {
-      "version": "8.4.0",
-      "resolved": "https://registry.npmjs.org/electron/-/electron-8.4.0.tgz",
-      "integrity": "sha512-SpgyccM5rjDJSGcpQjiviUBT44fZlSyhcjy8RpKSnAad+co4xY1vYj6T25U1CfSk0PH/dhvcp63P2sdXHCwq/Q==",
+      "version": "9.1.0",
+      "resolved": "https://registry.npmjs.org/electron/-/electron-9.1.0.tgz",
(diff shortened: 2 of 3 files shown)


app/renderer/js/utils/link-util.ts
```diff
@@ -19,24 +19,24 @@ export async function openBrowser(url: URL): Promise<void> {
...
+					html {
+						font-family: menu, "Helvetica Neue", sans-serif;
+					}
+				</style>
+			</head>
+			<body>
+				<p>Opening <a href="${escape(url.href)}">${escape(url.href)}</a>…</p>
+			</body>
+		</html>
+		`);
```
andersk: The lack of indentation was intentional. Any extra whitespace would literally be written to the file. Whitespace before `<!DOCTYPE html>` is incompatible with at least some old web browsers. I’m not sure if it matters to any current ones. But this question is, at the very least, totally unrelated to the purpose of this commit, so don’t change the indentation here.

app/renderer/js/utils/link-util.ts
```diff
@@ -19,24 +19,24 @@ export async function openBrowser(url: URL): Promise<void> {
...
+		await shell.openExternal(file); // Check
```
andersk: The [documentation](https://www.electronjs.org/docs/breaking-changes) says that `shell.openItem` was replaced with `shell.openPath`, which takes a path, not with `shell.openExternal`, which takes a URL.

andersk:
I know this is WIP and doesn’t work yet, but the `getWebContents` change is fine to do in Electron 8, so I split out that part and merged it as e97ab2e6dd98ddb04cdabb61e477ff4292b1f44d.

app/renderer/js/utils/link-util.ts
```diff
@@ -36,7 +36,7 @@ export async function openBrowser(url: URL): Promise<void> {
     </body>
 </html>
 `);
-		shell.openItem(file);
+		await shell.openPath(file);
```
vsvipul: @manavmehta  Can we also utilize the error message in the API design https://github.com/electron/governance/blob/master/wg-api/spec-documents/shell-openitem.md
in case an error occurs ? Just to be safe in this case.

vsvipul:
Checked the [breaking changes](https://www.electronjs.org/releases/stable?page=2) 
The getWebContents change is already fixed in master. 
The shell changes are handled in this PR.
Rest of the breaking changes should not be relevant for our case. I still didn't understand the change ["IPC between main and renderer processes now uses the Structured Clone Algorithm"](https://github.com/electron/electron/pull/20214) and what we need to check in this case. If someone could look into that, that would be great!

app/renderer/js/utils/link-util.ts
manavmehta: @vsvipul I have added a logger instance for the same. Do you have something else in mind?

andersk:
The breaking part of the Structured Clone Algorithm change is

> **NOTE**: Objects that aren't serializable with V8's Structured Clone algorithm, such as functions, DOM objects, special Node/Electron objects like `process.env` or `WebContents`, or any objects containing such items will be serialized with the old `base::Value`-based algorithm. However, this behavior is deprecated and will throw an exception beginning with Electron 9.

To see an example of what happens when we try to transmit a non-serializable object, here’s an error we get in the app console at startup:

```
Uncaught (in promise) Error: An object could not be cloned.
    at EventEmitter.i.send.i.send (electron/js2c/renderer_init.js:74)
    at ServerManagerView.activateTab (/home/anders/zulip/zulip-desktop/app/renderer/js/main.js:539)
    at ServerManagerView.initTabs (/home/anders/zulip/zulip-desktop/app/renderer/js/main.js:265)
    at async ServerManagerView.init (/home/anders/zulip/zulip-desktop/app/renderer/js/main.js:92)
    at async /home/anders/zulip/zulip-desktop/app/renderer/js/main.js:890
```

This leaves much of the app non-functional (e.g. the buttons for add organization, do not disturb, reload, back, settings).

andersk:
Note that the Spectron version is [tied](https://github.com/electron-userland/spectron#version-map) to the Electron version, so they need to be upgraded simultaneously. I’ve done a full dependency upgrade in 0f1245b9758e72344f6a963061e0530c4b3a6fde, including Electron 9 and Spectron 11.

I’m not sure that writing a log message to a separate log file nobody will read is a helpful way to deal with a failure to launch a web browser, so I’ll leave that for future discussion.

manavmehta:
Sure, I'll keep that in mind for further upgrades. Thanks.
Closing this in favor of 0f1245b

Issue #915, opened 2020-04-01
Remove uses of the electron.remote module


andersk:
The Electron security documentation recommends [disabling the `remote` module](https://www.electronjs.org/docs/tutorial/security#15-disable-the-remote-module) for improved sandboxing, and it also has a number of [performance and reliability problems](https://medium.com/@nornagon/electrons-remote-module-considered-harmful-70d69500f31). We should stop using it. This may require some rethinking of how our logic is split between the main process and the renderer processes.

abhigyank:
Electron has released a separate [remote](https://github.com/electron/remote) module, which is a replacement for the built-in remote module in Electron. It does still recommend to use `ipc` methods instead, in its readme. So what do you propose would be good for us to do here?

manavmehta:
@abhigyank from what I can comprehend, it is still available to use but not recommended. IPC primitives (`ipcRenderer` and `ipcMain`) are.
Earlier, it was supposed to be moved to userland.

timabbott:
I would guess they're moving it to a separate module so that they can delete it from `electron` core because a lot of use of it can result in security problems.  Since use of this module doesn't seem important to the design of Zulip, we should just figure out how to remove use of it from the codebase.

PR #993, opened 2020-06-27
request: Replace deprecated request module with net.request.

### LINKED ISSUE #886: Remove deprecated request module
The `request` module is [deprecated](https://www.npmjs.com/package/request). Although it is still used by our indirect dependencies, we should replace our direct uses of it.

manavmehta:
Fixes: #886 

  - [X] macOS Catalina 10.15.5
  - [ ] Windows
  - [ ] Linux/Ubuntu

**Checklist**

- [x] domain-util
- [x] reconnect-util
- [x] linuxupdater
- [x] check add-new-server
- [x] check UA on test server

We may think of a more appropriate and suitable name of the new added request.ts


app/renderer/js/utils/domain-util.ts
```diff
@@ -1,17 +1,14 @@
...
 import Logger from './logger-util';
-import {remote} from 'electron';
+import {remote, ClientRequestConstructorOptions} from 'electron';
 
-import * as RequestUtil from './request-util';
 import * as EnterpriseUtil from './enterprise-util';
 import * as Messages from '../../../resources/messages';
 
-const {app, dialog} = remote;
+const {app, dialog, net, session} = remote;
```
andersk: Unfortunately I don’t think `electron.net` can be used safely via `remote`; see electron/electron#12545. I think we’ll need to build our own IPC interface and use it in the main process. (Which is just as well, because `remote` is also [discouraged](https://github.com/electron/electron/blob/master/docs/tutorial/security.md#15-disable-the-remote-module) and we should aim to eliminate it eventually.)

abhigyank: So it would be better to move all `electron.net` requests to the main process via a IPCRenderer/Main interface call?

Following this up, we could have a separate module in the main process that handles the replacement of remote calls in the renderer process? Solving #915 .

manavmehta: Okay, I was facing the `end` problem too. Seems plausible. Thanks @andersk .
Also, 
> IPC primitives (`ipcRenderer` and `ipcMain`) will be unaffected. This only affects the `remote` module, which is a convenience layer wrapped on top of those IPC primitives.
> 
> The recommended migration path for apps is to replace usage of the remote module with explicit use of the IPC primitives, for example [`ipcRenderer.invoke`](https://electronjs.org/docs/api/ipc-renderer#ipcrendererinvokechannel-args) / [`ipcMain.handle`](https://electronjs.org/docs/api/ipc-main#ipcmainhandlechannel-listener).
>

Is it safe to say that the the `primitives` are safer and better, performance-wise, to use and hence to migrate to?
Are there any cases where they lag behind `remote`?

manavmehta:
Also, for the record, [`remote` is now available as a package](https://github.com/electron/electron/issues/21408#issuecomment-648398795) (on it's way to be moved out from electron core)

andersk:
Let’s move discussion about `remote` to #915.

(But to answer your questions in short: yes, we should eliminate `remote`. Yes, we should eliminate it completely. No, there are no cases of it that should not be eliminated. It’s being moved from core to a package because it should *not* be used, not because it should be used. In case you haven’t read this page, which I linked from #915 and is also linked from the Electron bug and the `@electron/remote` documentation, see [Electron’s ‘remote’ module considered harmful](https://medium.com/@nornagon/electrons-remote-module-considered-harmful-70d69500f31). I think we’re all on the same page here. Good? Good.)

abhigyank:
The original functionality remains intact for me after the migration. Further, now we don't need certificate-util, I tested adding certificates to Chrome store and using `certutil`on Linux, and server gets added without a hitch.
All in all LGTM. 
One minor issue, in case of certificate error on server addition, we show two dialog box errors - one from electron.net and one we create. But I think that's a bug in how we handle certificate errors, could be a follow-up PR.

manavmehta:
Going with [this approach](https://chat.zulip.org/#narrow/stream/16-desktop/topic/remove.20ignoreCerts), I think it will handle two dialog boxes.

app/main/request.ts
```diff
@@ -0,0 +1,148 @@
...
+						typeof realm_name === 'string' &&
+						typeof realm_uri === 'string' &&
+						typeof realm_icon === 'string'
+					) {
+						resolve({
+							// Some Zulip Servers use absolute URL for server icon whereas others use relative URL
+							// Following check handles both the cases
+							icon: realm_icon.startsWith('/') ? realm_uri + realm_icon : realm_icon,
+							url: realm_uri,
+							alias: escape(realm_name)
```
abhigyank: Need to import `import escape from 'escape-html';` .
Currently the PR uses built-in `escape`, which's behavior is different from what we have in `domain-util.ts` .

abhigyank:
Let's keep getProxy for now? I feel it completes the the utility of a file like `proxy-util` though unused. Could be wrong here.
Removing request-util is correct.

andersk:
Disagree. There’s no point in keeping around code that won’t be used with `electron.net`. If it burns, set it on fire.

abhigyank:
I agree it won't be used with electron.net, however I disagree that keeping it could be worse. I disagree also that we should fire to anything that burns, houses, cars everything burns, I don't suppose we should set fire to them. Just because anything burns, we shouldn't burn them.

andersk:
Uh…unused code is not a house or a car. Unused code is a readability complication, a maintenance burden, a potential attack surface, and an all around waste of time and space. There are rare cases where unused code is useful for API symmetry or to help the type checker or something, but this isn’t one of them. This code should just go away.

abhigyank:
> There are rare cases where unused code is useful for API symmetry.

Yeah I was going for this, so that a util could have a set and get function. Since, we don't have it as an API explicitly, we could let it go then. One can always re-add it later if required. We can keep the commit that removes it :)

andersk:
We don’t have a set function and a get function here. A get function should return exactly the data that you put into a set function—symmetry. This `getProxy` function, on the other hand, returns the result of some weird string-mangling logic and maybe also sets an environment variable as a side effect. Its only purpose was to support the specific library that we’re now removing.

abhigyank:
@andersk Can you review this for merging?

app/main/linuxupdater.ts
```diff
@@ -12,41 +10,48 @@ const logger = new Logger({
 	timestamp: true
 });
 
-export function linuxUpdateNotification(): void {
+export function linuxUpdateNotification(Session: Electron.session): void {
```
andersk: Don’t capitalize the variable `Session`.

app/main/request.ts
```diff
@@ -0,0 +1,149 @@
...
+	return `${dir}/${hash >>> 0}${extension}`;
+};
+
+export const _getServerSettings = async (domain: string, Session: Electron.session): Promise<ServerConf> => {
+	const serverSettingsOptions: Electron.ClientRequestConstructorOptions = {
+		url: domain + '/api/v1/server_settings',
+		session: Session
+	};
+
+	return new Promise((resolve, reject) => {
```
andersk: Let’s try to extract the raw `Promise` logic into minimal shared helper functions and lean on `async`/`await` as much as possible. With a small fix to the Electron typings in `typings.d.ts` (justified by the [documentation](https://www.electronjs.org/docs/api/incoming-message#class-incomingmessage): “`IncomingMessage` implements the [Readable Stream](https://nodejs.org/api/stream.html#stream_readable_streams) interface”):

```typescript
declare module 'electron' {
    interface IncomingMessage extends NodeJS.ReadableStream {}
}
```

we can do this:

```typescript
import getStream from 'get-stream';

async function fetchResponse(request: ClientRequest): Promise<IncomingMessage> {
    return new Promise((resolve, reject) => {
        request.on("response", resolve);
        request.on("abort", reject);
        request.on("error", reject);
        request.end();
    });
}

export async function _getServerSettings(…): Promise<ServerConf> {
    const response = await fetchResponse(net.request({
        url: domain + '/api/v1/server_settings',
        session,
    }));
    if (response.statusCode !== 200) {
        throw response;  // maybe we want this to be an actual Error?
    }
    const {realm_name, realm_uri, realm_icon} = JSON.parse(await getStream.buffer(response));
    …
}
```

app/main/request.ts
```diff
@@ -0,0 +1,149 @@
...
+				}
+			});
+		});
+
+		request.on('error', error => reject(error));
+		request.end();
+	});
+};
+
+export const _saveServerIcon = async (Url: string, Session: Electron.session): Promise<string> => {
```
andersk: Don’t capitalize the variables `Url` and `Session`.

app/main/request.ts
```diff
@@ -0,0 +1,149 @@
...
+
+	// Create 'server-icons' directory if not existed
+	if (!fs.existsSync(dir)) {
+		fs.mkdirSync(dir);
+	}
+
+	return `${dir}/${hash >>> 0}${extension}`;
+};
+
+export const _getServerSettings = async (domain: string, Session: Electron.session): Promise<ServerConf> => {
```
andersk: Don’t capitalize the variable `Session`.

app/main/request.ts
```diff
@@ -0,0 +1,149 @@
...
+			logger.log(error);
+			logger.reportSentry(error);
+			resolve(defaultIconUrl);
+		}
+	});
+};
+
+/* Request: reconnect-util */
+
+export const _isOnline = async (reconnectUtil: ReconnectUtil, Session: Electron.session): Promise<boolean> => {
```
andersk: Don’t capitalize the variable `Session`.

app/renderer/js/utils/reconnect-util.ts
```diff
@@ -32,23 +30,8 @@ export default class ReconnectUtil {
...
-							!error && response.statusCode >= 200 && response.statusCode < 400;
-						resolve(isValidResponse);
-					}
-				);
-			} catch (error) {
-				logger.log(error);
-			}
-		});
+		const isOnlinePromise: Promise<boolean> = ipcRenderer.invoke('is-online', this);
+		return isOnlinePromise;
```
andersk: `return ipcRenderer.invoke('is-online', this)` seems fine.

app/main/request.ts
```diff
@@ -0,0 +1,149 @@
...
+					logger.log(error);
+					logger.reportSentry(error);
+					resolve(defaultIconUrl);
+				});
+
+				(response as any).pipe(file); // Avoid assertion
+
+				response.on('end', () => {
+					resolve(filePath);
+				});
```
andersk: I think the `async` equivalent here would be
```typescript
const pipeline = util.promisify(stream.pipeline);
…
await pipeline(response, file);
return filePath;
```

andersk:
This also seems to trigger an extremely unfortunate bug where one of the renderer processes goes into a 100% CPU loop that eventually consumes all available memory. I’m trying to debug this…

andersk:
The bug had something to do with trying to serialize an entire `ReconnectUtil` over the IPC channel. We only need its `url`, so yeah, let’s not do that.

manavmehta:
Thanks @andersk. Though,

> one of the renderer processes goes into a 100% CPU loop that eventually consumes all available memory.

I didn't get this totally. Can you please explain it to me?

akashnimare:
Thanks, everyone 🤝

PR #994, opened 2020-07-01
add support for basic authentication

schrej:
**What's this PR do?**

This PR adds support for HTTP Basic Authentication.
Electron fires a `login` event whenever Basic Authentication is triggered within a webview. This event now gets handled by showing a authentication dialog.

We have a special 2 Factor Authentication proxy in place, that uses Basic Authentication. If we want to use Zulip Desktop, we need it to support Basic Authentication as well.

**Screenshots?**

image

  - [ ] Windows
  - [ ] Linux/Ubuntu
  - [x] macOS

I've noticed the CSS of the app is pretty messy. It would be a good refactoring task to streamline it, especially regarding form input styling.

### DIFF
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -22,6 +22,7 @@ import * as EnterpriseUtil from './utils/enterprise-util';
 import * as LinkUtil from './utils/link-util';
 import * as Messages from '../../resources/messages';
 import type {DNDSettings} from './utils/dnd-util';
+import Dialog from './components/dialog';
 
 interface FunctionalTabProps {
 	name: string;
@@ -130,6 +131,7 @@ class ServerManagerView {
 		await this.loadProxy();
 		this.initDefaultSettings();
 		this.initSidebar();
+		this.initLoginEventHandler();
 		this.removeUAfromDisk();
 		if (EnterpriseUtil.hasConfigFile()) {
 			await this.initPresetOrgs();
@@ -1038,6 +1040,24 @@ class ServerManagerView {
 
 		ipcRenderer.on('open-network-settings', async () => {
 			await this.openSettings('Network');
+		});
+	}
+
+	initLoginEventHandler() {
+		// This event gets fired whenever a webview triggers a basic authentication.
+		// We show an authentication dialog and return the username and password.
+		app.on('login', async (event, webContents, details, authInfo, callback) => {
+			// This doesn't seem to work, so it's done in app/main/index.ts as well.
+			event.preventDefault();
+			const dialog = new Dialog();
+			dialog.init();
+			try {
+				const {username, password} = await dialog.showAuthenticationDialog();
+				callback(username, password);
+			} catch {
+				// Cancel authentication
+				callback();
+			}
 		});
 	}
 }
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -814,14 +814,14 @@ class ServerManagerView {
 			});
 		}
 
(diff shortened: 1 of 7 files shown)


akashnimare:
> We have a special 2 Factor Authentication proxy in place, that uses Basic Authentication. If we want to use Zulip Desktop, we need it to support Basic Authentication as well.

Is there any way I can see/check how it looks? Need to reproduce this auth case in order to better understand this PR. @andersk check this out please.

app/renderer/js/main.ts
```diff
@@ -1040,6 +1042,24 @@ class ServerManagerView {
 			await this.openSettings('Network');
 		});
 	}
+
+	initLoginEventHandler() {
+		// This event gets fired whenever a webview triggers a basic authentication.
+		// We show an authentication dialog and return the username and password.
+		app.on('login', async (event, webContents, details, authInfo, callback) => {
+			// This doesn't seem to work, so it's done in app/main/index.ts as well.
+			event.preventDefault();
```
andersk: The reason it doesn’t work is that you’re accessing `app` via `remote`, which doesn’t preserve synchronicity. We want to get rid of `remote` (#915). So assuming we need to do this in the renderer process, we should instead have the main process `login` handler contact the renderer process using IPC, probably using the same `makeRendererCallback` mechanism that the permission request handler uses.

andersk (commented):
Does this stop one Zulip organization from popping up a login dialog over a different Zulip organization? I’m worried this could be abused for phishing.

Also, this appears to just hide the used dialogs without ever removing them, which will leak memory and slow the app down over time.

schrej:
> Is there any way I can see/check how it looks? Need to reproduce this auth case in order to better understand this PR.

It's just [http basic authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#Basic_authentication_scheme). The easiest is probably a reverse proxy that requests http basic auth in front of a zulip instance.

Our 2FA setup simply expects a password and a TOTP concatenated in the password field.

> Does this stop one Zulip organization from popping up a login dialog over a different Zulip organization? I’m worried this could be abused for phishing.

Good point! It might be hard to limit it to a single tab. Alternatively we could show the host that's requesting authentication in the dialog.

> Also, this appears to just hide the used dialogs without ever removing them, which will leak memory and slow the app down over time.

It does only hide them, but future Dialogs use the same div to inject the dialog. It gets overwritten every time and doesn't leak memory. The idea was to allow showing other Dialogs as well in the future if necessary.

app/renderer/js/main.ts
schrej: I've found PRs regarding replacing webview with BrowserView. That might make it easier/possible to attach the event listener directly to that as well. I tried it with the webview but didn't get it to work.

I have no experience with electron, so I probably missed something.

schrej:
Ok, I've implemented the suggestions now:

I'm using IPC calls instead of remote. As it's asynchronous I'm storing the callbacks in index.ts until a response occurs. They don't get cleaned up in case no response arrives (which shouldn't happen), but as there can only be one callback per WebContents that should be ok.

The dialog now shows the origin of the authentication request.
image

There is one dialog per tab now. It's only visible when the tab is visible. The `Dialog` objects only hold a reference to `<body>` as long as they're not in use. They add a `<div>` to `<body>` to show a dialog and clean it up when it's dismissed.

schrej:
Missed the remark regarding `makeRendererCallback`, now using that instead. I've also implemented a (cryptic looking) helper function that extracts the callback id and passes a callback function as well as the remaining arguments to the event handler.

PR #1022, opened 2020-08-18
darwin-notifications: Migrate darwin-notifications to electron's native Notifications.

manavmehta:
**What's this PR do?**
Electron's native Notifications now supports inline replies on macOS and thus we don't need any 3rd party package.
This commit removes the usage of node-mac-notifier and brings native Notification in picture.

For comprehensibility, ipc-helpers.ts will contain the functions which make use of Main process APIs with already functioning code in Renderer process.

This is a potential solution to people not getting notifications on macOS.

  - [X] macOS Catalina 10.15.6

**Follow-ups**
  - [x] Display sender's profile picture on each notification

### DIFF
--- a/app/main/ipc-helpers.ts
+++ b/app/main/ipc-helpers.ts
@@ -1,0 +1,16 @@
+/**
+ * This file contains the functions which make use of Main process APIs with already functioning code in Renderer process.
+ */
+import {Notification} from 'electron';
+
+export const showDarwinNotification = (
+	event: Electron.IpcMainEvent,
+	notificationOptions: Electron.NotificationConstructorOptions
+) => {
+	const notification = new Notification(notificationOptions);
+	notification.show();
+
+	notification.on('reply', (_event, response) => {
+		event.sender.send('replied', response);
+	});
+};
--- a/app/main/ipc-helpers.ts
+++ b/app/main/ipc-helpers.ts
@@ -1,16 +1,38 @@
 /**
  * This file contains the functions which make use of Main process APIs with already functioning code in Renderer process.
  */
-import {Notification} from 'electron';
+import {Notification, nativeImage} from 'electron';
 
-export const showDarwinNotification = (
+import fetch from 'node-fetch';
+
+export const showDarwinNotification = async (
 	event: Electron.IpcMainEvent,
-	notificationOptions: Electron.NotificationConstructorOptions
+	notificationOptions: Electron.NotificationConstructorOptions,
+	profilePicURL: string
 ) => {
+	const profilePic = await getNativeImagefromUrl(profilePicURL);
+	notificationOptions.icon = profilePic;
+
 	const notification = new Notification(notificationOptions);
 	notification.show();
-
 	notification.on('reply', (_event, response) => {
 		event.sender.send('replied', response);
 	});
+};
+
+/* TODO: Migrate url->dataUri conversion part to renderer process
+	to make use of browser caching of images.
+*/
+const getNativeImagefromUrl = async (imageURL: string) => {
+	try {
+		const response = await fetch(imageURL);
+		const buffer = await response.buffer();
+		const base64String = buffer.toString('base64');
+		const dataUri = `data:image/png;base64,${base64String}`;
+		const image = nativeImage.createFromDataURL(dataUri);
+		return image;
+	} catch (error) {
+		console.error(error);
+		return null;
+	}
 };
--- a/app/renderer/js/notification/darwin-notifications.ts
+++ b/app/renderer/js/notification/darwin-notifications.ts
@@ -1,12 +1,10 @@
 import {ipcRenderer} from 'electron';
 
-import MacNotifier from 'node-mac-notifier';
-
 import electron_bridge from '../electron-bridge';
 import * as ConfigUtil from '../utils/config-util';
 
 import {
-	appId, customReply, focusCurrentServer, parseReply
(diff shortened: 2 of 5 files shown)


app/main/ipc-helpers.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * This file contains the functions which make use of Main process APIs with already functioning code in Renderer process.
+ */
+import {Notification, nativeImage} from 'electron';
+
+import fetch from 'node-fetch';
```
andersk, line 6: `node-fetch` doesn’t use the Electron network stack, so this won’t work with custom certificates, proxies, etc. We should use `electron.net` like everything else.

app/main/ipc-helpers.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/* TODO: Migrate url->dataUri conversion part to renderer process
+	to make use of browser caching of images.
+*/
+const getNativeImagefromUrl = async (imageURL: string) => {
+	try {
+		const response = await fetch(imageURL);
+		const buffer = await response.buffer();
+		const base64String = buffer.toString('base64');
+		const dataUri = `data:image/png;base64,${base64String}`;
```
andersk, line 31: How do we know it’s a PNG?

app/renderer/js/notification/darwin-notifications.ts
```diff
@@ -22,30 +20,21 @@ class DarwinNotification {
...
+			body: options.body,
+			silent: ConfigUtil.getConfigItem('silent') || false,
+			hasReply: true,
+			timeoutType: 'default'
+		};
+
+		ipcRenderer.send('create-notification', notificationOptions, profilePicURL);
+		ipcRenderer.on('replied', async (_event: Electron.IpcRendererEvent, response: string) => {
+			await this.notificationHandler({response});
+			ipcRenderer.removeAllListeners('replied');
```
andersk, line 36: Why `removeAllListeners`? Did you just want to use `ipcRenderer.once` instead?

app/renderer/js/notification/darwin-notifications.ts
```diff
@@ -103,8 +92,8 @@ class DarwinNotification {
 		customReply(response);
 	}
 
-	// Method specific to notification api
-	// used by zulip
+	// Method specific to Zulip's notification module, not the package used to create the same.
+	// If at all the need arises to migrate to another API (or npm package), DO NOT REMOVE THIS.
```
andersk, line 96: I don’t understand this new comment. [`Notification.close()`](https://developer.mozilla.org/en-US/docs/Web/API/Notification/close) is part of the standard `Notification` API. It’s not Zulip-specific.

andersk (commented):
Chrome now also supports [inline notification replies](https://chromestatus.com/feature/5743740178137088). I think it would be better to use that API to implement this feature in the web app so we don’t have to maintain this platform-specific monkey patch in zulip-desktop.

app/main/ipc-helpers.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import fetch from 'node-fetch';
+
+export const showDarwinNotification = async (
+	event: Electron.IpcMainEvent,
+	notificationOptions: Electron.NotificationConstructorOptions,
+	profilePicURL: string
+) => {
+	const profilePic = await getNativeImagefromUrl(profilePicURL);
+	notificationOptions.icon = profilePic;
```
andersk, line 14: Why do we need to go through the `NativeImage` dance? Can we not just set `icon` to the URL?

app/main/ipc-helpers.ts
manavmehta: My bad. Referring to [this](https://github.com/RocketChat/Rocket.Chat.Electron/blob/062b4f958c942f2dc4eee9066ec8c28aa93db705/src/main/ui/notifications.js#L32-L44) piece of code, I missed this part somehow fiddling around. Adding it back.

app/main/ipc-helpers.ts
manavmehta: Trying to do that from the beginning but it throws this error while converting
this

app/renderer/js/notification/darwin-notifications.ts
manavmehta: The previous comment got me believing that this was specific to `node-mac-notifier`. So I tried to refine it. I'll remodel it.

app/main/ipc-helpers.ts
akashnimare: @andersk the API doesn't support passing the URL directly. That's why we have to go through the `nativeImage` process.
https://www.electronjs.org/docs/api/notification#new-notificationoptions

cc @priyank-p

akashnimare:
@manavmehta would you be up for finishing this?

mateuszmandera:
Any updates on this?

manavmehta:
@mateuszmandera Sorry I was really occupied the last two months. I'll try to get back by Tuesday :)

mateuszmandera:
@manavmehta Awesome, thanks!

manavmehta:
@mateuszmandera do you have macOS too?

mateuszmandera:
@manavmehta Sadly no :disappointed:

manavmehta:
Considering the [discussion](https://chat.zulip.org/#narrow/stream/16-desktop/topic/macOS.20notifications) going on in CZO stream, I will be closing this PR for now.

Issue #1037, opened 2020-10-27
Strip whitespace in the desktop app when inputting the server URL


timabbott:
https://chat.zulip.org/#narrow/stream/137-feedback/topic/.22Add.20organization.22.20weirdness/near/1048966 has context, but we should strip leading and trailing whitespace to avoid various common mistakes from slowing down users. 

This should be a great first issue for someone new to the project.

dhruvshah0803:
Hey i am new to open source development and want to work on this can you please assign me this.

abhi14nexu:
I would really like to work on this issue.

abhi14nexu:
@zulipbot claim

zulipbot:
Hello @abhi14nexu, it looks like you've currently claimed 1 issue in this repository. We encourage new contributors to focus their efforts on at most 1 issue at a time, so please complete your work on your other claimed issues before trying to claim this issue again.

We look forward to your valuable contributions!

abhi14nexu:
@zulipbot claim

zulipbot:
Hello @abhi14nexu, you claimed this issue to work on it, but this issue and any referenced pull requests haven't been updated for 10 days. Are you still working on this issue?

If so, please update this issue by leaving a comment on this issue to let me know that you're still working on it. Otherwise, I'll automatically remove you from this issue in 4 days.

If you've decided to work on something else, simply comment `@zulipbot abandon` so that someone else can claim it and continue from where you left off.

Thank you for your valuable contributions to Zulip!

aryanshridhar:
@zulipbot claim

aryanshridhar:
Taking this issue since <code>@zulipbot claim</code> doesn't seems to be working .

SidharajYadav:
Hey i am new to open source development and want to work on this can you please assign me this.

PR #1041, opened 2020-11-15
server_form: Added whitespace strip while entering url.

aryanshridhar:
Small fix for #1037

Would love an review @timabbott whenever you feel so :)


andersk:
Merged with an edited commit message.

PR #1043, opened 2020-11-18
server_form: Added autocomplete function to url.

### LINKED ISSUE #1012: add organization: url: attempt to autocomplete with zulipchat.com
**Problem Description**
image

**Proposed Solution**
If the entered string does:

aryanshridhar:
Added functionality for autocomplete when the user only passes a Organization URL string .
As discussed <a href = 'https://github.com/zulip/zulip-desktop/issues/1012#issuecomment-698234662'>here</a> : 

The following interpretation is made keeping the rest as it is : 
foo → ht<span>tps://</span>foo.zulipchat.com 

Fixes #1012

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS

### DIFF
--- a/app/renderer/js/pages/preference/new-server-form.ts
+++ b/app/renderer/js/pages/preference/new-server-form.ts
@@ -65,11 +65,22 @@ export default class NewServerForm extends BaseComponent {
 		this.$newServerUrl = this.$newServerForm.querySelectorAll('input.setting-input-value')[0] as HTMLInputElement;
 	}
 
+	autoComplete(url: string): string {
+		const pattern = /^[a-zA-Z\d-]*$/;
+		let serverUrl = url.trim();
+
+		if (pattern.test(serverUrl)) {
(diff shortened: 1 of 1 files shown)


tarun8718 (changes requested):
I think when the user enters the actual Url that is https://foo.zulipchat.com your code would still add the suffix and prefix.

>  https://foo.zulipchat.com →https://https://foo.zulipchat.com.zulipchat.com

aryanshridhar:
> I think when the user enters the actual Url that is https://foo.zulipchat.com your code would still add the suffix and prefix.
> 
> > https://foo.zulipchat.com →https://https://foo.zulipchat.com.zulipchat.com

That's interesting ! , It returns <code>false</code> when such a Url is passed , hence not adding the suffix and prefix.
<pre>
> const pattern = /^[a-zA-Z\d-]*$/;
undefined
> pattern.test("https://foo.zulipchat.com");
false
> 
</pre>

PR #1062, opened 2021-01-11
Shortcuts-section: Deduplicate templateMacHTML and templateWinLinHTML.

tarun8718:

---

**What's this PR do?**

- Deduplicates templateMacHTML and templateWinLinHTML functions.
- Solves inconsistencies between both templates.
- Adds missing '+' between every two keys.

  - [ ] Windows
  - [X] Linux/Ubuntu
  - [x] macOS

### DIFF
--- a/app/renderer/js/pages/preference/shortcuts-section.ts
+++ b/app/renderer/js/pages/preference/shortcuts-section.ts
@@ -16,12 +16,9 @@ export default class ShortcutsSection extends BaseSection {
 		this.props = props;
 	}
 
-	// TODO - Deduplicate templateMacHTML and templateWinLinHTML functions. In theory
-	// they both should be the same the only thing different should be the userOSKey
-	// variable but there seems to be inconsistences between both function, one has more
-	// lines though one may just be using more new lines and other thing is the use of +.
-	templateMacHTML(): string {
-		const userOSKey = '⌘';
+	// eslint-disable-next-line complexity
+	templateHTML(): string {
+		const cmdOrCtrl = process.platform === 'darwin' ? '⌘' : 'Ctrl';
 
 		return htmlEscape`
 						<div class="settings-pane">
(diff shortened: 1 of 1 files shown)


tarun8718:
Would love a review @abhigyank, @akashnimare.

app/renderer/js/pages/preference/shortcuts-section.ts
```diff
@@ -30,35 +26,35 @@ export default class ShortcutsSection extends BaseSection {
...
 										<td>${t.__('Keyboard Shortcuts')}</td>
 									</tr>
 									<tr>
-										<td><kbd>${userOSKey}</kbd> + <kbd>Shift</kbd> + <kbd>M</kbd></td>
+										<td><kbd>${OSKey}</kbd> + <kbd>Shift</kbd> + <kbd>M</kbd></td>
 										<td>${t.__('Toggle Do Not Disturb')}</td>
 									</tr>
-									<tr>
-										<td><kbd>Shift</kbd><kbd>${userOSKey}</kbd><kbd>D</kbd></td>
+									<tr style= "display:${process.platform === 'darwin' ? '' : 'none'}">
```
andersk: The empty string is not a valid value for the `display` property, and the `hidden` attribute is simpler:

```suggestion
									<tr ${process.platform === 'darwin' ? '' : 'hidden'}>
```

app/renderer/js/pages/preference/shortcuts-section.ts
```diff
@@ -30,35 +26,35 @@ export default class ShortcutsSection extends BaseSection {
...
+										<td><kbd>Shift</kbd> + <kbd>${OSKey}</kbd> + <kbd>D</kbd></td>
```
andersk: We should be consistent with `Shift`+`Ctrl` vs. `Ctrl`+`Shift`. The conventional order on Windows and Linux is definitely `Ctrl`+`Shift`. Not sure about macOS; it looks like its menus tend to show ⇧⌘ but that might just be for alignment?

app/renderer/js/pages/preference/shortcuts-section.ts
```diff
@@ -16,12 +16,8 @@ export default class ShortcutsSection extends BaseSection {
...
 	}
 
-	// TODO - Deduplicate templateMacHTML and templateWinLinHTML functions. In theory
-	// they both should be the same the only thing different should be the userOSKey
-	// variable but there seems to be inconsistences between both function, one has more
-	// lines though one may just be using more new lines and other thing is the use of +.
-	templateMacHTML(): string {
-		const userOSKey = '⌘';
+	templateHTML(): string {
+		const OSKey = (process.platform === 'darwin') ? '⌘' : 'Ctrl';
```
andersk: Upper-camel-case is typically reserved for class names in JavaScript. I’d call this `cmdOrCtrl`, which is consistent with the [Electron modifier name](https://www.electronjs.org/docs/api/accelerator#available-modifiers).

andersk:
Also this is just a TODO, not a TS TODO, though neither of these need to be in the commit message—I’d title this `shortcuts-section: Deduplicate templateMacHTML and templateWinLinHTML.`

tarun8718:
Hey @andersk, Thanks for the review I've made all the changes you mentioned.

app/renderer/js/pages/preference/shortcuts-section.ts
```diff
@@ -106,39 +114,63 @@ export default class ShortcutsSection extends BaseSection {
...
+										<td>${t.__('Toggle Sidebar')}</td>
+									</tr>
+									<tr ${process.platform === 'darwin' ? '' : 'hidden'}>
+										<td><kbd>Option</kbd> + <kbd>${cmdOrCtrl}</kbd> + <kbd>I</kbd></td>
 										<td>${t.__('Toggle DevTools for Zulip App')}</td>
 									</tr>
-									<tr>
-										<td><kbd>Option</kbd><kbd>${userOSKey}</kbd><kbd>U</kbd></td>
+									<tr ${process.platform === 'darwin' ? 'hidden' : ''}>
+										<td><kbd>Shift</kbd> + <kbd>${cmdOrCtrl}</kbd> + <kbd>I</kbd></td>
```
andersk: Should be `Ctrl`+`Shift`.

tarun8718: My bad, I checked it multiple times but I don't know how I missed it.

PR #1064, opened 2021-01-12
Reloads whenever a new App language is chosen.

tarun8718:

---

**What's this PR do?**

- Reloads the window once a new App language is chosen.
- Removes Text "(Restart required)" 
- Updates Translations

Language

  - [ ] Windows
  - [x] Linux/Ubuntu
  - [ ] macOS


tarun8718:
Would love a review @abhigyank, @akashnimare.

tarun8718:
@andersk would love a review for this PR.

andersk:
This results in an unexpected scroll to the top of the page that may confuse the user (especially if they just selected the wrong language!), and also fails to update the menu bar. I think we should leave the behavior as is.

PR #1095, opened 2021-03-29
request: Use a stable URL for the isOnline check

andersk:


timabbott:
I wonder if using `HEAD` is risky, in the context that the Zulip server might be behind some broken corporate proxy that doesn't pass HEAD requests through.  Is that possible?  I hope not.

Otherwise lgtm.

andersk:
I’ve never heard of such a broken proxy, and if there is one, the user would just need to manually click the refresh/reconnect button like they do now.

PR #1156, opened 2021-10-23
Enable context isolation on local content

andersk:
This is an important security hardening measure that’s been blocked on an upstream Electron bug I filed last year (electron/electron#26904); however, I think I’ve gotten it working with an ugly workaround.

This is still a draft because Spectron doesn’t support context isolation. So I’m also working on migrating the E2E tests from Spectron to Playwright in #1155. That PR seems to pass the E2E tests by itself, but not yet in combination with this one due to an upstream issue that needs further investigation.

### DIFF
--- a/app/common/config-util.ts
+++ b/app/common/config-util.ts
@@ -1,4 +1,3 @@
-import electron from "electron";
 import fs from "fs";
 import path from "path";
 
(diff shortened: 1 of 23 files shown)


andersk:
I think this introduces some kind of bug related to the `WebView.focus` “hacks”, which may explain the failure with #1155.

andersk:
Further investigation of the focus issue has resulted in a somewhat better understanding, a few workarounds, and the filing of electron/electron#31918.

andersk:
But here’s another problem:

* electron/electron#31924

PR #1201, opened 2022-04-13
preload: Drop compatibility injected JS for Zulip Server < 3.0

andersk:
The server was updated in zulip/zulip@a6fee2f18ef9d2ef6ac248e9ed82d580daff1a07 (`3.0-dev~1674`) and zulip/zulip@e701f208619b8b9b28a85f84ee16cf8d8df82b72 (`3.0-dev~1667`) to avoid relying on this wrapper. We no longer support servers older than 3.0, so we can delete it.


timabbott:
Yay, I'm excited that we get to drop this! I guess we need to test whether the desktop app give a good error message when connecting with a Zulip server that's 2.1 or older?

andersk:
> Yay, I'm excited that we get to drop this! I guess we need to test whether the desktop app give a good error message when connecting with a Zulip server that's 2.1 or older?

Adding a banner for this in #1304.

PR #1203, opened 2022-04-15
Fix blurry org icons in controls container

evinr:
**What's this PR do?**
Fixes blurry org icons 

Let me know if the width fix is preferred, but figure fixing without changing the size is a better approach. 

**Screenshots?**
Screenshot from 2022-04-15 13-17-24
Screenshot from 2022-04-15 13-14-14
Screenshot from 2022-04-15 12-09-15

- [ ] Windows
- [x] Linux/Ubuntu
- [ ] macOS

### DIFF
--- a/app/renderer/css/main.css
+++ b/app/renderer/css/main.css
@@ -163,6 +163,8 @@ body {
   width: 35px;
   vertical-align: top;
   border-radius: 4px;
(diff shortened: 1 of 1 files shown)


evinr:
Bumping this. Any ideas on how this fix can be accepted?

PR #1204, opened 2022-04-19
Fix autocomplete url

### LINKED ISSUE #1012: add organization: url: attempt to autocomplete with zulipchat.com

manila:
---

**What's this PR do?**

Add auto complete domain when adding a new server, similar UX to the mobile apps when adding a new server. 

Fixes #1012 

This is my first implementation and I anticipate that this could be farther improved. My main goal was to keep this logic in the scope of the new-server-form. I suspect this could be better implemented in a separate module

**Screenshots**
zulip-autocomplete-domain

- [ ] Windows
- [ ] Linux/Ubuntu
- [x] macOS

### DIFF
--- a/app/renderer/js/pages/preference/new-server-form.ts
+++ b/app/renderer/js/pages/preference/new-server-form.ts
@@ -111,6 +111,7 @@ export function initNewServerForm({$root, onChange}: NewServerFormProps): void {
     }
   });
   $newServerUrl.addEventListener("input", async () => {
+    $newServerUrl.value = $newServerUrl.value.trim();
     const url = $newServerUrl.value;
     $urlSizeCalc.textContent = url;
     $newServerUrl.style.width = `${$urlSizeCalc.offsetWidth}px`;
--- a/app/renderer/js/pages/preference/new-server-form.ts
+++ b/app/renderer/js/pages/preference/new-server-form.ts
@@ -60,7 +60,9 @@ export function initNewServerForm({$root, onChange}: NewServerFormProps): void {
     $saveServerButton.textContent = "Connecting...";
     let serverConf;
     try {
-      serverConf = await DomainUtil.checkDomain($newServerUrl.value.trim());
+      serverConf = await DomainUtil.checkDomain(
+        await autoComplete($newServerUrl.value.trim()),
+      );
     } catch (error: unknown) {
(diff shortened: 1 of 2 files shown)


app/renderer/js/pages/preference/new-server-form.ts
```diff
@@ -56,6 +56,7 @@ export function initNewServerForm({$root, onChange}: NewServerFormProps): void {
     "input.setting-input-value",
   )!;
 
+<<<<<<< HEAD
```
timabbott: Can you remove this merge conflict marker?

app/renderer/js/pages/preference/new-server-form.ts
```diff
@@ -17,11 +17,16 @@ export function initNewServerForm({$root, onChange}: NewServerFormProps): void {
...
-          class="setting-input-value"
-          autofocus
-          placeholder="your-organization.zulipchat.com or zulip.your-organization.com"
-        />
+        <label class="setting-input-value">
+          <input
+            class="setting-input-add-server"
+            autofocus
+            placeholder="your-organization.zulipchat.com"
+            style="width: 34ch"
```
timabbott: Can you do all the styles in CSS files? It's generally harder to read code split between places.

app/renderer/js/pages/preference/new-server-form.ts
timabbott: I think perhaps you remove this in the second commit; in general it's worth reviewing each individual commit with a took like `gitk --all` for readability.

app/renderer/js/pages/preference/new-server-form.ts
manila: That makes sense, I went ahead and made this change in my most recent commit 

My original thought was that by having this specific rule inline it would be contextual and less of a magic number. 34ch is the initial width of the input field which is the width of the placeholder text 

Maybe I should add a comment in the stylesheet? or perhaps it's obvious

app/renderer/js/pages/preference/new-server-form.ts
manila: Where has gitk been all my life!? Very useful tool, thanks for suggesting

This should be removed now

PR #1234, opened 2022-09-03
Adding basic rpm build

### LINKED ISSUE #1139: Provide a `.rpm` file for Fedora
Would it be possible to provide a `.rpm` file for Fedora and possibly publish to the relevant repository? Having a `.appimage` works, but it's not ideal for those of us that prefer to have things fully installed; while AppImageLauncher works, it feels like quite the hack. I have tried the snap installation but it freezes and crashes constantly. The only alternative is to use Zulip on the web, which isn't preferable.

batistein:
**What's this PR do?**
Adds a rpm build target to electron-builder

Fixes #1139 

- [ ] Windows
- [x] Linux/Ubuntu
- [x] macOS

### DIFF
--- a/packaging/rpm-after-install.sh
+++ b/packaging/rpm-after-install.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# Link to the binary
+ln -sf '/opt/${productFilename}/${executable}' '/usr/bin/${executable}'
+
+# SUID chrome-sandbox for Electron 5+
+chmod 4755 '/opt/${productFilename}/chrome-sandbox' || true
+
+update-mime-database /usr/share/mime || true
+update-desktop-database /usr/share/applications || true
+
+# Clean up configuration for old Bintray repository
+rm -f /etc/apt/zulip.list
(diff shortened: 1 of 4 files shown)


jhpratt:
Off chance someone could take a look at this PR? It's been open for over a year and is very small. I don't know the organizational structure of Zulip, so I don't know who to ping.

packaging/rpm-after-install.sh
```diff
@@ -0,0 +1,13 @@
...
+ln -sf '/opt/${productFilename}/${executable}' '/usr/bin/${executable}'
+
+# SUID chrome-sandbox for Electron 5+
+chmod 4755 '/opt/${productFilename}/chrome-sandbox' || true
+
+update-mime-database /usr/share/mime || true
+update-desktop-database /usr/share/applications || true
+
+# Clean up configuration for old Bintray repository
+rm -f /etc/apt/zulip.list
```
timabbott, line 13: This file will never exist in an RPM system.

timabbott:
It will be considered when we're next doing an Electron release. Though given the above bug, this seems to be somewhat sloppily implemented, so no promised that it'll prove easy to integrate.

PR #1251, opened 2022-12-01
Upgrade electron to final v18 release, especially for Wayland support on Linux.

klardotsh:
**What's this PR do?**

Manually branched from zulip/zulip-desktop#1245 to update Electron to the [final version of the v18 series](https://github.com/electron/electron/releases/tag/v18.3.15), which Dependabot did not automatically pick up.
    
Notably, this upgrade includes [native Wayland support on Linux](https://github.com/electron/electron/releases/tag/v18.3.8), which is particularly useful to avoid fuzzy nearest-neighbor scaling of XWayland apps in some Wayland compositors (see my screenshots below).

**Screenshots?**

Currently-published `Zulip-5.9.3-x86_64.AppImage` on swaywm/sway v1.7, 1080p monitor scaled 125%. Note the blurry nearest-neighbor upscaling that occurs due to XWayland apps not understanding HiDPI in this environment:

screenshot-2022-12-01-00:27:53Z

After applying this patch and running `npm ci && npm run dist`, the resulting AppImage on the same setup supports Wayland natively (as confirmed by not seeing `[XWayland]` in my title bar, and by the DPI awareness the Chromium renderer exhibits). While it's subtle in the screenshots (less so if you focus on the octopus in the topic header...), the fonts and images are noticeably crisper, and the difference on my physical display is noteworthy 😃 

screenshot-2022-12-01-00:27:12Z

I manually spliced `~/.local/share/applications/Zulip.desktop` to provide the appropriate Ozone CLI flags; I'm not sure where to provide those as a default to new installs (or actually, I might have hand-written this file anyway, I'm not sure Zulip provides one... it's been a few weeks so I don't really remember).

```
[Desktop Entry]
Type=Application
Name=Zulip
Exec="/home/j/src/zulip/desktop/dist/Zulip-5.9.3-x86_64.AppImage" --ozone-platform-hint=auto %u
Icon=Zulip
Comment=
Terminal=false
Categories=Communication;
MimeType=x-scheme-handler/zulip;
```

- [ ] Windows
- [X] Linux (Void Linux glibc)
- [ ] macOS

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,7 @@
         "@types/requestidlecallback": "^0.3.4",
         "@types/yaireo__tagify": "^4.3.2",
         "dotenv": "^16.0.0",
-        "electron": "^18.3.7",
+        "electron": "^18.3.15",
         "electron-builder": "^23.0.3",
         "electron-notarize": "^1.0.0",
         "eslint-import-resolver-typescript": "^2.4.0",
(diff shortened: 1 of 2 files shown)


klardotsh:
Hmmmm there *is* a new line between each message I can't repro on the public 5.9.3 release which might imply a regression here, would be curious if any Xorg users, other Wayland folk, or non-Linux users could (or could not) repro. A zoomed in view for emphasis:
screenshot-2022-12-01-00:41:38Z

timabbott:
I feel like I saw someone reporting those weird lines in some previous version of Chrome a while back that went away; so probably upgrading to an even newer Electron will help.

klardotsh:
Did some local upgrade testing today while chasing an unrelated issue to the core reason I opened this PR.

- This version of Electron v18 is the end of that series, so no further upgrades available there.
- Upgrading to Electron 19.1.9 (which, despite being released this year, is already EOL) produces a working binary, but does not eliminate the extra lines issue.
- Upgrading to v20 or v21 both fail to produce working applications (the shell window shows up, but no realms in the sidebar, and nothing in the main frame)

I'll be diving into the CSS (or if I find my sane limits there, tagging in a frontend person as their time allows) next, since this seems to be a noteworthy UI regression, but upgrading Electron further isn't bearing fruit.

PR #1264, opened 2023-01-10
Update installation instructions for macOS

fwcd:
**What's this PR do?**

Update the installation instructions for macOS, in particular the `brew cask install ...` command, which has been replaced with `brew install --cask ...` in more recent versions of Homebrew.


andersk:
Merged, thanks @fwcd!

PR #1265, opened 2023-01-10
Add new Big Sur-style rectangle icon for macOS

### LINKED ISSUE #1253: Use square icon in MacOS dock
To better align with MacOS design patterns, we should use a square Zulip icon in the Mac dock, instead of the circle icon we currently use: `/static/images/logo/zulip-icon-square.svg`.

[CZO discussion](https://chat.zulip.org/#narrow/stream/101-design/topic/Zulip.20mac.20app.20bar.20icon)

fwcd:
**What's this PR do?**

This fixes #1253 by replacing the current app/Dock icon on macOS with a rectangular version based on [`zulip-icon-square.svg`](https://github.com/zulip/zulip/blob/main/static/images/logo/zulip-icon-square.svg) (see images below).

This PR provides both an SVG and an ICNS version of the icon, along with a small shell script to automatically regenerate the ICNS version from SVG (requires Inkscape).

**Screenshots?**

Icon

With this new icon, Zulip neatly fits into the Dock along with other macOS apps:

Screenshot 2023-01-10 at 23 50 39

- [x] macOS (only relevant there)


fwcd:
@andersk  Would you mind taking a look?

Issue #1252, opened 2022-12-07
Receiving files in a private chat is strange


flaviozluca:
Friends,

It's very strange because when you received a file in a private chat, using the Desktop app, you click on the received file and nothing happens. Click again and you hear a beep.

I eventually discovered it was downloading the file, but there was no visual indication whatsoever of that of where it went to.

The files are downloaded to my download folder. Its just very unclear, the whole process.

I found there's a switch to change this and then it makes it all clearer.

![image](https://user-images.githubusercontent.com/44004645/206201898-65debd84-82b2-4d22-85b1-2db6f902dc91.png)

Ask where to save... That makes it clear, now it asks where I want to save the file to.

Maybe make this on by default?

Because, believe me, I suffered to understand what was going on with the default method. Id click to download the file and nothing would happen, when in reality, it had downloaded it.

Other idea, make downloading clearer, somehow do like browsers, show that the file is being dowloaded and a link to go to the download folder.

Thanks.

https://user-images.githubusercontent.com/44004645/206202321-5e2d8309-a699-4c31-bcc0-658b19719a9b.mp4

varun-s22:
@zulipbot claim

varun-s22:
hey @flaviozluca, For this issue i was guessing maybe i could display a notification on the top, and when the file is downloaded i can open the folder where it is downloaded. 
This behaviour is not working on my machine and i think i can work it out

zulipbot:
Hello @varun-s22!

Thanks for your interest in Zulip! You have attempted to claim an issue without the label "help wanted". You can only claim and submit pull requests for issues with the [help wanted](https://github.com/zulip/zulip-desktop/issues?q=is%3Aopen+is%3Aissue+no%3Aassignee+label%3A%22help+wanted%22) label.

If this is your first time here, we recommend reading our [guide for new contributors](https://zulip.readthedocs.io/en/latest/overview/contributing.html) before getting started.

flaviozluca:
Dear Varun,

That would be awesome! Appreciate it!!

varun-s22:
> Dear Varun,
> 
> That would be awesome! Appreciate it!!

Hey, thanks.. Can you add a label `help-wanted` or maybe assign this issue to me?

flaviozluca:
Hi, Varun
How do I do that?
Thanks.

varun-s22:
I have submitted a PR regarding this. If anyone can review it, it would be great

varun-s22:
[...]

PR #1266, opened 2023-01-20
Added progress bar for electron

varun-s22:
* Added progress on the icon to show progress of files downloaded.
* Dialog box for asking file location is switched on by default.
* File location where the file is downloaded is opened by default.

---

**What's this PR do?**
This PR is in reference to #1252 , where the user gets no notification about the file downloaded. This is currently Work in Progress, as it requires a review. 

I used electron's native progress-bar to show the progress made on downloading. Since I didn't have the size of files downloaded, I used kind of a "hacky" way to deal out of it. 
If i could get the progress of how much file is being downloaded (i.e x MB/ y MB) format, I can show a real-time progress bar

Also this is currently below the

**Screenshots?**
It currently looks like this
Screenshot 2023-01-21 at 02 58 00

**I have tested this PR on:**

- [ ] Windows
- [ ] Linux/Ubuntu
- [x] macOS

Future: I'm looking for external packages in TypeScript, which provides a progress-bar in the UI, so i can render a component on the screen itself.

### DIFF
--- a/app/main/handle-external-link.ts
+++ b/app/main/handle-external-link.ts
@@ -1,5 +1,6 @@
 import {shell} from "electron/common";
 import type {
+  BrowserWindow,
   HandlerDetails,
   SaveDialogOptions,
   WebContents,
@@ -104,6 +105,7 @@ export default function handleExternalLink(
   contents: WebContents,
   details: HandlerDetails,
   mainContents: WebContents,
+  win: BrowserWindow,
 ): void {
   let url: URL;
   try {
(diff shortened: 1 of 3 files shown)


varun-s22:
@akashnimare @andersk @vsvipul Please if you could review this PR, it would be great.

varun-s22:
https://user-images.githubusercontent.com/93476421/214092465-225c3f78-0024-4131-8cec-2dccd72ed44d.mov

This is how it currently looks. Look at the bottom of the dock icon, you can see the progress bar.

alexmv:
Thanks for the contribution!  This looks to be just getting asymptotically close to 100%, with no notion of how long the downlaod actually will take.  This may actually be _more_ confusing for users, since they may be wondering why a big file is "stuck."

The download provides a [Content-Length](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Length) response header on it, which tells you how big the file is.  You should be able to use this to make the progress bar accurate.

varun-s22:
Hey @alexmv, thanks for the review. I updated the file now. I didn't use `Content-Length`, as i couldn't pass the headers into WebContents. I had to change the entire downloading scheme, and use `fetch` for using the above. 
Instead I found a workaround, I explored the `downloadItem` in electron, and found two member functions which are useful for us, namely `getTotalBytes()`, and `getReceivedBytes()`, which solves our problem, and updates the progress bar live with the downloading of the file. 

I hope now it works fine, and gets merged

varun-s22:
Closes #1252

varun-s22:
This fixes #1252

varun-s22:
hey @alexmv @andersk @timabbott, will you please review this PR?

PR #1323, opened 2023-09-07
Update notarize.js : Converting to typescript

RachitSrivastava12:
As the Zulip desktop app is written in typescript converting this code to typescript

### DIFF
--- a/scripts/notarize.js
+++ b/scripts/notarize.js
@@ -1,25 +1,39 @@
-"use strict";
-const path = require("node:path");
-const process = require("node:process");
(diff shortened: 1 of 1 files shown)


andersk:
This is a script that needs to run in Node, and TypeScript does not directly run in Node; it needs a compilation step.

RachitSrivastava12:
Thank you for correcting me. I appreciate it. You are right adding those compilation steps like Installing TypeScript as a Development Dependency, configuring it will only make the task more complex.
Thanks

PR #1352, opened 2024-01-13
macOS: Show alternate tray icon for unread messages

darrinsmart:
This PR implements an alternate tray icon for macOS, which is displayed when there are unread messages.

Many Mac users set the dock to auto-hide in order to maximize screen area. Unfortunately this also hides the unread messages badge and makes it very hard to notice when there are new messages available.

But the menu bar is always visible, and the tray icon is a great way to show that there are unread messages. I can quickly glance up to the menu bar and see that messages are pending, then click that icon to bring Zulip to the foreground.

My graphic design skills are lacking so I simply used an opacity variation to differentiate the icons. This is probably worthy of discussion and I expect could be improved if anyone would like to contribute.

Menubar with unread messages:
Unread

Menubar with no unread messages:
Read

The change only affects macOS clients. I've also tested on Linux, but not Windows.


shubham-padia:
@darrinsmart Please open an issue for any usability problems you're facing. Issues are where we discuss what might the best way to go about solving a user problem. For enhancements, PRs should only be opened for issues where maintainers agree with the proposed solution. Closing this for now, thank you for your contribution!

Issue #1357, opened 2024-03-07
Settings menu buttons are invisible (white on white)


nuke-web3:
Kinda related to #1082 it seems that the white on white toggles when off are not visible

**Screenshots**

![image](https://github.com/zulip/zulip-desktop/assets/35669742/d45a4982-e887-4459-a34a-0e526f37e39d)

I do see a very very slight ghost of them playing with my monitor settings... but I would think a better color or atleast more contrast on the grey used could resolve for everyone (dark mode :pray:)

**Desktop (please complete the following information):**

- Operating System: Pop!_OS
- Zulip Desktop Version: 5.10.5 

---

[CZO discussion](https://chat.zulip.org/#narrow/stream/16-desktop/topic/Button.20color.20change.20vs.20dark.20mode.20in.20settings.20page/near/1770738)

zhngharry:
Is this being tackled already? I would like to try working on this to get into contributing to Zulip.

CodeMaverick2:
hey @timabbott i would like to work on this issue

drashtiv03:
I would like to work on this issue.Please assign it to me @nuke-web3 @timabbott

PR #1367, opened 2024-04-23
Settings: improved visibility of text and buttons

Aitchessbee:
**What's this PR do?**
Improves text and button visibility in settings page as reported in #1357 

**Screenshots?**
Old:
image

New:
image

- [ ] Windows
- [x] Linux/Ubuntu
- [ ] macOS

### DIFF
--- a/app/renderer/css/preference.css
+++ b/app/renderer/css/preference.css
@@ -381,6 +381,10 @@ img.server-info-icon {
   margin: 6px;
 }
 
+.setting-description {
+  color: rgb(34 34 34 / 100%);
+}
+
 #note {
   font-size: 10px;
 }
@@ -459,7 +463,7 @@ i.open-tab-button {
 }
 
 .disallowed:hover {
-  background-color: rgb(241 241 241 / 100%);
+  background-color: rgb(153 153 153 / 100%);
   cursor: not-allowed;
 }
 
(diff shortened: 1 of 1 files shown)


timabbott:
Can you present before/after screenshots? Also, I'm not sure which component is the `settings_description` part.

Aitchessbee:
Update the description with old and new images... `settings_description` is a class given to the text in each row. It was already present in the code but wasn't used in css.

timabbott:
OK. Text definitely feels a lot better. The dark grey slider background might be overdoing it a bit; is there somewhere a bit in between worth trying?

alya:
Maybe, but overall this is a definite improvement.

Axshatt:
> OK. Text definitely feels a lot better. The dark grey slider background might be overdoing it a bit; is there somewhere a bit in between worth trying?

I think the tab headings should also be in dark colour as it will be a good user experience

shubham-padia (changes requested):
Thanks for your PR! Please rebase your PR. Requested 1 change, rest looks good.

app/renderer/css/preference.css
```diff
@@ -381,6 +381,10 @@ img.server-info-icon {
   margin: 6px;
 }
 
+.setting-description {
+  color: rgb(34 34 34 / 100%);
+}
+
```
shubham-padia, line 387: We don't need this anymore, because of https://github.com/zulip/zulip-desktop/commit/b42f9de27d92c358bc4f0979d892d34a27be053d

shubham-padia:
Opened #1468 with the requested change and commit message fixes while keeping the original author same.

timabbott:
Closing in favor of #1468.

Issue #1360, opened 2024-03-15
[Feat]  Option to set default overall zoom for all servers (and minor setting zoom incrementally bug)


nuke-web3:
**Problem Description**

The zoom level set with the view menu or `crtl+<=/->` is configurable per-server, and this is sticky for each instance. One would usually want all app windows to use the same zoom level that would be a global default setting.

**Proposed Solution**

Ideally one would have a menu option to set the default zoom level for all servers. Most [browsers support this kinda thing](https://www.howtogeek.com/657255/how-to-set-a-default-zoom-level-in-firefox/) and it seems there are ways to [expose this in electron](https://ourcodeworld.com/articles/read/547/how-to-change-the-window-zoom-level-in-electron-framework).

Example from the desktop telegram app:

![image](https://github.com/zulip/zulip-desktop/assets/35669742/b19ca37d-d4d9-48d8-936f-a3acb9605884)

The server could remember and override the default setting, using keybindings or perhaps a server level setting (rather than app level)

andersk:
I cannot reproduce your first problem. The per-domain zoom level is remembered across restart for me.

There is a minor bug where the zoom is reset to 100% the first time you change it, though.

(The `setZoomFactor` API is what we’re using; Chromium takes care of remembering it per-domain.)

nuke-web3:
Ah indeed, I cannot seem to reproduce the restart reset of the zoom. I do see the same issue with jump to 100% on any change. 

I will change the title & OP to reflect the minor bug & request to have a global setting to default to, as it's a minor UX nuisance to have them independently configured (unless you for some reason  explicitly want that as an override)

shubham-padia:
Closed, [#desktop > Zoom level global setting](https://chat.zulip.org/#narrow/channel/16-desktop/topic/Zoom.20level.20global.20setting/with/2284438)

PR #1369, opened 2024-04-27
Feature Extension Re: #1360: Option to set default overall zoom for all servers

coxtrent:
**What's this PR do?** Resolves issue #1360 by adding option to synchronize zoom factor across all organizations. Beforehand, the zoom was per organization, making for something of a clunky and sometimes confusing experience. 

- Added a new button/setting to app.renderer.js.pages.preference.general-section.ts
- Used `ipc` to send a message that synchronizes the zoom factors to 1 when the button is turned on, added "sync-zooms" to `type MainMessage` and `type RendererMessage` in typed-ipc.ts
- Added `getZoomFactor()` and `setZoomFactor()` methods to WebView. in app.renderer.js.components.webview.ts (breaking the abstraction is bad)
- Added private member function `syncZooms()` to ServerManagerView that actually sets the values of all tabs' zoom factors to its one argument (default = 1), then call this function from the WebviewListeners for zoom functions. It must be handled here and not from the BrowserWindow object or else it will change the size of the sidebar with the organizations and settings, etc. The function has no effect if the option is off. This handles the synchronous zoom functionality and optionality almost completely outside the old functions while not changing pre-existing code.

- [ ] Windows
- [ ] Linux/Ubuntu
- [ X ] macOS


alya:
Thanks for the work here! Please clean up your commit history and post again to request a review. See [here](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html) for guidelines.

A link to the chat.zulip.org discussion from this PR's description would also be helpful.

coxtrent:
Sure! I noticed  a couple minor kinks in my implementation so I've been taking a bit more time to work on it. Thanks for the pointers by the way

shubham-padia:
Thanks for your fix! Closed due to issue not being relevant anymore: [#desktop > Zoom level global setting](https://chat.zulip.org/#narrow/channel/16-desktop/topic/Zoom.20level.20global.20setting/with/2284438)

PR #1374, opened 2024-07-01
Replace Travis CI badge with GitHub Actions

mbrukman:
Travis CI was replaced with GitHub Actions in this project, and the [current badge][1] appears broken, and the [project link][2] shows an error:

> This is not an active repository

Since this PR doesn't update any code, we can [skip ci] to save resources.

[1]: https://travis-ci.com/zulip/zulip-desktop.svg?branch=main
[2]: https://travis-ci.com/github/zulip/zulip-desktop

---

**What's this PR do?**

Updates badge from Travis CI to GitHub Actions.

Self-evident, I think.

**Screenshots?**

Even better, here's a live preview!

Current, i.e., before my change (yes, it's actually broken):

> [Build Status](https://travis-ci.com/github/zulip/zulip-desktop)

After my change:

> [Build Status](https://github.com/zulip/zulip-desktop/actions/workflows/node.js.yml?query=branch%3Amain)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Zulip Desktop Client
 
-[![Build Status](https://travis-ci.com/zulip/zulip-desktop.svg?branch=main)](https://travis-ci.com/github/zulip/zulip-desktop)
+[![Build Status](https://github.com/zulip/zulip-desktop/actions/workflows/node.js.yml/badge.svg)](https://github.com/zulip/zulip-desktop/actions/workflows/node.js.yml?query=branch%3Amain)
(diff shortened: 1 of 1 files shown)


timabbott:
Thanks for the PR! I'm not entirely sure "Node.js CI" is the most accurate label, but this seems like clearly progress.

mbrukman:
> Thanks for the PR!

You're welcome!

> I'm not entirely sure "Node.js CI" is the most accurate label, but this seems like clearly progress.

That's not something I chose or wrote; this is how the workflow in `node.js.yml` is named:

https://github.com/zulip/zulip-desktop/blob/92260b0f97b55986fd8d6e0e0166a1db2d70b207/.github/workflows/node.js.yml#L1

so if you want it to say anything else, just modify that line as you wish.

PR #1382, opened 2024-08-20
Update to pt_PT language

supermarsx:
**What's this PR do?**

- Updates pt_PT language with better wording
- Update to supported languages to include specific locale pt_PT

- Changed on transifex, downloaded and included here

- [X] Windows
- [ ] Linux/Ubuntu
- [ ] macOS


andersk:
Our translations are handled through Transifex. If you edit them directly in Git they’ll just be overwritten next time we sync from Transifex. See our [Translators’ workflow](https://zulip.readthedocs.io/en/latest/translating/translating.html#translators-workflow) for contributing translations.

supermarsx:
what about adding pt_PT to supported locales?

andersk:
I will handle that when I sync translations for the next release.

PR #1400, opened 2024-12-12
translations: Remove no-op translations which repeat English.

alexmv:


andersk:
supported-locales.json was intentionally sorted by the human-readable name, not by the machine-readable identifier.

andersk:
Also I would expect that many of these no-op translation are legitimate, especially in en_GB.

alexmv:
Updated to leave `en_GB` in there.

The reason to (for now) strip out all no-op translations is that it appears like that's what Transifex does by default, which leaves Weblate thinking the languages are all "100% translated" but with a bunch of alerts about no-op translations.  Doing a one-time clearing out of the no-op translations is a heuristic to try and bring the translation percentages back into line with reality -- and any _actual_ no-op translations can be done one-by-one and marked as intentional to clear their alerts.

andersk:
Part of the issue here is that we should be using the [`--mode=onlytranslated` option](https://developers.transifex.com/docs/using-the-client#other-flags-1) for `tx pull`. That would remove 2020 of the no-op translations. Of the remaining 1455 that are really present in Transifex, most of them are due to an incorrect `tx push` from @akashnimare on Sat, Jun 20, 2020 3:43 PM.

alexmv:
Abandoning -- see [#translation > Desktop translations that repeat English @ 💬](https://chat.zulip.org/#narrow/channel/58-translation/topic/Desktop.20translations.20that.20repeat.20English/near/2256701)

PR #1405, opened 2025-02-03
Fixed: Improved error handling for invalid JSON in global_config.json

sammamama:
Fixed #1404 . A dialog box error now pops up for the people with invalid JSON in their `global_config,json` file.

Fixes: 
I changed the code to throw an error, whenever an error is caught while trying to read `global_config.json` file.

**Platforms this PR was tested on:**

- [x] Windows
- [ ] macOS
- [ ] Linux (specify distro)

<details>
<summary>Self-review checklist</summary>

- [ ] [Self-reviewed](https://zulip.readthedocs.io/en/latest/contributing/code-reviewing.html#how-to-review-code) the changes for clarity and maintainability
      (variable names, code reuse, readability, etc.).

Communicate decisions, questions, and potential concerns.

- [x] Explains differences from previous plans (e.g., issue description).
- [x] Highlights technical choices and bugs encountered.
- [ ] Calls out remaining decisions and concerns.
- [x] Automated tests verify logic where appropriate.

Individual commits are ready for review (see [commit discipline](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html)).

- [x] Each commit is a coherent idea.
- [x] Commit message(s) explain reasoning and motivation for changes.

Completed manual review and testing of the following:

- [x] Visual appearance of the changes.
- [x] Responsiveness and internationalization.
- [ ] Strings and tooltips.
- [x] End-to-end functionality of buttons, interactions and flows.
- [x] Corner cases, error conditions, and easily imagined bugs.
</details>

PR #1425, opened 2025-06-06
Show error dialog box on invalid global_config.json

### LINKED ISSUE #1404: Improve error handling for invalid JSON in `global_config.json`
Users who have invalid JSON in their `global_config.json` file, which is kinda easy to do if you hand-edit JSON and are not good at remembering to skip commas, experience a blank screen when trying to open the application; the only visible errors are in the JS console for the app.

We should introduce a more reasonable user-facing failure here.

shubham-padia:
Fixes #1404.

Supersedes #1405. Since the content of the commit was very similar, I have added @sammamama as a co author.

I wasn't sure about quitting the app after the error so kept it as a separate commit if we want to skip it.

Screenshot 2025-06-06 at 1 15 16 PM

- [ ] Windows
- [x] macOS
- [ ] Linux (specify distro)

<details>

- [ ] Strings and tooltips.
</details>

### DIFF
--- a/app/common/enterprise-util.ts
+++ b/app/common/enterprise-util.ts
@@ -3,6 +3,7 @@ import path from "node:path";
 import process from "node:process";
 
 import {z} from "zod";
+import {dialog} from "zulip:remote";
 
 import {enterpriseConfigSchemata} from "./config-schemata.js";
 import Logger from "./logger-util.js";
@@ -40,6 +41,10 @@ function reloadDatabase(): void {
         .partial()
         .parse(data);
     } catch (error: unknown) {
+      dialog.showErrorBox(
+        "Error loading global_config",
+        "We encountered an error while reading global_config.json, please make sure the file contains valid JSON.",
+      );
       logger.log("Error while JSON parsing global_config.json: ");
       logger.log(error);
     }
--- a/app/common/enterprise-util.ts
+++ b/app/common/enterprise-util.ts
@@ -1,3 +1,4 @@
+import {app} from "electron/main";
 import fs from "node:fs";
 import path from "node:path";
 import process from "node:process";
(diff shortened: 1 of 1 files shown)


timabbott:
Would it make sense to show the actual JSON parser error message? Having a line number in the file to look at can help folks a lot. We could probably cut some of the text if we added that.

shubham-padia:
I set global_config.json to just be `test` for the below error message, changing it to valid JSON makes the app work again.

The message is not really useful, the stacktrace I get is the following.
```
endering chunks (1)...App threw an error during load
ReferenceError: Cannot access 'logger$7' before initialization
    at getConfigItem (/Users/shubhampadia/zulip-desktop/dist-electron/index.js:14239:5)
    at Object.<anonymous> (/Users/shubhampadia/zulip-desktop/dist-electron/index.js:14191:18)
    at Module._compile (node:internal/modules/cjs/loader:1569:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1722:10)
    at Module.load (node:internal/modules/cjs/loader:1296:32)
    at Module._load (node:internal/modules/cjs/loader:1115:12)
    at c._load (node:electron/js2c/node_init:2:18008)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:227:24)
    at cjsLoader (node:internal/modules/esm/translators:266:5)
```

which I think makes things more confusing. Might be better to just stick with the current generic message?

timabbott:
Yeah, this seems fine given that, merged, thanks @shubham-padia!

app/common/enterprise-util.ts
```diff
@@ -1,8 +1,10 @@
+import {app} from "electron/main";
```
andersk, line 1: This does not work. We’re in `app/common` here, which is used from both the main process and the renderer process. We’re not allowed to use main process APIs. In practice, we get `app === undefined` in the renderer process.

shubham-padia: Would #1430 be the right approach for rectifying this?

andersk:
- Reverted in #1432.

PR #1431, opened 2025-06-19
settings: Always show the sidebar in case of settings page.

### LINKED ISSUE #1077: Stuck in preferences
The Mac desktop client is great, thank you!

1. Open Mac desktop client.
2. Select "Desktop Settings" from the "Zulip" menu in the menubar (or type command-, )
3. There doesn't seem to be a way out of this screen.
4. Quit and restart application to make settings go away.

Thank you.

shubham-padia:
Fixes #1077.
Users were getting stuck in the settings page, unable to click on an organisation tab to go back. Enabling the sidebar for settings page in all cases will help ensure that the users can navigate back when needed.

show sidebar

- [ ] Windows
- [x] macOS
- [ ] Linux (specify distro)

<details>

- [ ] Strings and tooltips.
</details>

### DIFF
--- a/app/renderer/js/main.ts
+++ b/app/renderer/js/main.ts
@@ -622,6 +622,9 @@ export class ServerManagerView {
     });
     this.$settingsButton.classList.add("active");
     this.preferenceView!.handleNavigation(navigationItem);
+
+    // We always want to show the sidebar in case of the settings page.
+    this.toggleSidebar(true);
   }
 
   async openAbout(): Promise<void> {
@@ -714,6 +717,12 @@ export class ServerManagerView {
         this.loading.has((await tab.webview).properties.url),
     );
 
+    // We might be coming here from the settings page, which has the
+    // sidebar always enabled. We need to recheck that setting here
+    // and toggle the sidebar as needed.
+    const showSidebar = ConfigUtil.getConfigItem("showSidebar", true);
+    this.toggleSidebar(showSidebar);
+
     ipcRenderer.send("update-menu", {
       // JSON stringify this.tabs to avoid a crash
       // util.inspect is being used to handle circular references
@@ -793,6 +802,11 @@ export class ServerManagerView {
   }
 
   toggleSidebar(show: boolean): void {
+    // In case when user manually clicks on toggleSidebar.
+    if (this.tabs[this.activeTabIndex]?.properties?.page === "Settings") {
+      show = true;
+    }
+
     this.$sidebar.classList.toggle("sidebar-hide", !show);
   }
 


app/renderer/js/main.ts
```diff
@@ -620,6 +620,13 @@ export class ServerManagerView {
     });
     this.$settingsButton.classList.add("active");
     this.preferenceView!.handleNavigation(navigationItem);
+
+    // We need to call toggleSidebar again since we always want to show
+    // the sidebar in case of the settings page. We still pass the real
+    // config value for showSidebar so we  do not reset showSidebar
```
timabbott: spacing type here.

app/renderer/js/main.ts
```diff
@@ -712,6 +719,12 @@ export class ServerManagerView {
         this.loading.has((await tab.webview).properties.url),
     );
 
+    // We might be coming here from the settings page, which has the
+    // sidebar always enabled. We need to recheck that setting here
+    // and toggle the sidebar as needed.
+    const showSidebar = ConfigUtil.getConfigItem("showSidebar", true);
```
timabbott, line 723: Is there a place we declare all the supported items in `ConfigUtil` that this should be added to?

And how persistent are these items?

shubham-padia: It is already declared in https://github.com/zulip/zulip-desktop/blob/51ff949d34bec52089662f889e422d92a19274ec/app/common/config-schemata.ts#L31.

setConfigItem stores it in our JSON settings file: https://github.com/zulip/zulip-desktop/blob/51ff949d34bec52089662f889e422d92a19274ec/app/common/config-util.ts#L58

app/renderer/js/main.ts
shubham-padia: Fixed!

app/renderer/js/main.ts
```diff
@@ -791,6 +804,12 @@ export class ServerManagerView {
   }
 
   toggleSidebar(show: boolean): void {
+    ConfigUtil.setConfigItem("showSidebar", show);
+    // We do not want to hide the sidebar in case of the settings page.
+    if (this.tabs[this.activeTabIndex]?.properties?.page === "Settings") {
+      show = true;
+    }
```
timabbott, line 808: So it appears this is changing the behavior to now write to the settings (the default of whether to show the sidebar on startup) when toggling the sidebar manually. I don't think that's what we were looking to do?

shubham-padia: I'm not sure if that is the case i.e. the default of whether to show the sidebar is changed by [setConfigItem](https://github.com/zulip/zulip-desktop/blob/d270d56309f6a4948abbc039b3b99038de543aa3/app/common/config-util.ts#L58).

From my understanding, it does not override any global config if `override` is not set to true, which it is not in that case.

timabbott: I'm still confused. Previously, this function, which is called from multiple code paths, did not call `setConfigItem` . Now it does. This to me seems like a behavior change that is not part of "Always showing the sidebar in case of settings page". 

If this is fixing another issue that you noticed while working on this PR, maybe it should be its own commit.

But to me it looks like it is introducing a bug. I think previously, the behavior was we have a setting for whether the sidebar is hidden in general in the configuration that you access via the settings panel, and also there's a quick way to toggle it without changing what happens on next startup. This seems to make clicking the quick toggle always change it on next startup.

I don't really understand the purpose of adding that `setConfigItem` here at all, but that's my understanding of what impacts it has.

timabbott:
I think this should not be hard to finish, but we do need to resolve my correctness concerns above before it's ready for Anders to review.

app/renderer/js/main.ts
shubham-padia: Yeah, I think I had a wrong understanding of how getConfigItem and setConfigItem work, fixed!

timabbott:
This version looks good to me.

shubham-padia:
Requesting a review.

PR #1462, opened 2025-10-08
Use tab id and order field where appropriate instead of tabIndex.

shubham-padia:

- [ ] Windows
- [x] macOS
- [ ] Linux (specify distro)

<details>

- [ ] Strings and tooltips.
</details>


app/common/typed-ipc.ts
```diff
@@ -63,7 +63,7 @@ export type RendererMessage = {
   "set-idle": () => void;
   "show-keyboard-shortcuts": () => void;
   "show-notification-settings": () => void;
-  "switch-server-tab": (index: number) => void;
+  "switch-server-tab": (id: string) => void;
```
gnprice: Is this a server ID or a tab ID? (Are those two different things?)

It'd be good to make the answer explicit: use specific names like `serverId` or `tabId`, rather than just `id`, everywhere except the ID property that's right on the object it identifies. (And then if it seems most convenient to use the same name there too, that's also a popular choice.)

app/common/types.ts
```diff
@@ -12,19 +12,23 @@ export type NavigationItem =
   | "Shortcuts";
 
 export type ServerConfig = {
+  id: string;
```
gnprice, line 15: OTOH if the ID property here is just the tab ID of some tab, rather than an independent ID for the server, then this should get a name like `tabId` that makes that explicit.

app/common/typed-ipc.ts
shubham-padia: > Is this a server ID or a tab ID? (Are those two different things?)

We set tab ID to server ID in the app for server tabs, but not the other way around. For functional tabs, tabId is just a unique string ID. 

tabId would be more appropriate to use here, since we are talking about switching tabs.

app/common/types.ts
shubham-padia: It is an independent ID for the server. We set tabId to that independent id in the app. 

Now that I think about it, if this seems confusing, it might be good to have two fields on the tab object, `tabId` can be a unique ID that is independent of the server id. And we store server id in a separate field called `serverId`, which will be `undefined` in case of functional tabs. Functional tabs are settings, reload, etc. 

Let me know what you think of the current approach vs the proposed one vs anything else you might have in mind based on the explanation above, thanks!

gnprice: Good question. Probably a good discussion to have in a chat thread — that'll make it more visible for Anders and/or Tim to chime in too.

shubham-padia: [#desktop > Early feedback for using Tab object instead of indices @ 💬](https://chat.zulip.org/#narrow/channel/16-desktop/topic/Early.20feedback.20for.20using.20Tab.20object.20instead.20of.20indices/near/2285630)

app/renderer/js/main.ts
```diff
@@ -811,14 +809,21 @@ export class ServerManagerView {
     }
   }
 
-  async isLoggedIn(tabId: number): Promise<boolean> {
-    const tab = this.tabs[tabId];
+  async isLoggedIn(index: number): Promise<boolean> {
+    const tab = this.tabs[index];
```
shubham-padia: The previous naming was wrong, this is in fact an index, we remove the use of index in future commits.

app/common/types.ts
shubham-padia: Made this change, ready for another look.

app/main/menu.ts
```diff
@@ -313,8 +313,6 @@ function getWindowSubmenu(
       type: "separator",
     });
     for (const tab of tabs) {
-      // Skip missing elements left by `delete this.tabs[index]` in
-      // ServerManagerView.
       if (tab === undefined) continue;
```
andersk: Should be deleted.

PR #1466, opened 2025-10-29
download: Show in-app notification on downlaoding a file.

### LINKED ISSUE #1371: strange downloading a file
Friends,

Win 11.

Strange, when I click to download a file, nothing happens, but it goes to my download folder.
I click, nothing happens, cant it open the folder where it downloaded to or inform something? It's strange.
image

Thanks

shubham-padia:
Fixes #1371.
We do not accept a filePath from the web app when recieving the event to show the file in a folder, since that could introduce securithy vulnerabilies. We store the latest downloaded file path instead in a variable. Only one in-app notification is active at a time, so storing the latest file path is enough for out use case.

To test this PR, run a zulip server that runs the branch on https://github.com/zulip/zulip/pull/36408 and add it to the zulip desktop app running on this PR. Upload a file and download it to test.

Screenshot 2025-11-06 at 3 14 14 PM
Screenshot 2025-11-06 at 3 04 50 PM

- [ ] Windows
- [x] macOS
- [ ] Linux (specify distro)

<details>

- [ ] Strings and tooltips.
</details>

### DIFF
--- a/app/main/handle-external-link.ts
+++ b/app/main/handle-external-link.ts
@@ -6,6 +6,7 @@ import {
   type WebContents,
   app,
 } from "electron/main";
+import {randomBytes} from "node:crypto";
 import fs from "node:fs";
 import path from "node:path";
 
@@ -14,6 +15,35 @@ import * as LinkUtil from "../common/link-util.ts";
 import * as t from "../common/translation-util.ts";
 
 import {send} from "./typed-ipc-main.ts";
+
+const maxTrackedDownloads = 50;
+
+type DownloadedFile = {
+  id: string;
+  filePath: string;
+};
+
+const downloadedFiles = new Map<string, DownloadedFile>();
+
+function trackDownloadedFile(filePath: string): DownloadedFile {
+  const downloadedFile: DownloadedFile = {
+    id: randomBytes(16).toString("hex"),
+    filePath,
+  };
+
+  downloadedFiles.set(downloadedFile.id, downloadedFile);
+
+  if (downloadedFiles.size > maxTrackedDownloads) {
+    const oldestDownloadedFileId = [...downloadedFiles.keys()][0];
+    downloadedFiles.delete(oldestDownloadedFileId);
+  }
+
+  return downloadedFile;
+}
+
+export function getDownloadedFilePath(downloadId: string): string | undefined {
+  return downloadedFiles.get(downloadId)?.filePath;
+}
 
 function isUploadsUrl(server: string, url: URL): boolean {
   return url.origin === server && url.pathname.startsWith("/user_uploads/");
@@ -125,9 +155,14 @@ export default function handleExternalLink(
       url: url.href,
       downloadPath,
       async completed(filePath: string, fileName: string) {
+        const notificationTitle = t.__("Downloaded {{{fileName}}}.", {
+          fileName,
+        });
+        const notificationBody = t.__("Click to open downloads folder.");
+        // Show native notification
         const downloadNotification = new Notification({
-          title: t.__("Download Complete"),
-          body: t.__("Click to show {{{fileName}}} in folder", {fileName}),
+          title: notificationTitle,
+          body: notificationBody,
           silent: true, // We'll play our own sound - ding.ogg
         });
         downloadNotification.on("click", () => {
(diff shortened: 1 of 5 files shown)


app/main/handle-external-link.ts
```diff
@@ -125,16 +131,30 @@ export default function handleExternalLink(
       url: url.href,
       downloadPath,
       async completed(filePath: string, fileName: string) {
+        const notificationTitle = t.__("Download Complete");
```
alya: We should use our standard capitalization pattern.

```suggestion
        const notificationTitle = t.__("Download complete");
```

app/main/handle-external-link.ts
```diff
@@ -125,16 +131,30 @@ export default function handleExternalLink(
...
           shell.showItemInFolder(filePath);
         });
         downloadNotification.show();
+        latestDownloadPath = filePath;
+        send(
+          contents,
+          "show-download-success",
+          notificationTitle,
+          notificationBody,
+        );
```
timabbott, line 182: This seems to do both the old notification mechanism and the new one -- is that intendedd? 

It is important for backwards-compatibility to make sure we do the old thing on existing web app versions. But I don't know if doing both would be problematic, and in any case, I would expect some sort of detection logic and comments explaining the reasoning behind what we're doing.

app/main/handle-external-link.ts
shubham-padia: Fixed!

app/main/handle-external-link.ts
shubham-padia: Yes, that is intended. My understanding of the requirements is that we want to show both in-app and native notifications. @alya let me know if that is not the case.

Added comments.

alya: Good question, I was imagining we'd have both, but I'm not sure if it's silly to do it that way. If you'd be up for it, it might be good to check how a few other apps handle it. Trying Discord, Slack, and WhatsApp, it seems like they pick one or the other, unless my desktop notifications aren't working properly.

alya: So possibly we'd suppress the native notification when showing the in-app one.

timabbott: I think we don't want to send both. There are some potential advantages to the native notification; It appears even if your desktop app is not focused -- say if you started downloading a long video and then tabbed away.

Is there a way to check if the native notification actually sent?

app/main/handle-external-link.ts
```diff
@@ -125,16 +131,33 @@ export default function handleExternalLink(
...
+          },
+        );
+        // Show native notification
         const downloadNotification = new Notification({
-          title: t.__("Download Complete"),
-          body: t.__("Click to show {{{fileName}}} in folder", {fileName}),
+          title: notificationTitle,
+          body: notificationBody,
           silent: true, // We'll play our own sound - ding.ogg
         });
```
timabbott, line 167: So the original issue is a complaint about the native notifications not working on Windows. Do we understand how that's happening? I feel like papering over a bug like that without understanding it is risky -- have you tried arranging an interactive debugging session with someone experiencing the bug?

shubham-padia: The user didn't confirm whether they had desktop notifications enabled or not. 
Screenshot 2025-11-06 at 2 31 18 PM

The original intention was to solve [#desktop > several clicks to download a file](https://chat.zulip.org/#narrow/channel/16-desktop/topic/several.20clicks.20to.20download.20a.20file/with/2289097), I can unlink the issue from the PR if we think it's a windows bug (which I highly doubt). Reply from the issue poster felt incoherent.

app/main/handle-external-link.ts
shubham-padia: > Is there a way to check if the native notification actually sent?

I checked this and there's not for all OS. Only windows emits a failed event: https://www.electronjs.org/docs/latest/api/notification#event-failed-windows.

Might be ok having both notifications in that case?

shubham-padia:
Updated the PR to store the last 50 downloaded files in memory. I tested the oldest file removal logic by printing the downloadedFiles map in console. The PR shows both native and in-app notification right now since it is not possible to know when the user has disabled notifications from system-side.

PR #1501, opened 2026-03-13
Autocomplete for new organisation

### LINKED ISSUE #1012: add organization: url: attempt to autocomplete with zulipchat.com

Paperyduke6:
Fixes #1012

Autocomplete to the Organisation URL input and adds inline ghost-text to show the behaviour visually while typing.

Problem
Typing myteam and hitting Connect now resolves to https://myteam instead of https://myteam.zulipchat.com.  it should be shown to the user within the input field before they submit.

Changes Made
HTML (new-server-form.ts)

Restored the autoComplete function — bare subdomains are expanded to https://<subdomain>.zulipchat.com on submit, anything containing a dot, slash, or protocol is passed through untouched
Wrapped the <input> inside a <label> styled to look like the input box to allow the ghost suffix to sit inline with the typed text
Added a ghost suffix <span> that displays .zulipchat.com directly after the cursor when a bare subdomain is detected, giving the user a preview before hitting Connect
Added an invisible measuring <span> that mirrors the typed text so its pixel width can be read and applied to the input, keeping the suffix flush against the last character

CSS (preferences.css)

Added label#new-server-url-label:focus-within alongside the existing .setting-input-value:focus rule so the teal focus border fires correctly through the label wrapper
Added width: 100% as the default on #new-server-url-input so the placeholder renders in full, with JS overriding this to a pixel value only while the suffix is visible
All new rules are scoped to four unique IDs (#new-server-url-label, #new-server-url-input, #new-server-url-domain, #new-server-url-size-calc) so there is zero risk of affecting any other input or form in the preferences component

Testing

Verified myteam now correctly resolves to https://myteam.zulipchat.com on submit
Tested with full URLs, self-hosted domains, and explicit http:// to confirm they are passed through untouched

Screen Capture

https://github.com/user-attachments/assets/b4ae83d6-4738-480a-9d97-9842b154d5c4

### DIFF
--- a/app/renderer/js/pages/preference/new-server-form.ts
+++ b/app/renderer/js/pages/preference/new-server-form.ts
@@ -12,6 +12,10 @@ type NewServerFormProperties = {
   onChange: () => void;
 };
 
+// Subdomain names may only contain lowercase letters, digits, and hyphens.
+// See issue #1012: https://github.com/zulip/zulip-desktop/issues/1012
+const subdomainPattern = /^[a-z\d-]+$/;
+
 export function initNewServerForm({
   $root,
   onChange,
@@ -20,13 +24,17 @@ export function initNewServerForm({
     <div class="server-input-container">
       <div class="title">${t.__("Organization URL")}</div>
       <div class="add-server-info-row">
-        <input
-          class="setting-input-value"
-          autofocus
-          placeholder="${t.__(
-            "your-organization.zulipchat.com or zulip.your-organization.com",
-          )}"
-        />
+        <label class="setting-input-value" id="new-server-url-label">
+          <input
+            id="new-server-url-input"
+            autofocus
+            placeholder="${t.__(
+              "your-organization.zulipchat.com or zulip.your-organization.com",
+            )}"
+          />
+          <span id="new-server-url-domain"></span>
+          <span id="new-server-url-size-calc"></span>
+        </label>
       </div>
       <div class="server-center">
         <button id="connect">${t.__("Connect")}</button>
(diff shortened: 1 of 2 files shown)


timabbott:
The commit history here needs to be cleaned up. The zulip/zulip contributing guide will be very helpful on this and other advice for making a quality reviewable PR.

Check out our [GitHub guide](https://zulip.readthedocs.io/en/latest/git-guide.html) and [commit guidelines](https://zulip.readthedocs.io/en/latest/contributing/commit-discipline.html) for more details.

app/renderer/js/pages/preference/new-server-form.ts
```diff
@@ -12,6 +12,8 @@ type NewServerFormProperties = {
   onChange: () => void;
 };
 
+const subdomainPattern = /^[a-zA-Z\d-]+$/;
```
timabbott: How did you check whether this was correct? Ideally a reference to zulip/zulip API docs would be good here.

Paperyduke6: For the pattern I referred the issue discussion where the spec is described as "subdomain can only have lowercase letters, numbers, and dashes." I also checked zulip.com/help/change-organization-url which confirms the <subdomain>.zulipchat.com format but doesn't mention character constraints.

Looking at this more carefully, the current pattern /^[a-zA-Z\d-]+$/ permits uppercase letters. I'll tighten it to /^[a-z\d-]+$/ and add a comment explaining the constraint.

I'll also clean up the commit history into a single well-described commit now.

PR #1519, opened 2026-07-03
linux: Fix Wayland app ID

ReillyBrogan:
If `desktopName` is present in the package.json Electron will set the Wayland app ID to the value sans the `.desktop` suffix. It will also set the X11 WINDOW_CLASS if launched as an X11 window.

This means that windows will be correctly associated with a desktop file with name `org.zulip.Zulip.desktop` (already true of the flatpak). `desktopName` has no effect on Windows or Mac.

Before:
```
> niri msg pick-window

Window ID 18: (focused)
  ...
  App ID: "zulip"
  ...
```

After:
```
> niri msg pick-window

Window ID 19: (focused)
  ...
  App ID: "org.zulip.Zulip"
  ...
```

Note that Linux distributions that ship zulip packages with `zulip.desktop` or `Zulip.desktop` will need to patch or sed the package.json prior to building zulip if they wish to avoid renaming their .desktop files. They can just run something like this during their builds:
```
> sed -i "s|org.zulip.Zulip.desktop|Zulip.desktop|" package.json
```

- [ ] Windows
- [ ] macOS
- [*] Linux (AerynOS)


andersk:
Isn’t this the purpose of [`StartupWMClass`](https://specifications.freedesktop.org/desktop-entry/latest-single/#key-startupwmclass), which is already set in our `.desktop` files by electron-builder?

We ship `zulip.desktop` in our first-party `.deb` and `.snap` packages, so any change needs to account for this. The Flathub package has shipped `org.zulip.Zulip.desktop` since before I took over its maintenance from a third party, even though upstream has never used that identifier (there’s a [Flathub issue](https://github.com/flathub/org.zulip.Zulip/issues/4) for changing it, though I’m not sure how to determine whether that’s a good idea).

All of these `.desktop` files have `StartupWMClass=Zulip`. We may need to lowercase that, or we might consider switching to a reversed DNS identifier (we already use `org.zulip.zulip-electron` as `appId`, and I think that can’t be changed without breaking the auto-updater). Either way we need to carefully investigate the compatibility implications.

ReillyBrogan:
> Isn’t this the purpose of [`StartupWMClass`](https://specifications.freedesktop.org/desktop-entry/latest-single/#key-startupwmclass)

No, StartupWMClass is an X11-specific key intended for the [startup notification protocol](https://specifications.freedesktop.org/startup-notification/0.1/). It's not part of the Wayland specs and only a handful of compositors (namely Mutter) continue to support it in order to get a few more non-compliant applications working. In general it should almost always be removed if present in modern .desktop files.

The name of the .desktop file is the only fully compliant way that Wayland windows should be matched to the corresponding Wayland AppId. Since the Flatpak uses `org.zulip.Zulip.desktop` then that means that Zulip needs to set `org.zulip.Zulip` for the Wayland AppId which is accomplished by this PR.

